PR #3, opened 2019-01-04
OpenShift/K8S Operator

eivantsov:
# Che/Codeready Operator

Che Operator creates Eclipse Che k8s and OpenShift resources such as pvcs, services, deployments, routes, ingresses etc.

The operator is a k8s a pod that runs an image with Go runtime and a compiled binary of an operator itself.

Though operator-sdk framework is used, Che operator is rather **an installer** since no CRD and API group are created,
and thus, the operator **does not watch resources**. Once deployment is completed, the operator pod exits.

## Pre-Reqs

OpenShift/K8S cluster with at least 4GB or RAM and 2 PVs, local `oc` or `kubectl`.

## How to deploy

Deploy script will create a namespace, operator service account and a rolebinding for it (admin privileges within the namespace),
and run an operator pod that will create all required objects and perform provisioning:

```bash
deploy/deploy.sh $infra $namespace
```

Deploy to MiniKube to a default namespace:

```bash
deploy/deploy.sh minikube

```
`minikube` as infra argument will instruct the script to detect MiniKube IP and pass global ingress domain value to the operator.

Create default namespace eclipse-che and deploy to k8s:

```
deploy/deploy.sh k8s
```

Create a default eclipse-che namespace and deploy to OpenShift:

```
deploy/deploy.sh
```

Create a namespace of choice and deploy to OpenShift:

```
deploy/deploy.sh openshift myproject
```

If a namespace cannot be created (for example, it exists), a new namespace with a generated name (`eclipse-che${random int}`) will be created.

This will deploy Che operator with the default settings:

* upstream Che
* no tls
* no login with OpenShift in Che
* Postgres passwords are auto-generated, Keycloak admin password is `admin`
* Some object names are default ones (eg databases, users etc)
* Common PVC strategy (all workspaces use one shared PVC)
* All workspace objects get created in a target namespace (Che server uses service account token)
* Multi-host ingress strategy when on k8s

## Deploy custom image

Provide your own image in `deploy/config.yaml`:

```shell
CHE_IMAGE: "myRegistry/myImage:myTag"
```

## Deploy CodeReady Workspaces

In `deploy/config.yaml`:

```
CHE_FLAVOR: "codeready"
```
Che flavor is a string that will be used in names of certain objects like routes, realms, clients etc.

## Defaults and Configuration

To deploy to OpenShift with all defaults, no user input is required. You may configure Che installation in a configmap `deploy/config.yaml`.
The operator will use envs from this configmap and make decisions accordingly.

Currently, only the most critical envs are added to configmap, and it will expand in time, making it possible to fine tune Che before deploying

## What is deployed?

The Operator creates a handful of objects for:

* Postgres DB
* Keycloak/Red Hat SSO
* Che server itself

After Postgres and Keycloak pods start and health checks confirm the services are up, the operator run execs into db and keycloak pods to provision
databases, users, Keycloak realm, client and user. The operator watches respective deployments to get events signalling that the deployment is scaled to 1.

## How to configure installation

`deploy/config.yaml` is a config map with env variables that influences choices an operator makes. Each env is commented, and each one has defaults.
This configmap is an operator's env, and evn variables are then taken to Che server configmap.

What can be configured:

* external DB and Keycloak: `CHE_EXTERNAL_DB` and `CHE_EXTERNAL_KEYCLOAK` default to false.
If you do not need instances of Postgres and Keycloak and want to connect to own infra, set both envs to `true` and provide connection details in envs below the above booleans.

Your DB user **MUST** be a `SUPERUSER`.

**Important!** The operator does not perform Postgres and Keycloak provisioning if external instances are used.
Thus, you need to pre-create db and user for Che.
Also create (or use existing) realm and client that should be public and should have:

Redirect URIs: `${PROTOCOL}://${CHE_HOST}/*`
WebOrigins: `${PROTOCOL}://${CHE_HOST}`

* Login with OpenShift in Che. Not supported on k8s infra

* TLS. Set `TLS_SUPPORT` to true if you want to deploy Che in https mode.
When on k8s, make sure you create a secret in che namespace and provide its name in `TLS_SECRET_NAME`

* TLS and self signed certs. Provide a base64 string of your self signed cert in `SELF_SIGNED_CERT`. When on k8s, you can get it from crt part of the secret.

* Fake dns. If, for example, you want to deploy on k8s with a fake ingress domain of example.com and you need to point it to your Minikube IP, `HOST_ALIAS_IP` and `HOST_ALIAS_HOSTNAME` are the two envs to use.

### Build an image

To build an Operator and package it into a Docker image run:

`docker build -t che-operator .`

If you want to deploy your custom Operator image to a remote k8s/OpenShift cluster, make sure you push an image and change its name in
`--image` and `--overrides` in kubectl/oc run command in the script.

### Build and deploy to a local cluster

Run `deploy/build_deploy_local.sh $infra $namespace`

Minishift and Minikube users will need to execute the below command to use VM Docker daemon when building an image:

```
eval $(minikube docker-env)
```

## How to add new envs to configmap?

If you need to add new envs to be added to Che configmap by default:

* add key:value to deploy/config.yaml
* add operator variable to `pkg/operator/vars.go` that will take its value from environment
* add env and operator variable as value to `pkg/operator/che_cm.go`

## Deploy Script

It is something quick and dirty and is likely to be substituted with a feature rich CLI.


skabashnyuk:
@eivantsov WDYT can this pr be split on two. One without vendors folder and vendors folder separately? Shouldn't it be easier to review then? 
@garagatyi @AndrienkoAleksandr @sleshchenko  thoughts?

sleshchenko:
@skabashnyuk Vendor changes in this PR are OK for me since they are separated in a separate commit.

skabashnyuk:
@sleshchenko ok. missed that.

AndrienkoAleksandr:
+1 for split commit.

garagatyi:
49627770_1615906388555129_1877023561596534784_n
contributed by @mshaposhnik

eivantsov:
@garagatyi what's that supposed to mean?

garagatyi:
selection_047
```
                        |
                        V
```
selection_048

eivantsov:
operator-sdk dependencies and dependencies of those dependencies

garagatyi:
@skabashnyuk @sleshchenko Would be happy to share PR review duty with you

README.md
```diff
@@ -0,0 +1,156 @@
...
+Though operator-sdk framework is used, Che operator is rather an installer since no CRD and API group are created,
+and thus, the operator does not watch resources. Once deployment is completed, the operator pod exits.
+
+## Pre-Reqs
+
+OpenShift/K8S cluster with at least 4GB or RAM and 2 PVs, local `oc` or `kubectl`.
+
+## How to deploy
+
+Deploy script will create a namespace, operator service account and a rolebinding for it (admin privileges within the namespace),
```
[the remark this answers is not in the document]
eivantsov: Yes, the operator does not do clean up.

README.md
```diff
@@ -0,0 +1,156 @@
...
+databases, users, Keycloak realm, client and user. The operator watches respective deployments to get events signalling that the deployment is scaled to 1.
+
+## How to configure installation
+
+`deploy/config.yaml` is a config map with env variables that influences choices an operator makes. Each env is commented, and each one has defaults.
+This configmap is an operator's env, and evn variables are then taken to Che server configmap.
+
+What can be configured:
+
+* external DB and Keycloak: `CHE_EXTERNAL_DB` and `CHE_EXTERNAL_KEYCLOAK` default to false.
```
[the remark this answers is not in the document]
eivantsov: Makes sense. Will fix it.

README.md
```diff
@@ -0,0 +1,156 @@
...
+If you do not need instances of Postgres and Keycloak and want to connect to own infra, set both envs to `true` and provide connection details in envs below the above booleans.
+
+Your DB user **MUST** be a `SUPERUSER`.
```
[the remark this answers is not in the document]
eivantsov: @skabashnyuk may tell more, but Che4 server won;t start unless a db user is a SUPERUSER.. it installs some extension.

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN cd /go/src/github.com/eclipse/che-operator && go test -v ./...
+RUN OOS=linux GOARCH=amd64 CGO_ENABLED=0 \
+    go build -o /tmp/che-operator/che-operator \
+    /go/src/github.com/eclipse/che-operator/cmd/che-operator/main.go
+
+FROM alpine:3.7
+COPY --from=builder /tmp/che-operator/che-operator /usr/local/bin/che-operator
+COPY --from=builder /go/src/github.com/eclipse/che-operator/deploy/keycloak_provision /tmp/keycloak_provision
+RUN adduser -D che-operator
+USER che-operator
```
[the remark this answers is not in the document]
eivantsov: Thanks, will look at it

README.md
```diff
@@ -0,0 +1,156 @@
...
+
+If you need to add new envs to be added to Che configmap by default:
+
+* add key:value to deploy/config.yaml
+* add operator variable to `pkg/operator/vars.go` that will take its value from environment
+* add env and operator variable as value to `pkg/operator/che_cm.go`
+
+## Deploy Script
+
+It is something quick and dirty and is likely to be substituted with a feature rich CLI.
```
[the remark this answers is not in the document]
eivantsov: Yes, operators are mostly deployed with kubectl create -f /path/to/yaml. And yeah, I have not seen a CLI that deploys and manipulates an Operator.

pkg/operator/exec.go
```diff
@@ -0,0 +1,26 @@
...
+	pgCommand = "psql -c \"CREATE USER keycloak WITH PASSWORD '" + keycloakPostgresPassword + "'\" " +
+		"&& psql -c \"CREATE DATABASE keycloak\" " +
+		"&& psql -c \"GRANT ALL PRIVILEGES ON DATABASE keycloak TO keycloak\" " +
+		"&& psql -c \"ALTER USER " + chePostgresUser + " WITH SUPERUSER\""
+)
+
+func ExecIntoPod(podName string, provisionCommand string) {
+	k8s := GetK8SConfig()
+	command := []string{"/bin/bash", "-c", provisionCommand}
+	logrus.Infof("Provisioning resources in pod %s", podName)
```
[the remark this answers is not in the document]
eivantsov: Copied from somewhere else? Why?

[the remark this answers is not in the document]
garagatyi: Because it says `provisioning resources in pod` but function name is `ExecIntoPod` and exec doesn't mean provisioning of any resource. The function name and log messages need to be aligned to avoid confusion.

pkg/util/util.go
```diff
@@ -0,0 +1,90 @@
...
+	"math/rand"
+	"os"
+)
+
+// GetEnvValue looks for env variables in Operator pod to configure Code Ready deployments
+// with things like db users, passwords and deployment options in general. Envs are set in
+// a ConfigMap at deploy/config.yaml. Find more details on deployment options in README.md
+func GetEnv(key, defaultValue string) string {
+	value := os.Getenv(key)
+	if len(value) == 0 {
```
garagatyi: What if someone decided to set an empty value? Would this function behave correctly?

pkg/util/util.go
```diff
@@ -0,0 +1,90 @@
...
+		return defaultValue
+	}
+	if value == "true" {
+		return true
+	}
+	return false
+}
+
+func GeneratePasswd(stringLength int) (passwd string) {
+	chars := []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZ" +
```
garagatyi: I believe that there must be string generation utils in golang stdlib, but as an initial solution it is also OK

pkg/operator/vars.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"github.com/eclipse/che-operator/pkg/util"
+	"strings"
+)
+
+var (
+	// general config
+	cheFlavor           = util.GetEnv("CHE_FLAVOR", "che")
+	namespace           = util.GetNamespace()
+	protocol            = "http"
```
garagatyi: Can you elaborate why it is hardcoded?

pkg/operator/che_cm.go
```diff
@@ -0,0 +1,133 @@
...
+			"-XX:MinHeapFreeRatio=10 -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 " +
+			"-Dsun.zip.disableMemoryMapping=true " +
+			"-Xms20m -Djava.security.egd=file:/dev/./urandom " + cheWorkspaceProxyJavaOptions,
+		"CHE_WORKSPACE_MAVEN__OPTIONS": "-XX:MaxRAM=150m -XX:MaxRAMFraction=2 -XX:+UseParallelGC " +
+			"-XX:MinHeapFreeRatio=10 -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 " +
+			"-Dsun.zip.disableMemoryMapping=true " +
+			"-Xms20m -Djava.security.egd=file:/dev/./urandom " + cheWorkspaceProxyJavaOptions,
+		"CHE_WORKSPACE_HTTP__PROXY__JAVA__OPTIONS": cheWorkspaceProxyJavaOptions,
+		"CHE_WORKSPACE_HTTP__PROXY":                cheWorkspaceHttpProxy,
+		"CHE_WORKSPACE_HTTPS__PROXY":               cheWorkspaceHttpsProxy,
```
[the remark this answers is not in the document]
eivantsov: Yes, that is an excellent idea. I was thinking about giving a user possibility to create a configmap that is then either used in Che deployment or merged into an existing che cm. Probably the first option is better since merging/updating brings unnecessary complexity.

pkg/util/util.go
eivantsov: Hm.. good question. Let me experiment a bit and I'll write up a respose

pkg/operator/vars.go
eivantsov: Che flavor is a default that can be overridden in a configmap

pkg/operator/che_cm.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	// k8s specific envs
+	k8sCheEnv := map[string]string{
+		"CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_FS__GROUP":     "0",
+		"CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_RUN__AS__USER": "0",
+		"CHE_INFRA_KUBERNETES_NAMESPACE":                           workspacesNamespace,
+		"CHE_INFRA_KUBERNETES_INGRESS_DOMAIN":                      ingressDomain,
+		"CHE_INFRA_KUBERNETES_SERVER__STRATEGY":                    strategy,
+		"CHE_INFRA_KUBERNETES_INGRESS_ANNOTATIONS__JSON":           "{\"kubernetes.io/ingress.class\": " + ingressClass + ", \"nginx.ingress.kubernetes.io/rewrite-target\": \"/\",\"nginx.ingress.kubernetes.io/ssl-redirect\": " + tls + ",\"nginx.ingress.kubernetes.io/proxy-connect-timeout\": \"3600\",\"nginx.ingress.kubernetes.io/proxy-read-timeout\": \"3600\"}",
```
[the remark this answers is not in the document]
garagatyi: Yes

pkg/operator/vars.go
garagatyi: I mean protocol

pkg/operator/che_cm.go
```diff
@@ -0,0 +1,133 @@
...
+		"CHE_WEBSOCKET_ENDPOINT": wsprotocol + "://" + cheHost + "/api/websocket",
+		// todo Make configurable
+		"CHE_DEBUG_SERVER":                              "false",
+		"CHE_INFRASTRUCTURE_ACTIVE":                     infra,
+		"CHE_INFRA_KUBERNETES_BOOTSTRAPPER_BINARY__URL": protocol + "://" + cheHost + "/agent-binaries/linux_amd64/bootstrapper/bootstrapper",
+		"CHE_INFRA_OPENSHIFT_PROJECT":                   workspacesNamespace,
+		"CHE_INFRA_KUBERNETES_PVC_STRATEGY":             pvcStrategy,
+		"CHE_INFRA_KUBERNETES_PVC_QUANTITY":             pvcClaimSize,
+		"CHE_INFRA_OPENSHIFT_TLS__ENABLED":              tls,
+		"CHE_INFRA_KUBERNETES_TRUST__CERTS":             tls,
```
sleshchenko: Should we really trust certs if tls is enabled? Should not right certificate be provided for fabric8 client instead?

pkg/operator/che_cm.go
```diff
@@ -0,0 +1,133 @@
...
+		"CHE_JDBC_URL":                                  "jdbc:postgresql://" + postgresHostName + ":" + postgresPort + "/" + chePostgresDb,
+		"CHE_JDBC_USERNAME":                             chePostgresUser,
+		// todo Create a secret for it?
+		"CHE_JDBC_PASSWORD":                             chePostgresPassword,
+		"CHE_LOG_LEVEL":                                 "INFO",
+		"CHE_KEYCLOAK_AUTH__SERVER__URL":                keycloakURL + "/auth",
+		"CHE_INFRA_OPENSHIFT_OAUTH__IDENTITY__PROVIDER": openShiftIdentityProviderId,
+		"CHE_PREDEFINED_STACKS_RELOAD__ON__START":       "true",
+		"CHE_INFRA_KUBERNETES_SERVICE__ACCOUNT__NAME":   "che-workspace",
```
sleshchenko: Can we use constant value here. Then the same constant may be reused during workspace service account creating https://github.com/eclipse/che-operator/pull/3/commits/673c24c26c5b24cd01c831fba7a9a5f7f9d0ad1c#diff-0b5c445f7c199f4fc2bc74e3cc61cef8R30

pkg/operator/reconcile.go
```diff
@@ -0,0 +1,110 @@
...
+	// Create and watch Postgres related resources, unless external DB hostname is provided
+	if ! externalDb {
+		CreatePVC("postgres-data", "1Gi", postgresLabels)
+		CreateService("postgres", postgresLabels, "postgres", 5432)
+		CreatePostgresDeployment()
+		// provision 2 databases, 1 user, and grant superuser privileges to Che pg user
+		ExecIntoPod(k8s.GetDeploymentPod("postgres"), pgCommand)
+	}
+
+	// Che related envs required for Keycloak
```
sleshchenko: Could you explain how `Che related envs` are related to `che-host` service?

pkg/operator/reconcile.go
```diff
@@ -0,0 +1,110 @@
...
+	CreateService("che-host", cheLabels, "http", 8080)
+
+	if tlsSupport {
+		protocol = "https"
+		wsprotocol = "wss"
+	}
+	// create Che route when on OpenShift infra
+	if infra == "openshift" {
+		rt, _ := CreateRouteIfNotExists(cheFlavor, "che-host")
```
sleshchenko: What if any exception occurred during route creating? Should we check error?

pkg/operator/reconcile.go
```diff
@@ -0,0 +1,110 @@
...
+	cheConfigMap := CreateCheConfigMap(cheHost, keycloakURL)
+	err := sdk.Update(cheConfigMap)
+	if err != nil {
+		logrus.Errorf("Failed to update Che configmap : %v", err)
+	}
+	if cheFlavor == "codeready" {
+		cheImage = util.GetEnv("CHE_IMAGE", "registry.access.redhat.com/codeready-workspaces-beta/server:latest")
+	}
+	cheDc, _ := CreateCheDeployment(cheImage)
+	err = sdk.Update(cheDc)
```
sleshchenko: Could you explain why do you need to update `cheDc` just after creation?

pkg/operator/vars.go
```diff
@@ -0,0 +1,75 @@
...
+	wsprotocol          = "ws"
+	cheHost             string
+	tlsSupport          = util.GetEnvBool("CHE_TLS_SUPPORT", false)
+	pvcStrategy         = util.GetEnv("CHE_INFRA_KUBERNETES_PVC_STRATEGY", "common")
+	pvcClaimSize        = util.GetEnv("CHE_INFRA_KUBERNETES_PVC_QUANTITY", "1Gi")
+	selfSignedCert      = util.GetEnv("CHE_SELF__SIGNED__CERT", "")
+	openshiftOAuth      = util.GetEnvBool("CHE_OPENSHIFT_OAUTH", false)
+	oauthSecret         = util.GeneratePasswd(12)
+	oAuthClientName     = "openshift-identity-provider-" + strings.ToLower(util.GeneratePasswd(4))
```
sleshchenko: it's a bit strange to see using password as prefix of oAuth client. Maybe it makes sense (maybe not) rename `GeneratePasswd` to `GenerateString`

pkg/operator/vars.go
```diff
@@ -0,0 +1,75 @@
...
+	cheWorkspaceHttpProxy        = util.GetEnv("CHE_WORKSPACE_HTTP__PROXY", "")
+	cheWorkspaceHttpsProxy       = util.GetEnv("CHE_WORKSPACE_HTTPS__PROXY", "")
+	cheWorkspaceNoProxy          = util.GetEnv("CHE_WORKSPACE_NO__PROXY", "")
+
+	// plugin registry url
+	pluginRegistryUrl = util.GetEnv("CHE_WORKSPACE_PLUGIN__REGISTRY__URL", "https://che-plugin-registry.openshift.io")
+
+	// k8s specific config
+
+	ingressDomain = util.GetEnv("CHE_INFRA_KUBERNETES_INGRESS_DOMAIN", "192.168.42.114")
```
sleshchenko: Are you sure that this IP should be used as default value?

pkg/operator/che_cm.go
```diff
@@ -0,0 +1,133 @@
...
+	strategy := util.GetEnv("CHE_INFRA_KUBERNETES_SERVER__STRATEGY", "multi-host")
+	workspacesNamespace := namespace
+	tls := "false"
+	openShiftIdentityProviderId := "NULL"
+	if openshiftOAuth {
+		workspacesNamespace = ""
+		openShiftIdentityProviderId = "openshift-v3"
+	}
+	if tlsSupport {
+		protocol = "https"
```
sleshchenko: here you update global vars that were already updated during keycloak deploying https://github.com/eclipse/che-operator/pull/3/commits/673c24c26c5b24cd01c831fba7a9a5f7f9d0ad1c#diff-0b5c445f7c199f4fc2bc74e3cc61cef8R52

Please consider initialization such global variable in one place before deploying.

pkg/operator/che_sa.go
```diff
@@ -0,0 +1,48 @@
...
+	labels := map[string]string{"app": "che"}
+	return &corev1.ServiceAccount{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "ServiceAccount",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:     name,
+			Namespace: namespace,
+			Labels:    labels,
```
sleshchenko: Should not `cheLabels` from global vars be used here?

pkg/operator/oauth_client.go
```diff
@@ -0,0 +1,51 @@
...
+
+func newOAuthClient(name string, oauthSecret string, keycloakURL string, keycloakRealm string ) *oauth.OAuthClient {
+	return &oauth.OAuthClient{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "OAuthClient",
+			APIVersion: oauth.SchemeGroupVersion.String(),
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Labels: map[string]string{"app":"che"},
```
sleshchenko: Should not `cheLabels` from global var be used here?

pkg/operator/pvc.go
```diff
@@ -0,0 +1,58 @@
...
+			},
+		},
+	}
+
+}
+// CreatePVC creates a persistent volume claim with a given name, claim size, access mode and labels
+func CreatePVC(name string, pvcClaimSize string, labels map[string]string) *corev1.PersistentVolumeClaim {
+	pvc := pvc(name, pvcClaimSize, labels)
+	if err := sdk.Create(pvc); err != nil && !errors.IsAlreadyExists(err) {
+		logrus.Errorf("Failed to create "+name+" PVC : %v", err)
```
sleshchenko: Since you use `Errorf` you can avoid strings concating by rewriting it in the following way
```go
logrus.Errorf("Failed to create `%s` PVC: %v", name, err)
```

pkg/operator/k8s_helpers.go
```diff
@@ -0,0 +1,176 @@
...
+	"k8s.io/client-go/tools/remotecommand"
+	"log"
+	"strconv"
+)
+
+type k8s struct {
+	clientset kubernetes.Interface
+}
+
+func GetK8SConfig() *k8s {
```
sleshchenko: as far as I see it returns `K8sClient` but not `K8sConfig`. Please consider renaming it.

eivantsov:
Thanks for reviews. I will take all of the comments into account when updating operator to 0.3.0 SDK which basically means a complete rewrite.

PR #15, opened 2019-04-22
Add tracing and metrics support

eivantsov:
This PR:

1. Adds metrics and tracing support (configured confimap with envs)
2. Improves create service function that accepts a slice of ports and names
3. Reworks service creation (service object is passed to create func rather than retrieved in the func itself)
4. Makes server pod request and limit configurable


slemeur:
Can we get the docs in place before we merge the PR please - so that everyone can easily leverage those nice capabilities.

eivantsov:
Closing this one. Will open a new one once clear requirements for metrics are in place.

PR #21, opened 2019-05-07
WIP Update default plugin registry URL

garagatyi:
Update default plugin registry URL since old value is no longer compatible with new wsmaster


eivantsov:
Will be merged when a new Che tag is released and the operator uses it in defaults.

davidfestal:
replaced by https://github.com/eclipse/che-operator/pull/28

PR #28, opened 2019-05-22
s/7.0.0-beta-4.0/7.0.0-beta-5.0/

nickboldt:
Change-Id:...

s/7.0.0-beta-4.0/7.0.0-beta-5.0/

Change-Id: Ic87e8875872390c1631b874942d90ac3bbfa2f9d


davidfestal:
@nickboldt @benoitf I've added the update of the plugin registry to v3 in commit https://github.com/eclipse/che-operator/pull/28/commits/92dc18f6619a4cf0e0e2698e832c6351debcbf8d

That was a change already proposed in PR https://github.com/eclipse/che-operator/pull/21 but since it is conflicting with more recent changes, we could simply close PR 21 and include the v3 change here, since it is indeed related to the upstream version upgrade.

PR #29, opened 2019-05-27
Add field for workspace cluster role to operator

johnmcollier:
Ports https://github.com/eclipse/che/pull/13178 to the Che operator

Adds an option for the user to specify an additional cluster role for workspaces.

This allows the user to create/use workspaces that require further roles than are provided by default (such as sidecar that needs to list/edit ConfigMaps, creating a service, etc).

**Note:** Some of the changes are from VS Code running `gofmt` upon save. I couldn't find a way to disable it.

### DIFF
--- a/deploy/crds/org_v1_che_cr.yaml
+++ b/deploy/crds/org_v1_che_cr.yaml
@@ -1,3 +1,4 @@
+
 #
 #  Copyright (c) 2012-2019 Red Hat, Inc.
 #    This program and the accompanying materials are made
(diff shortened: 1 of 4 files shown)


slemeur:
@davidfestal : Could you review the PR ?

davidfestal:
@slemeur yes, I'll have a look.

johnmcollier:
@davidfestal Sure, I can update che_controller_test.go to specify a cluster role.

I've tested on Docker for Desktop, OpenShift 3.11 and Minishift.

The recommended scenario to test this PR is to specify a cluster role in the cheWorkspaceClusterRole value (easiest one to verify is `cluster-admin`). Then verify that the `che-workspace-custom` role binding is created.

johnmcollier:
@davidfestal I've pushed up changes to `che_controller_test.go` so that the new field is properly tested.

davidfestal:
@johnmcollier Thanks !
@l0rd @nickboldt I assume this is not required for the CRW 1.2 release. Do you confirm ?

So we could wait for the next week to merge this and then integrate it into the next community Che 7 RC1 operators ?

nickboldt:
As you wish, @davidfestal. I've created a tag for 1.2.0.GA and a branch for 1.x so you can got to town on master branch (for Che 7 / CRW 2). 

I've also updated the sync job we use to push changes in che operator to the pkgs.devel repo that is used to build the operator container in Brew so that it uses branch = 1.x instead of master.

davidfestal:
Great ! Thanks @nickboldt  !

l0rd:
@davidfestal yes I exclude that we need this for CRW 1.2

PR #30, opened 2019-05-28
Fix Che operator to upgrade to 7.0.0-beta-5.0

davidfestal:
This PR fixes some problems encountered during the upgrade of the Che operator to the `7.0.0-beta-5.0` upstream Che release:
- `kcadm.sh` configuration could not be created on the read-only `/opt/jboss` folder in the Keycloak container when running under an arbitrary UID on Openshift.
- Postgres deployment on Kubernetes should contain a security context that avoids it to run as `root`, just as it is done on upstream deployment method: https://github.com/eclipse/che/blob/master/deploy/kubernetes/helm/che/custom-charts/che-postgres/templates/deployment.yaml#L26

This was  tested on both Openshift 4.1 (POC 4.1.0-rc.5) and Minikube 1.1.0

### DIFF
--- a/pkg/deploy/deployment_postgres.go
+++ b/pkg/deploy/deployment_postgres.go
@@ -20,14 +20,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func NewPostgresDeployment(cr *orgv1.CheCluster, chePostgresPassword string) *appsv1.Deployment {
+func NewPostgresDeployment(cr *orgv1.CheCluster, chePostgresPassword string, isOpenshift bool) *appsv1.Deployment {
 	chePostgresUser := util.GetValue(cr.Spec.Database.ChePostgresUser, "pgche")
 	chePostgresDb := util.GetValue(cr.Spec.Database.ChePostgresDb, "dbche")
 	postgresAdminPassword := util.GeneratePasswd(12)
 	postgresImage := util.GetValue(cr.Spec.Database.PostgresImage, DefaultPostgresImage)
 	name := "postgres"
 	labels := GetLabels(cr, name)
-	return &appsv1.Deployment{
+	deployment := appsv1.Deployment{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Deployment",
 			APIVersion: "apps/v1",
@@ -47,6 +47,8 @@ func NewPostgresDeployment(cr *orgv1.CheCluster, chePostgresPassword string) *ap
 					Labels: labels,
 				},
 				Spec: corev1.PodSpec{
+					SecurityContext: &corev1.PodSecurityContext{
+					},
 					Volumes: []corev1.Volume{
 						{
 							Name: name + "-data",
@@ -122,4 +124,12 @@ func NewPostgresDeployment(cr *orgv1.CheCluster, chePostgresPassword string) *ap
 			},
 		},
 	}
+	if ! isOpenshift {
+		var runAsUser int64 = 26
+		deployment.Spec.Template.Spec.SecurityContext = &corev1.PodSecurityContext {
+			RunAsUser: &runAsUser,
+			FSGroup: &runAsUser,
+		}
+	}
+	return &deployment
 }
--- a/pkg/deploy/deployment_postgres.go
+++ b/pkg/deploy/deployment_postgres.go
@@ -47,8 +47,6 @@ func NewPostgresDeployment(cr *orgv1.CheCluster, chePostgresPassword string, isO
 					Labels: labels,
 				},
 				Spec: corev1.PodSpec{
-					SecurityContext: &corev1.PodSecurityContext{
-					},
 					Volumes: []corev1.Volume{
 						{
 							Name: name + "-data",
--- a/pkg/deploy/exec_commands.go
+++ b/pkg/deploy/exec_commands.go
@@ -72,7 +72,7 @@ func GetKeycloakProvisionCommand(cr *orgv1.CheCluster, cheHost string) (command 
 	createRealmClientUserCommand := r.Replace(str)
 	command = createRealmClientUserCommand
 	if cheFlavor == "che" {
-		command = "cd /scripts && " +createRealmClientUserCommand
+		command = "cd /scripts && export JAVA_TOOL_OPTIONS=-Duser.home=. && " +createRealmClientUserCommand
 	}
 	return command
 }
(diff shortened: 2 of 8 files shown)


pkg/deploy/deployment_postgres.go
```diff
@@ -122,4 +124,12 @@ func NewPostgresDeployment(cr *orgv1.CheCluster, chePostgresPassword string) *ap
 			},
 		},
 	}
+	if ! isOpenshift {
```
davidfestal: @nickboldt Can you confirm that the CRW operator is only for Openshift, and not for pure Kubernetes ?

davidfestal: Seems yes: @rhopp just confirmed it to me.

davidfestal:
@nickboldt don't know why I cannot add you as a reviewer? Could you please review ?

pkg/deploy/exec_commands.go
```diff
@@ -72,7 +72,7 @@ func GetKeycloakProvisionCommand(cr *orgv1.CheCluster, cheHost string) (command
 	createRealmClientUserCommand := r.Replace(str)
 	command = createRealmClientUserCommand
 	if cheFlavor == "che" {
-		command = "cd /scripts && " +createRealmClientUserCommand
+		command = "cd /scripts && export JAVA_TOOL_OPTIONS=-Duser.home=. && " +createRealmClientUserCommand
```
nickboldt: Why do you need to set user.home=. ?

pkg/deploy/deployment_postgres.go
nickboldt: Correct, if you want pure k8s, you use Che.

pkg/deploy/deployment_postgres.go
```diff
@@ -20,14 +20,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func NewPostgresDeployment(cr *orgv1.CheCluster, chePostgresPassword string) *appsv1.Deployment {
+func NewPostgresDeployment(cr *orgv1.CheCluster, chePostgresPassword string, isOpenshift bool) *appsv1.Deployment {
```
nickboldt: I don't speak golang but will this break any existing implementations which don't have the bool param?

pkg/deploy/exec_commands.go
davidfestal: Let me explain a bit more:
- The Keycloak docker image for the `che` (=> upstream) case is [`eclipse/che-keycloak`](https://github.com/eclipse/che/blob/master/dockerfiles/keycloak/Dockerfile). And as you can see from the source code, it uses user `root`. So on Openshift, the used user becomes an arbitrary user. So finally the only place where the user has write access is the [`scripts` directory](https://github.com/eclipse/che/blob/master/dockerfiles/keycloak/Dockerfile#L16).
- The operator code tries to run the `kcadm.sh` CLI in the POD `keycloak` container through exec. But the problem is that by default the `kcadm` login process tries to create a file in: `$HOME/.keycloak/kcadm.config`... well more precisely in `<java user.home system property>//.keycloak/kcadm.config` since it's a Java process.
- This is why I add the reset of the `user.home` before running the various `kcadm` commands.

pkg/deploy/deployment_postgres.go
davidfestal: I'm not aware that the `NewPostgresDeployment` is used in any other repo than the `che-operator` repo. Did I miss smething ?
Is this code used a a library in some external repo ?

If not, it shouldn't hurt since I updated the [only one call site](https://github.com/eclipse/che-operator/pull/30/files#diff-d9445e614f0d728fd140771b0fd29c50L311).

Afaict, this method is public only because it was defined in a distinct package.

pkg/deploy/exec_commands.go
l0rd: @davidfestal The user should **not** be `root` but the user with `UID 1000`. At least that's how the base image [has been designed](https://github.com/jboss-dockerfiles/keycloak/blob/master/server/Dockerfile#L24). We are changing the user in our [Dockerfile](https://github.com/eclipse/che/blob/master/dockerfiles/keycloak/Dockerfile) to be able to `chown` and `chgrp` but we should revert it to user `1000` afterwards. I may miss something but it looks like an issue in our Dockerfile.

pkg/deploy/deployment_postgres.go
```diff
@@ -122,4 +124,12 @@ func NewPostgresDeployment(cr *orgv1.CheCluster, chePostgresPassword string) *ap
 			},
 		},
 	}
+	if ! isOpenshift {
+		var runAsUser int64 = 26
```
l0rd: Why 26?

pkg/deploy/deployment_postgres.go
l0rd: @davidfestal if I understand correctly you are doing the following assumption: if it's OpenShift the container is run as an unprivileged user, if it's not the container is run as root. Am I right?

pkg/deploy/deployment_postgres.go
davidfestal: To be consistent with upstream Che k8s helm charts: https://github.com/eclipse/che/blob/master/deploy/kubernetes/helm/che/custom-charts/che-postgres/templates/deployment.yaml#L26

pkg/deploy/exec_commands.go
davidfestal: Yes, I think the same way as well, and would probably open a PR on the `che-keycloak` docker image in the future. But in order to avoid any unmanaged side-effect of coming back to the `1000` user ID too quickly, I proposed this change as a temporary solution. Does it seem an acceptable temporary solution for you ?

l0rd: @davidfestal I am ok with doing that in a separate PR

pkg/deploy/deployment_postgres.go
l0rd: ok thanks and it looks consistent with the [upstream image user](https://github.com/sclorg/postgresql-container/blob/master/src/Dockerfile#L113)

pkg/deploy/deployment_postgres.go
davidfestal: Yes, that's mainly it. In fact I reproduced the rules that can be found in the [upstream Che deployment](https://github.com/eclipse/che/tree/master/deploy) for kubernetes and Openshift. But this might miss the case of Minishift installations with the `anyuid` plugin.

PR #37, opened 2019-06-19
Fix unnecessary cluster perm requirement

davidfestal:
This PR is to fix an unnecessary requirement of cluster roles for the `che-operator` running on Openshift 4 without Openshift OAuth.

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -108,15 +108,15 @@ func IsTestMode() (isTesting bool) {
 
 func GetClusterPublicHostname(isOpenShift4 bool) (hostname string, err error) {
 	if isOpenShift4 {
-		return GetClusterPublicHostnameForOpenshiftV4()
+		return getClusterPublicHostnameForOpenshiftV4()
 	} else {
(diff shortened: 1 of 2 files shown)


pkg/util/util.go
```diff
@@ -108,15 +108,15 @@ func IsTestMode() (isTesting bool) {
 
 func GetClusterPublicHostname(isOpenShift4 bool) (hostname string, err error) {
```
nickboldt: If you're refactoring `GetClusterPublicHostnameForOpenshiftV4` to `getClusterPublicHostnameForOpenshiftV4` should you also lowercase the `get` in this method name too?

davidfestal: in fact upper-case in Go means *public*, which is why I only kept this method with a upper-case.

nickboldt: ugh, that's about as non-obvious to a new user as syntactic whitespace in yaml. Not a fan. :(

davidfestal: not a fan either, TBH

PR #39, opened 2019-06-20
Update to the `7.0.0-RC-2.0` release after some last fixes

davidfestal:
This PR updates the default tags for images the Che and keycloak images to the `7.0.0-RC-2.0`.

But it also fixes a number of bugs found in the last round of tests performed for the RC 2 update:
- The Che operator did crash badly on Openshift V4 when enabling Openshift OAuth without the related cluster permissions. Implemented graceful failure => commit 34a9c62
- When disabling Openshift AOuth, the command run inside the Keycloak POD was failing due to file system permissions => commit 46803d8
- When enabling Openshift OAuth (=> adding the OS API certificate for Keycloak), the Keycloac server was not automatically restarted, and Openshift OAuth was failing => commit 7cc07ca
- After a rolling update of the Keycloak deployment, the controller reference was lost on the Keycloak deployment, so that it would not be deleted when deleting the `CheCluster` CR => commit e17c388
- When disabling Openshift OAuth, the related finalizer was not removed, which finally prevented the deletion of the `CheCluster` CR => commit e2d1801

### DIFF
--- a/pkg/controller/che/finalizer.go
+++ b/pkg/controller/che/finalizer.go
@@ -27,7 +27,7 @@ func (r *ReconcileChe) ReconcileFinalizer(instance *orgv1.CheCluster) (err error
 			if err := r.client.Delete(context.TODO(), oAuthClient); err != nil {
 				logrus.Errorf("Failed to delete %s oAuthClient: %s", oAuthClientName, err)
 				return err
-			}
+			}		
 			instance.ObjectMeta.Finalizers = util.DoRemoveString(instance.ObjectMeta.Finalizers, oAuthFinalizerName)
 			logrus.Infof("Updating %s CR", instance.Name)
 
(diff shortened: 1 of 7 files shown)


pkg/controller/che/finalizer.go
```diff
@@ -27,7 +27,7 @@ func (r *ReconcileChe) ReconcileFinalizer(instance *orgv1.CheCluster) (err error
 			if err := r.client.Delete(context.TODO(), oAuthClient); err != nil {
 				logrus.Errorf("Failed to delete %s oAuthClient: %s", oAuthClientName, err)
 				return err
-			}
+			}
```
skabashnyuk: I think this extra witespaces are not needed

pkg/controller/che/finalizer.go
```diff
@@ -40,3 +40,13 @@ func (r *ReconcileChe) ReconcileFinalizer(instance *orgv1.CheCluster) (err error
...
 }
+
+func (r *ReconcileChe) DeleteFinalizer(instance *orgv1.CheCluster) (err error) {
+	instance.ObjectMeta.Finalizers = util.DoRemoveString(instance.ObjectMeta.Finalizers, oAuthFinalizerName)
+	logrus.Infof("Removing OAuth finalizer on %s CR", instance.Name)
+	if err := r.client.Update(context.Background(), instance); err != nil {
+		logrus.Errorf("Failed to update %s CR: %s", instance.Name, err)
+		return err
+	}
+	return nil
```
skabashnyuk: new line

pkg/controller/che/finalizer.go
davidfestal: fixed

pkg/controller/che/finalizer.go
davidfestal: fixed

davidfestal:
@skabashnyuk I fixed your 2 comments. Is the rest OK for you ?

skabashnyuk:
yes

nickboldt:
If Sergii likes it, so do I. (I don't know enough about the changes in RC2 to be able to intelligently comment other than agreeing with the experts.)

PR #41, opened 2019-06-20
CRW-316 we don't need to keep the CRW...

nickboldt:
CRW-316 we don't need to keep the CRW controller-manifests in this repo

Change-Id: Ibb7c5b1695e527e5851c3a00f27f0548f553fa73


nickboldt:
In addition to this change the CRW operator sync job will now need to apply one extra sed patch:

```
sed -e "s@# *ADD controller-manifests /manifests@/ADD controller-manifests /manifests@g"
```

nickboldt:
> So the corresponding yaml files will be stored in the downstream repo already now ?

They already are: http://pkgs.devel.redhat.com/cgit/containers/codeready-workspaces-operator/tree/controller-manifests?h=crw-1.2-rhel-8

nickboldt:
Also cherrypicked to master: https://github.com/eclipse/che-operator/commit/80c96cc9be9fe6f2a0c24636c86dcf23b5d11aea

davidfestal:
Thanks @nickboldt !

PR #42, opened 2019-06-20
should we use an official image here,...

nickboldt:
should we use an official image here, instead of https://hub.docker.com/r/eivantsov/operator ? Is there a 6.19.x version somewhere?

Change-Id: I0146059407f787fb90cd2c7e345cbb40f9d106da


nickboldt:
Assuming this is a community version of the operator (Che 6.19.x, not CRW 1.x) then based on conversation w/ David today in Slack we probably don't have to do anything about this for CRW 1.2.2. Closing.

PR #48, opened 2019-07-16
Gracefully ignore the error when trying to delete an unexisting OAuthClient

davidfestal:
This PR is the fix for the issue https://github.com/eclipse/che/issues/13783.

It has been tested on OpenShift 4.1 with the Che operator.

### DIFF
--- a/pkg/controller/che/finalizer.go
+++ b/pkg/controller/che/finalizer.go
@@ -1,6 +1,7 @@
 package che
 
 import (
+	"k8s.io/apimachinery/pkg/api/errors"
 	"context"
 	orgv1 "github.com/eclipse/che-operator/pkg/apis/org/v1"
 	"github.com/eclipse/che-operator/pkg/util"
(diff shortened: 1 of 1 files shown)


pkg/controller/che/finalizer.go
```diff
@@ -20,14 +21,15 @@ func (r *ReconcileChe) ReconcileFinalizer(instance *orgv1.CheCluster) (err error
 			oAuthClientName := instance.Spec.Auth.OauthClientName
 			logrus.Infof("Custom resource %s is being deleted. Deleting oAuthClient %s first", instance.Name, oAuthClientName)
 			oAuthClient, err := r.GetOAuthClient(oAuthClientName)
-			if err != nil {
+			if err == nil {
+				if err := r.client.Delete(context.TODO(), oAuthClient); err != nil {
+					logrus.Errorf("Failed to delete %s oAuthClient: %s", oAuthClientName, err)
+					return err
+				}
+			} else if !errors.IsNotFound(err) {
```
l0rd: What if `err != nil` and `errors.IsNotFound(err)`?

l0rd: Maybe you should

```go
if errors.IsNotFound(err) {
  ...
} else {
  ...
}

Comment (davidfestal) [reply pkg/controller/che/finalizer.go]: if `err != nil` and `errors.IsNotFound(err)`, we:
- ignore the error
- don't try to delete the `OAuthClient` that doesn't exist
- continue normally (=> delete the finalizer and go on) 

in brief:
```go
if errors.IsNotFound(err) {
  // do nothing with OAuthClient and continue
}
```

PR #49, opened 2019-07-16
Add support for configuring devfile registry

metlos:
This is to support https://github.com/che-incubator/chectl/issues/114 (implemented in https://github.com/che-incubator/chectl/pull/189).

### DIFF
--- a/pkg/deploy/che_configmap.go
+++ b/pkg/deploy/che_configmap.go
@@ -14,12 +14,13 @@ package deploy
 import (
 	"encoding/json"
 	"fmt"
+	"os"
+
 	orgv1 "github.com/eclipse/che-operator/pkg/apis/org/v1"
 	"github.com/eclipse/che-operator/pkg/util"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"os"
 )
 
 func addMap(a map[string]string, b map[string]string) {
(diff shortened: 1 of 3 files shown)


l0rd:
That's related to endgame issue https://github.com/eclipse/che/issues/13557

davidfestal:
Successfully tested it on Openshift V4. LGTM.

pkg/deploy/che_configmap.go
```diff
@@ -71,7 +73,7 @@ func GetCustomConfigMapData() (cheEnv map[string]string) {
 		"CHE_INFRA_KUBERNETES_SERVICE__ACCOUNT__NAME":           "che-workspace",
 		"CHE_WORKSPACE_AUTO_START":                              "true",
 		"CHE_INFRA_KUBERNETES_WORKSPACE__UNRECOVERABLE__EVENTS": "FailedMount,FailedScheduling,MountVolume.SetUp failed,Failed to pull image",
-		"CHE_LIMITS_WORKSPACE_IDLE_TIMEOUT": "-1",
+		"CHE_LIMITS_WORKSPACE_IDLE_TIMEOUT":                     "-1",
```
nickboldt: weird indent here but not a showstopper

pkg/deploy/defaults.go
```diff
@@ -28,6 +28,7 @@ const (
 	DefaultIngressClass              = "nginx"
 	DefaultPluginRegistryUrl         = "https://che-plugin-registry.openshift.io"
 	DefaultUpstreamPluginRegistryUrl = "https://che-plugin-registry.openshift.io/v3"
+	DefaultDevfileRegistryUrl        = "https://che-devfile-registry.openshift.io"
```
nickboldt: will these eventually change to be part of the Che deployment, once the registries are local to the deployment rather than shared instances on openshift.io?

davidfestal: Yes, I will change that in a next PR

PR #50, opened 2019-07-16
Switch from `eclipse-che` org to `eclipse`

davidfestal:
This PR:
- Manages the move of the Che operator docker image from the `eclipse-che` Quay.io organization to the `eclipse` organization.
- Fixes some remaining small bugs, in the `update-nightly-olm-files.sh` script, that were discovered when using it to update the docker image version

### DIFF
--- a/olm/update-nightly-olm-files.sh
+++ b/olm/update-nightly-olm-files.sh
@@ -33,13 +33,26 @@ do
     echo "   - Updating new package version with roles defined in: ${role}"
     cp "$role" generated/current-role.yaml
     operator-sdk olm-catalog gen-csv --csv-version "${newNightlyPackageVersion}" --from-version="${lastPackageVersion}" 2>&1 | sed -e 's/^/      /'
+
+    containerImage=$(sed -n 's|^ *image: *\([^ ]*/che-operator:[^ ]*\) *|\1|p' "${packageFolderPath}/${newNightlyPackageVersion}/${packageName}.v${newNightlyPackageVersion}.clusterserviceversion.yaml")
+    createdAt=$(date -u +%FT%TZ)
+    echo "   - Updating new package version fields:"
+    echo "       - containerImage => ${containerImage}" 
+    echo "       - createdAt => ${createdAt}" 
+    sed \
+    -e "s|containerImage:.*$|containerImage: ${containerImage}|" \
+    -e "s/createdAt:.*$/createdAt: \"${createdAt}\"/" \
+    "${packageFolderPath}/${newNightlyPackageVersion}/${packageName}.v${newNightlyPackageVersion}.clusterserviceversion.yaml" \
+    > "${packageFolderPath}/${newNightlyPackageVersion}/${packageName}.v${newNightlyPackageVersion}.clusterserviceversion.yaml.new"
+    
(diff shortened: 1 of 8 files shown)


olm/update-nightly-olm-files.sh
```diff
@@ -33,13 +33,26 @@ do
...
+    
+    mv "${packageFolderPath}/${newNightlyPackageVersion}/${packageName}.v${newNightlyPackageVersion}.clusterserviceversion.yaml.new" \
+    "${packageFolderPath}/${newNightlyPackageVersion}/${packageName}.v${newNightlyPackageVersion}.clusterserviceversion.yaml"
+
   done
   echo "   - Copying the CRD file"
   cp "${packageFolderPath}/${lastPackageVersion}/eclipse-che-preview-${platform}.crd.yaml" "${packageFolderPath}/${newNightlyPackageVersion}/eclipse-che-preview-${platform}.crd.yaml"
   echo "   - Updating the 'nightly' channel with new version in the package descriptor: ${packageFilePath}"
-  echo "     (the previous one is saved with the .old suffix)"
   sed -e "s/${lastPackageVersion}/${newNightlyPackageVersion}/" "${packageFilePath}" > "${packageFilePath}.new"
```
nickboldt: if you use `-i` you can apply the sed transform in-place to the file, without having to redirect to .new then move .new to the actual filename.

`sed -i -e "s..." filename`

olm/update-nightly-olm-files.sh
```diff
@@ -33,13 +33,26 @@ do
...
+    createdAt=$(date -u +%FT%TZ)
+    echo "   - Updating new package version fields:"
+    echo "       - containerImage => ${containerImage}" 
+    echo "       - createdAt => ${createdAt}" 
+    sed \
+    -e "s|containerImage:.*$|containerImage: ${containerImage}|" \
+    -e "s/createdAt:.*$/createdAt: \"${createdAt}\"/" \
+    "${packageFolderPath}/${newNightlyPackageVersion}/${packageName}.v${newNightlyPackageVersion}.clusterserviceversion.yaml" \
+    > "${packageFolderPath}/${newNightlyPackageVersion}/${packageName}.v${newNightlyPackageVersion}.clusterserviceversion.yaml.new"
+
```
nickboldt: if you use `-i` you can apply the sed transform in-place to the file, without having to redirect to .new then move .new to the actual filename.

`sed -i -e "s..." filename`

davidfestal: Well, it seemed to me that the `-i` option doesn't work the same between MacOS and linux. But maybe I'm wrong.

olm/update-nightly-olm-files.sh
davidfestal: see the answer [here](https://github.com/eclipse/che-operator/pull/50/files#r303958738)

l0rd:
From an endgame point of view this is related to "Update release process to include new repositories"

davidfestal:
@nickboldt do you still want me to change the `sed` command to use in-place replacement ?
TBH I'd prefer merging this PR asap and possibly beautify the scripts in a next PR

PR #51, opened 2019-07-18
Include registries in Che deployment

davidfestal:


pkg/deploy/defaults.go
```diff
@@ -13,41 +13,47 @@
...
-	DefaultPvcJobsImage              = "registry.redhat.io/ubi8-minimal:8.0-127"
-	DefaultPvcJobsUpstreamImage      = "registry.access.redhat.com/ubi8-minimal:8.0-127"
-	DefaultPostgresImage             = "registry.redhat.io/rhscl/postgresql-96-rhel7:1-40"
-	DefaultPostgresUpstreamImage     = "centos/postgresql-96-centos7:9.6"
-	DefaultKeycloakImage             = "registry.redhat.io/redhat-sso-7/sso73-openshift:1.0-11"
-	DefaultKeycloakUpstreamImage     = "eclipse/che-keycloak:7.0.0-RC-2.0"
-	DefaultJavaOpts                  = "-XX:MaxRAMFraction=2 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10 " +
+	DefaultCheServerImageRepo           = "eclipse/che-server"
+	DefaultCodeReadyServerImageRepo     = "registry.redhat.io/codeready-workspaces/server-rhel8"
+	DefaultCheServerImageTag            = "7.0.0-RC-2.0"
```
nickboldt: this should be 7.0.0-rc-3.0 (lowercase rc, 3rd release)

pkg/deploy/defaults.go
```diff
@@ -13,41 +13,47 @@
...
+	DefaultCheFlavor                    = "che"
+	DefaultChePostgresUser              = "pgche"
+	DefaultChePostgresHostName          = "postgres"
+	DefaultChePostgresPort              = "5432"
+	DefaultChePostgresDb                = "dbche"
+	DefaultPvcStrategy                  = "common"
+	DefaultPvcClaimSize                 = "1Gi"
+	DefaultIngressStrategy              = "multi-host"
+	DefaultIngressClass                 = "nginx"
+	DefaultPluginRegistryImage          = "quay.io/eclipse/che-devfile-registry:7.0.0-RC-2.0"
```
nickboldt: same again. rc-3.0?

pkg/deploy/defaults.go
```diff
@@ -13,41 +13,47 @@
...
+	DefaultPluginRegistryPullPolicy     = "Always"
+	DefaultPluginRegistryMemoryLimit    = "32Mi"
+	DefaultPluginRegistryMemoryRequest  = "16Mi"
+	DefaultCodereadyPluginRegistryUrl   = "https://che-plugin-registry.openshift.io"
+	DefaultDevfileRegistryImage         = "quay.io/eclipse/che-devfile-registry:7.0.0-RC-2.0"
```
nickboldt: should this be rc-3.0?

pkg/deploy/ingress.go
```diff
@@ -27,11 +27,8 @@ func NewIngress(cr *orgv1.CheCluster, name string, serviceName string, port int)
 	}
 	ingressDomain := cr.Spec.K8SOnly.IngressDomain
 	ingressClass := util.GetValue(cr.Spec.K8SOnly.IngressClass, DefaultIngressClass)
-	labels := GetLabels(cr, util.GetValue(cr.Spec.Server.CheFlavor, DefaultCheFlavor))
+	labels := GetLabels(cr, name)
```
nickboldt: is name never equal to rh-sso or identityprovider? always "keycloak" ?

release-operator-code.sh
```diff
@@ -30,21 +30,27 @@ cd "${BASE_DIR}"
...
+lastDefaultDevfileRegistryVersion=$(grep 'DefaultDevfileRegistryImage' "pkg/deploy/defaults.go" | sed -e 's/.*DefaultDevfileRegistryImage *= *"[^":]*:\([^"]*\)"/\1/')
 if [ "${lastDefaultCheVersion}" != "${lastDefaultKeycloakVersion}" ]
 then
   echo "#### ERROR ####"
   echo "Current default Che version: ${lastDefaultCheVersion}"
   echo "Current default Keycloak version: ${lastDefaultKeycloakVersion}"
-  echo "Current default version for Che and keycloak are not the same in file 'pkg/deploy/defaults.go'."
+  echo "Current default Devfile Registry version: ${lastDefaultDevfileRegistryVersion}"
+  echo "Current default Plugin Registry version: ${lastDefaultPluginRegistryVersion}"
+  echo "Current default version for various Che containers are not the same in file 'pkg/deploy/defaults.go'."
```
nickboldt: why can't the default versions differ? what if ther's a 7.0.1 version of the plugin registry, and 7.0.3 version of the devfile registry, and che/keycloak are still on 7.0.0 ?

Or what happens in the CRW case there the version of SSO is 7.3.3 but the version of Che/CRW server is 2.0?

pkg/deploy/defaults.go
davidfestal: yeah, @benoitf noticed that and provided a PR to fix it, which I already merged.

release-operator-code.sh
davidfestal: > why can't the default versions differ? what if ther's a 7.0.1 version of the plugin registry, a 7.0.3 version of the devfile registry, and che/keycloak are still on 7.0.0 ?

Afaik, as concerning **upstream Che**, we expect the releases of `che-server`,  `che-keycloak`, `che-plugin-registry` and `che-devfile-registries` to be synchronized. That's the whole point of the new Che release process, isn't it @l0rd ?

> Or what happens in the CRW case there the version of SSO is 7.3.3 but the version of Che/CRW server is 2.0? I expect that in CRW we may release registry container updates more frequently (eg., weekly? once per sprint?) while CRW updates may be slower (every 6 weeks? every 12 weeks?) so these number will likely diverge.

This script is tor preparing a new release of **upstream Che** operator.
In the context of CRW 2.0, we might need to distinguish defaults between `che` and `codeready` flavours for registries as well (the same as we already did for other docker images). Then CRW releasing script would perform replacements in this file as it is convenient for the CRW-specific release process.

But I'd prefer letting this for a distinct PR.

pkg/deploy/ingress.go
davidfestal: no, it's always called `keycloak` according to the code.

pkg/deploy/defaults.go
davidfestal: For now I've not released the `rc-3` release of the operator. And I'd like to be consistent between releases of the underlying docker images and the release of the operator, especially when registries are included inside the deployment. (see my comment [here](https://github.com/eclipse/che-operator/pull/51#discussion_r305358374))

pkg/deploy/defaults.go
davidfestal: For now I've not released the `rc-3` release of the operator. And I'd like to be consistent between releases of the underlying docker images and the release of the operator, especially when registries are included inside the deployment. (see my comment [here](https://github.com/eclipse/che-operator/pull/51#discussion_r305358374))

davidfestal:
@nickboldt @l0rd is it OK for you now ?

release-operator-code.sh
nickboldt: Agreed, I was just making sure we're not boxed into a corner or forced to fork downstream. :)

davidfestal: Great. Thanks.

pkg/deploy/ingress.go
nickboldt: hmm. we did a big refactoring in CRW to replace keycloak with identityProvider so I was concerned we'd done and incomplete job over in the che-operator. if you're satisfied, so am I. (It's not ideal to always refer to the identityProvider as keycloak when it can be something else... but neither is it ideal to call the postgres image "postgres" when it can in theory be replaced with another DB - https://issues.jboss.org/browse/CRW-224 for example.

nickboldt: anyway refactoring can be a future fix for Che 8. :)

pkg/deploy/defaults.go
nickboldt: OK

release-operator-code.sh
l0rd: Yes for simplicity the versions of the servers (che, postgres, keycloak, registries) should always been in sync when releasing. 

If for instance we release a new version of che-server we will release a new version of che-devfile-registry even if there hasn't been any change on the registry side.

pkg/deploy/ingress.go
davidfestal: OK. Thanks.

PR #53, opened 2019-07-22
Fix installation defaults for Chectl and OLM files.  

davidfestal:
This PR fixes a number of inconsistencies between OLM files (CSVs) and the main custom resource example used by  `CheCtl` to start workspaces, so that installing from CheCtl, and from the OperatorHub would result in the same settings being applied.

### DIFF
--- a/deploy/crds/org_v1_che_cr.yaml
+++ b/deploy/crds/org_v1_che_cr.yaml
@@ -19,7 +19,11 @@ spec:
     # server image used in Che deployment
     cheImage: ''
     # tag of an image used in Che deployment
-    cheImageTag: ''
+    cheImageTag: 'nightly'
+    # image:tag used in Devfile registry deployment
+    devfileRegistryImage: 'quay.io/eclipse/che-devfile-registry:nightly'
+    # image:tag used in plugin registry deployment
+    pluginRegistryImage: 'quay.io/eclipse/che-plugin-registry:nighlty'
     # defaults to `che`. When set to `codeready`, CodeReady Workspaces is deployed
     # the difference is in images, labels, exec commands
     cheFlavor: ''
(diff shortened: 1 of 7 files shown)


deploy/crds/org_v1_che_cr.yaml
```diff
@@ -67,9 +71,9 @@ spec:
   storage:
     # persistent volume claim strategy for Che server. Can be common (all workspaces PVCs in one volume),
     # per-workspace (one PVC per workspace for all declared volumes) and unique (one PVC per declared volume). Defaults to common
-    pvcStrategy: ''
+    pvcStrategy: 'per-workspace'
```
sleshchenko: The line above says - defaults to common.
And `defaults.go` actually has `common` as a default value https://github.com/eclipse/che-operator/blob/master/pkg/deploy/defaults.go#L32

There is an issue to use common PVC strategy as default https://github.com/eclipse/che/issues/14614

@davidfestal @l0rd Seems common should be here

davidfestal: @sleshchenko I assume I mainly kept what was set initially in the operator by Eugene, possibly in his initial CSV examples. TBH I don't remember. @l0rd wdyt ?

l0rd: I would say +1 for common but by default the namespace strategy is still one distinct per namespace per workspace right? How does common PVC strategy work with per-workspace namespace strategy?

sleshchenko: > I would say +1 for common but by default the namespace strategy is still one distinct per namespace per workspace right?

It depends on OAuth, if it's configured - <workspace-id> is used, if it's not workspaces are run in the same namespace as Che Server.

> How does common PVC strategy work with per-workspace namespace strategy?

Combination of:
- common PVC + per workspace namespace
- per workspace PVC + per workspace namespace

Works pretty the same the only difference PVC name that will be used - `claim-che-workspace-<workspaceid>`(for per-workspace) or `claim-che-workspace`(for common).

sleshchenko: The PVC strategy is changed in PR that introduce <username>-che namespace by default https://github.com/eclipse/che-operator/pull/92/commits/651c65184dd0caef6542974bf6910a087e484048
Probably OLM should be changed as well. Till add the corresponding TODO item in that PR

PR #57, opened 2019-07-25
Allow configuring pull policy (with `Always` by default for nightlies)

davidfestal:
Currently, the pull policy for the Che, Keycloak and Postgres containers is fixed and set at `IfNotPresent`.
Obviously this make the use of the `nightly` operator (and nightly Che components) useless.

This PR:
- adds fields in the `CheCluster` custom resource to drive the image pull policy of the various Che components (Che, Keycloak, etc...)
- If those fields are not set (or set to empty string) then a default pull policy is applied according to the following rule:
  - use `Always` if the related docker image tag is `latest` or `nightly`
  - use `IfNotPresent` in other cases

Please consider this PR as a followup of issue https://github.com/eclipse/che/issues/13780 that was part of endgame plan.

### DIFF
--- a/pkg/apis/org/v1/che_types.go
+++ b/pkg/apis/org/v1/che_types.go
@@ -12,8 +12,8 @@
 package v1
 
 import (
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
@@ -35,6 +35,8 @@ type CheClusterSpecServer struct {
 	CheImage string `json:"cheImage"`
 	// CheImageTag is a tag of an image used in Che deployment
 	CheImageTag string `json:"cheImageTag"`
+	// CheImagePullPolicy is the image pull policy used in Che registry deployment: default value is Always
+	CheImagePullPolicy corev1.PullPolicy `json:"cheImagePullPolicy"`
 	// CheFlavor is an installation flavor. Can be 'che' - upstream or 'codeready' - CodeReady Workspaces. Defaults to 'che'
 	CheFlavor string `json:"cheFlavor"`
 	// CheHost is an env consumer by server. Detected automatically from Che route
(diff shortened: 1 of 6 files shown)


pkg/deploy/defaults.go
```diff
@@ -27,12 +31,10 @@ const (
 	DefaultIngressStrategy              = "multi-host"
 	DefaultIngressClass                 = "nginx"
 	DefaultPluginRegistryImage          = "quay.io/eclipse/che-plugin-registry:7.0.0-RC-2.0"
-	DefaultPluginRegistryPullPolicy     = "Always"
```
nickboldt: do you need to also set DefaultXXPullPolicy for Che and identityProvider images?

davidfestal: I don't use default values for pull policy, because the default is calculated based on the image tag.

nickboldt: ah, of course. I was reading this in reverse (as if you ADDED two PullPolicy values rather than removing them). +1

pkg/apis/org/v1/che_types.go
```diff
@@ -143,6 +147,8 @@ type CheClusterSpecAuth struct {
 	OauthSecret string `json:"oAuthSecret"`
 	// KeycloakImage is image:tag used in Keycloak deployment
 	KeycloakImage string `json:"identityProviderImage"`
+	// KeycloakImagePullPolicy is the image pull policy used in Keycloak registry deployment: default value is Always
+	KeycloakImagePullPolicy corev1.PullPolicy `json:"identityProviderImagePullPolicy"`
```
nickboldt: it's too bad we have Keycloak* varibles mixed in with identityProvider* images but that's a refactoring for another time :)

davidfestal: Yeah, the problem here is that it would be a breaking change in the API. So we would have to manage several versions of the API, which is not straightforward afaik.

pkg/deploy/defaults.go
```diff
@@ -57,3 +59,15 @@ const (
 	DefaultSecurityContextFsGroup       = "1724"
 	DefaultSecurityContextRunAsUser     = "1724"
 )
+
+func DefaultPullPolicyFromDockerImage(dockerImage string) string {
```
l0rd: I am surprised about that @nickboldt didn't comment about your usage of the `docker` word here :-)

davidfestal: haha. He still can do a PR afterwards if he prefers other words :-)

nickboldt:
Once this is merged, I expect to see Always as the pull policy here:

image

(Screenshot taken after doing a chectl operator deployment of che 7 nightly, with custom images. See https://github.com/redhat-developer/codeready-workspaces/blob/master/devdoc/building/che7-custom-resource-airgap.adoc for steps to reproduce.)

davidfestal:
Yes, exactly.

davidfestal:
If you use chectl, you can even have this today by overriding the `che-operator-image` to `quay.io/dfestal/che-operator:add-pull-policies` I assume

nickboldt:
So is this being merged for 7 RC5 / GA?

PR #58, opened 2019-07-25
remove DefaultCodereadyPluginRegistryUrl...

nickboldt:
remove `DefaultCodereadyPluginRegistryUrl` from pkg/deploy/defaults.go and `pluginRegistryUrl` from deploy/crds/org_v1_che_cr.yaml as these are no longer used/needed

Change-Id: I8bf17e40d7ed55c30fe28f640614fe6202cf66a5

### DIFF
--- a/pkg/deploy/defaults.go
+++ b/pkg/deploy/defaults.go
@@ -30,7 +30,6 @@ const (
 	DefaultPluginRegistryPullPolicy     = "Always"
 	DefaultPluginRegistryMemoryLimit    = "32Mi"
 	DefaultPluginRegistryMemoryRequest  = "16Mi"
-	DefaultCodereadyPluginRegistryUrl   = "https://che-plugin-registry.openshift.io"
 	DefaultDevfileRegistryImage         = "quay.io/eclipse/che-devfile-registry:7.0.0-RC-2.0"
 	DefaultDevfileRegistryPullPolicy    = "Always"
 	DefaultDevfileRegistryMemoryLimit   = "32Mi"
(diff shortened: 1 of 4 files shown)


nickboldt:
SHould we also remove `DevfileRegistryUrl` and `PluginRegistryUrl` from `pkg/apis/org/v1/che_types.go` ?

pkg/deploy/defaults.go
```diff
@@ -30,7 +30,6 @@ const (
 	DefaultPluginRegistryPullPolicy     = "Always"
 	DefaultPluginRegistryMemoryLimit    = "32Mi"
 	DefaultPluginRegistryMemoryRequest  = "16Mi"
-	DefaultCodereadyPluginRegistryUrl   = "https://che-plugin-registry.openshift.io"
```
davidfestal: In fact I had left this line for the Che 6 CRW case, where plugin registry was already used. But since we now manage Che 6 from a distinct branch, I assume we're OK to remove this line here. However did you check this compiles correctly after removing this line ?

davidfestal:
> Should we also remove `DevfileRegistryUrl` and `PluginRegistryUrl` from `pkg/apis/org/v1/che_types.go` ?

Now users should still have the ability to override the URLs in the Custom Resource.

> Or just update the comment to set new correct default URLs?

We should just update the comment to remove the default.

> And what about `pkg/deploy/che_configmap.go` ?

I assume that in `che_configmap.go`, we should remove the following lines:
```go
	if pluginRegistryUrl == "" && cheFlavor == "codeready" {
		pluginRegistryUrl = DefaultCodereadyPluginRegistryUrl
	}
```
That should be sufficient to have successful builds

pkg/deploy/defaults.go
nickboldt: nope, wanted to discuss how much to change first :)

nickboldt:
Built OK locally in docker image. Your suggestions forcepushed to the PR. What do you think now, @davidfestal  ?

PR #62, opened 2019-08-01
docs(olm): information for generating nightlies

benoitf:
Adds howto

### DIFF
--- a/olm/README.md
+++ b/olm/README.md
@@ -3,7 +3,7 @@
 OLM packages scripts are using some required dependencies that need to be installed
  - [curl](https://curl.haxx.se/)
  - [https://github.com/kislyuk/yq](https://github.com/kislyuk/yq) and not [http://mikefarah.github.io/yq/](http://mikefarah.github.io/yq/)
- - [Operator SDK v0.8.2](https://github.com/operator-framework/operator-sdk/blob/master/doc/user/install-operator-sdk.md)
+ - [Operator SDK v0.8.2](https://github.com/operator-framework/operator-sdk/blob/v0.8.2/doc/user/install-operator-sdk.md)
 
 If these dependencies are not installed, `docker-run.sh` can be used as a container bootstrap to run a given script with the appropriate dependencies.
 
--- a/olm/README.md
+++ b/olm/README.md
@@ -1,0 +1,31 @@
+# Pre-Requisites
+
+OLM packages scripts are using some required dependencies that need to be installed
+ - [curl](https://curl.haxx.se/)
+ - [https://github.com/kislyuk/yq](https://github.com/kislyuk/yq) and not [http://mikefarah.github.io/yq/](http://mikefarah.github.io/yq/)
+ - [Operator SDK v0.8.2](https://github.com/operator-framework/operator-sdk/blob/master/doc/user/install-operator-sdk.md)
+
+If these dependencies are not installed, `docker-run.sh` can be used as a container bootstrap to run a given script with the appropriate dependencies.
+
+Example : `$ docker-run.sh update-nightly-olm-files.sh`
+
+
+# Make new changes to OLM artifacts
+
+Every change needs to be done in a new OLM artifact as previous artifacts are frozen.
+
+A script is generating new folders/files that can be edited.
+
+In `olm` folder
+
+- If all dependencies are installed on the system:
+``` 
+$ update-nightly-olm-files.sh
+```
+
+- To use a docker environment
+``` 
+$ docker-run.sh update-nightly-olm-files.sh
+```
+
+Then the changes can be applied in the newly created CSV files.
--- a/olm/docker-run.sh
+++ b/olm/docker-run.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+
+# git ROOT directory used to mount filesystem
+GIT_ROOT_DIRECTORY=$(git rev-parse --show-toplevel)
+
+# Container image
+IMAGE_NAME="eclipse/che-operator-olm-build"
+
+# Operator SDK
+OPERATOR_SDK_VERSION=v0.8.2
+
+init() {
+  BLUE='\033[1;34m'
+  GREEN='\033[0;32m'
+  RED='\033[0;31m'
+  NC='\033[0m'
+  BOLD='\033[1m'
+}
+
+check() {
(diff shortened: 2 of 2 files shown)


olm/README.MD
```diff
@@ -0,0 +1,29 @@
+# Creating new OLM artifacts
+
+Script `update-nightly-olm-files.sh` is used for that
+
+Pre-requisites
+ - [yq/https://pypi.org/project/yq](https://pypi.org/project/yq) and not [http://mikefarah.github.io/yq/](http://mikefarah.github.io/yq/)
+ 
+      $ brew install python-yq (on macOS)
```
nickboldt: why call out the MacOS install instructions? I mean you could also mention `pip install yq jq` too if you want to cover another 'nix method. :)

nickboldt: Also just a nitpick but should the filename be README.md to be consistent w/ the one in the root of this project?

benoitf: I can remove macOS instruction if you want

BTW why jq if only yq is required ?

benoitf: fixed

nickboldt: "Before using yq, you also have to install its dependency, jq. See the jq installation instructions for details and directions specific to your platform." - https://pypi.org/project/yq/

olm/README.md
```diff
@@ -0,0 +1,27 @@
+# Creating new OLM artifacts
+
+Script `update-nightly-olm-files.sh` is used for that
+
+Pre-requisites
+ - [yq/https://pypi.org/project/yq](https://pypi.org/project/yq) and not [http://mikefarah.github.io/yq/](http://mikefarah.github.io/yq/)
```
nickboldt: link label is "yq/https://pypi.org/project/yq" ... perhaps you just want "yq" ?

benoitf:
it seems that you prefer no documentation so I'm closing

nickboldt:
Uh, I never said we shouldn't have docs. I just asked a couple nitpick questions.

olm/README.md
davidfestal: Or `https://pypi.org/project/yq` simply ?

olm/README.md
```diff
@@ -0,0 +1,27 @@
...
+It can be done in a docker environment:
+
+from the root cloned directory of che-operator:
+```bash
+$ cd git/eclipse/che-operator
+$ docker run --rm -it -v $(pwd):/workdir --entrypoint=sh evns/yq
+# now in docker context
+$ apk update
+$ apk add curl bash
+$ RELEASE_VERSION=v0.9.0
```
davidfestal: Could you use version `v0.8.2` instead of `v9.0.0` (which changes the way it manages OLM CSV and package files) ?

olm/README.md
```diff
@@ -0,0 +1,27 @@
+# Creating new OLM artifacts
+
+Script `update-nightly-olm-files.sh` is used for that
+
+Pre-requisites
+ - [yq/https://pypi.org/project/yq](https://pypi.org/project/yq) and not [http://mikefarah.github.io/yq/](http://mikefarah.github.io/yq/)
+ - [Operator SDK](https://github.com/operator-framework/operator-sdk/blob/master/doc/user/install-operator-sdk.md)
```
davidfestal: Could you mention that we should install `v0.8.2` instead of `v0.9.0` ?

benoitf: yes but usually people use the latest version of all products

olm/README.md
benoitf: yes

benoitf:
you want a docker run script directly ?
I assume then it would be more efficient if an image is provided with  all dependencies instead of installing it at each run. But where to put this Dockerfile ?

benoitf:
ok guys, I've reworked everything and now provide `docker-run.sh` script and use operator-sdk v0.8.2

benoitf:
@nickboldt please review

olm/README.md
```diff
@@ -0,0 +1,31 @@
+# Pre-Requisites
+
+OLM packages scripts are using some required dependencies that need to be installed
+ - [curl](https://curl.haxx.se/)
+ - [https://github.com/kislyuk/yq](https://github.com/kislyuk/yq) and not [http://mikefarah.github.io/yq/](http://mikefarah.github.io/yq/)
+ - [Operator SDK v0.8.2](https://github.com/operator-framework/operator-sdk/blob/master/doc/user/install-operator-sdk.md)
```
nickboldt: If we're locked down to 0.787.2, maybe we should link to https://github.com/operator-framework/operator-sdk/blob/v0.8.2/doc/user/install-operator-sdk.md instead of master = 0.9.0 ?

benoitf: fixed

olm/docker-run.sh
```diff
@@ -0,0 +1,77 @@
...
+    printf "%bError: %bscript %b provided is not existing. Command is $ docker-run.sh <script-to-run> [optional-arguments-of-script-to-run]\n" "${RED}" "${NC}" "${1}"
+    exit 1
+  fi
+}
+
+# Build image
+build() {
+  printf "%bBuilding image %b${IMAGE_NAME}${NC}..." "${BOLD}" "${BLUE}"
+  if docker build --build-arg OPERATOR_SDK_VERSION=${OPERATOR_SDK_VERSION} -t ${IMAGE_NAME} > docker-build-log 2>&1 -<<EOF
+  FROM evns/yq
```
nickboldt: I don't love that we have a nested dependency on an alpine image here, since it's YET ANOTHER place where I'll have to fork this code for the product (assuming this is needed for the CRW operator?).

Could we use some variation on https://github.com/nickboldt/containers/tree/master/ubi8-yq -> https://quay.io/repository/nickboldt/ubi8-yq instead of https://github.com/evnsio/yq -> https://hub.docker.com/r/evns/yq

benoitf: but for CRW you can install stuff directly, no need to use docker wrapper

nickboldt: OK, that's fair.

So... does it make sense that we start using ubi-minimal base images instead of alpine ones, in order to encourage more people to try the RHEL8 experience?

benoitf: IMHO images are too big:
```
quay.io/nickboldt/ubi8-yq                                 latest              00af3dd6c8ed        19 hours ago        148MB
mikefarah/yq                                              latest              86d84d6c94bc        2 months ago        14.9MB
evns/yq                                                   latest              a5a8e69c4986        17 months ago       54.1MB
```

benoitf: 150MB to a single binary

nickboldt: hmm. https://quay.io/repository/nickboldt/ubi8-yq?tab=tags says it's 50M. Does Quay lie, or are we looking at packed vs. unpacked sizes?

BTW the mikefarah version of yq is the golang implementation, not the python wrapper for `jq`, so we can't really compare that one :)

nickboldt: ANYWAY if this is just something for Che 7 which productization doesn't have to worry about, then +1 for this PR.

PR #63, opened 2019-08-02
Fix Custom Resource update bugs and external registry URLs not taken in account

davidfestal:
This PR fixes a number of weirds behaviors / hard bugs happening when the `CheCluster` Custom Resource is updated.

It fixes issue https://github.com/eclipse/che/issues/14111

### DIFF
--- a/pkg/controller/che/che_controller.go
+++ b/pkg/controller/che/che_controller.go
@@ -470,20 +470,19 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 		return nil, nil
 	}
 
+	pluginRegistryURL := instance.Spec.Server.PluginRegistryUrl
 	// Create Plugin registry resources unless an external registry is used
 	externalPluginRegistry := instance.Spec.Server.ExternalPluginRegistry
 	if !externalPluginRegistry {
-		pluginRegistryURL, err := addRegistryRoute("plugin")
+		guessedPluginRegistryURL, err := addRegistryRoute("plugin")
 		if err != nil {
 			return reconcile.Result{}, err
 		}
 		if cheFlavor != "codeready" {
-			pluginRegistryURL += "/v3"
+			guessedPluginRegistryURL += "/v3"
 		}
-		instance.Status.PluginRegistryURL = pluginRegistryURL
-		if err := r.UpdateCheCRStatus(instance, "status: Plugin Registry URL", pluginRegistryURL); err != nil {
-			instance, _ = r.GetCR(request)
-			return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 1}, err
+		if pluginRegistryURL == "" {
+			pluginRegistryURL = guessedPluginRegistryURL
 		}
 
 		pluginRegistryImage := util.GetValue(instance.Spec.Server.PluginRegistryImage, deploy.DefaultPluginRegistryImage)
@@ -499,18 +498,22 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 			return *result, err
 		}
 	}
+	instance.Status.PluginRegistryURL = pluginRegistryURL
+	if err := r.UpdateCheCRStatus(instance, "status: Plugin Registry URL", pluginRegistryURL); err != nil {
+		instance, _ = r.GetCR(request)
+		return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 1}, err
+	}
 
+	devfileRegistryURL := instance.Spec.Server.DevfileRegistryUrl
 	// Create devfile registry resources unless an external registry is used
 	externalDevfileRegistry := instance.Spec.Server.ExternalDevfileRegistry
 	if !externalDevfileRegistry {
-		devfileRegistryURL, err := addRegistryRoute("devfile")
+		guessedDevfileRegistryURL, err := addRegistryRoute("devfile")
 		if err != nil {
 			return reconcile.Result{}, err
 		}
-		instance.Status.DevfileRegistryURL = devfileRegistryURL
-		if err := r.UpdateCheCRStatus(instance, "status: Devfile Registry URL", devfileRegistryURL); err != nil {
-			instance, _ = r.GetCR(request)
-			return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 1}, err
+		if devfileRegistryURL == "" {
+			devfileRegistryURL = guessedDevfileRegistryURL
 		}
 
 		devfileRegistryImage := util.GetValue(instance.Spec.Server.DevfileRegistryImage, deploy.DefaultDevfileRegistryImage)
@@ -525,6 +528,11 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 		if err != nil || result != nil {
 			return *result, err
 		}
+	}
+	instance.Status.DevfileRegistryURL = devfileRegistryURL
+	if err := r.UpdateCheCRStatus(instance, "status: Devfile Registry URL", devfileRegistryURL); err != nil {
+		instance, _ = r.GetCR(request)
+		return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 1}, err
 	}
 
 	// create Che service and route
--- a/pkg/controller/che/che_controller.go
+++ b/pkg/controller/che/che_controller.go
@@ -448,7 +448,14 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 					return &reconcile.Result{Requeue: true, RequeueAfter: time.Second * 5}, err
 				}
 			}
-			if effectiveDeployment.Spec.Template.Spec.Containers[0].Image != registryImage {
+			actualMemRequest := effectiveDeployment.Spec.Template.Spec.Containers[0].Resources.Requests[corev1.ResourceMemory]
+			actualMemLimit := effectiveDeployment.Spec.Template.Spec.Containers[0].Resources.Limits[corev1.ResourceMemory]
+			limitStr := actualMemLimit.String()
+			requestStr := actualMemRequest.String()
+			if effectiveDeployment.Spec.Template.Spec.Containers[0].Image != registryImage ||
+				registryMemoryRequest != requestStr ||
+				registryMemoryLimit != limitStr ||
+				effectiveDeployment.Spec.Template.Spec.Containers[0].ImagePullPolicy != registryImagePullPolicy {
 				newDeployment := deploy.NewRegistryDeployment(
 					instance,
 					registryType,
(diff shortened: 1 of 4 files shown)


nickboldt:
Not sure I'm qualified to review this one.

davidfestal:
@l0rd Do you plan to review it finally ?

rhopp:
@davidfestal I've tried this and seems to be working reasonably well with some exceptions (I'm not sure all are related to this PR)
1) After first CR edit, I'm getting `time="2019-08-06T14:44:00Z" level=info msg="Updating deployment che with new memory settings. Request: 512Mi, limit: 1Gi" ` every few seconds - the `che` deployment is edited every few seconds.
2) Changing `externalDevfileRegistry` to `true` from `false` does use external registry, but the local registry is not undeployed - I guess it should be? This is not blocking 7.0 anyway.
3) Changing `externalDevfileRegistry` to `false` from `true`, but leaving `devfileRegistryUrl` not empty does nothing. Only deleting `devfileRegistryUrl` edits the CM and redeploys che-server.

I didn't have enough time to try all possible scenarios, but I think at least 1) and 3) should be addressed.

Sidenote - This is the type of PR which would be great, if it was covered by tests... We can create a separate issue for this.

pkg/controller/che/che_controller.go
```diff
@@ -470,20 +477,19 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
 	externalPluginRegistry := instance.Spec.Server.ExternalPluginRegistry
 	if !externalPluginRegistry {
-		pluginRegistryURL, err := addRegistryRoute("plugin")
+		guessedPluginRegistryURL, err := addRegistryRoute("plugin")
 		if err != nil {
 			return reconcile.Result{}, err
 		}
 		if cheFlavor != "codeready" {
-			pluginRegistryURL += "/v3"
+			guessedPluginRegistryURL += "/v3"
```
nickboldt: Should this be externalized as a property somehow? like a default registry version variable, rather than encoded this way?

pkg/controller/che/che_controller.go
```diff
@@ -470,20 +477,19 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
 		}
-		instance.Status.PluginRegistryURL = pluginRegistryURL
-		if err := r.UpdateCheCRStatus(instance, "status: Plugin Registry URL", pluginRegistryURL); err != nil {
-			instance, _ = r.GetCR(request)
-			return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 1}, err
+		if pluginRegistryURL == "" {
+			pluginRegistryURL = guessedPluginRegistryURL
```
nickboldt: Does it matter that you're no longer throwing a status report here?

pkg/controller/che/che_controller.go
```diff
@@ -499,18 +505,24 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 			return *result, err
 		}
 	}
+	if pluginRegistryURL != instance.Status.PluginRegistryURL {
+		instance.Status.PluginRegistryURL = pluginRegistryURL
+		if err := r.UpdateCheCRStatus(instance, "status: Plugin Registry URL", pluginRegistryURL); err != nil {
```
nickboldt: Ah, I see the block deleted from line 491 has moved down here. So this is the fallback value if the guess doesn't work?

pkg/controller/che/che_controller.go
```diff
@@ -448,7 +448,14 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 					return &reconcile.Result{Requeue: true, RequeueAfter: time.Second * 5}, err
 				}
 			}
-			if effectiveDeployment.Spec.Template.Spec.Containers[0].Image != registryImage {
+			actualMemRequest := effectiveDeployment.Spec.Template.Spec.Containers[0].Resources.Requests[corev1.ResourceMemory]
+			actualMemLimit := effectiveDeployment.Spec.Template.Spec.Containers[0].Resources.Limits[corev1.ResourceMemory]
+			limitStr := actualMemLimit.String()
+			requestStr := actualMemRequest.String()
+			if effectiveDeployment.Spec.Template.Spec.Containers[0].Image != registryImage ||
```
nickboldt: why do we need these extra checks here?

pkg/controller/che/che_controller.go
```diff
@@ -720,32 +729,32 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 		cheImageRepo = util.GetValue(instance.Spec.Server.CheImage, deploy.DefaultCodeReadyServerImageRepo)
 		cheImageTag = util.GetValue(instance.Spec.Server.CheImageTag, deploy.DefaultCodeReadyServerImageTag)
 	}
-	cheDeployment, err := deploy.NewCheDeployment(instance, cheImageRepo, cheImageTag, cmResourceVersion, isOpenShift)
+	cheDeploymentToCreate, err := deploy.NewCheDeployment(instance, cheImageRepo, cheImageTag, cmResourceVersion, isOpenShift)
```
nickboldt: you have variables called `newKeycloakDeployment` and `cheDeploymentToCreate` ... assuming these are both new things to be created, should the names be more consistent? eg., both prefix with `new` or suffix with `ToCreate` ?

Also any chance we can call it "IdentityProvider" instead of "Keycloak" ? :) Or is that a bad refactoring to do piecemeal, which we should tackle in Che 8.0 ?

pkg/controller/che/che_controller.go
```diff
@@ -861,14 +870,18 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
 	limitStr := actualMemLimit.String()
 	requestStr := actualMemRequest.String()
 	desiredRequest := util.GetValue(instance.Spec.Server.ServerMemoryRequest, deploy.DefaultServerMemoryRequest)
 	desiredLimit := util.GetValue(instance.Spec.Server.ServerMemoryLimit, deploy.DefaultServerMemoryLimit)
-	if desiredRequest != requestStr || desiredLimit != limitStr {
+	desiredImagePullPolicy := util.GetValue(string(instance.Spec.Server.CheImagePullPolicy), deploy.DefaultPullPolicyFromDockerImage(cheImageRepo + ":" + cheImageTag))
+	if desiredRequest != requestStr ||
+		desiredLimit != limitStr ||
+		string(effectiveCheDeployment.Spec.Template.Spec.Containers[0].ImagePullPolicy) != desiredImagePullPolicy ||
+		(r.GetDeploymentEnv(effectiveCheDeployment, "CHE_SELF__SIGNED__CERT") != "") != instance.Spec.Server.SelfSignedCert {
```
nickboldt: these last 2 checks are new. Why do we need them?

pkg/controller/che/che_controller.go
davidfestal: this could probably be externalized, but since it was already like that before this PR, I would let this for a next PR, no ?

pkg/controller/che/che_controller.go
davidfestal: The status report is thrown in the same way, but just some lines later, where this block has indeed been moved. because we have to update the registry URL on the status in both cases (when using embedded registry or when an external registry URL is specified)

pkg/controller/che/che_controller.go
davidfestal: more precisely, if an external URL is provided we use it (even if the embedded registry is deployed => a guessed value is available, but finally not used).

pkg/controller/che/che_controller.go
davidfestal: To identify situations when caracteristics of the current deployment are not the same as those required by the custom resource. In such a situation, we should redeploy => restart the POD.

pkg/controller/che/che_controller.go
davidfestal: I already tried to clean naming to make the code easier to read / understand in the most critical parts. But I might have forgotten some of left some inconsistencies. Is it a blocker for you ?

davidfestal: As for the `Keycloak` to `ìdentityProvider` change, I'd prefer doing that in a next PR, probably in the context of the wider refactoring that will certainly be required after GA.

pkg/controller/che/che_controller.go
davidfestal: Because these values are used in the Che deployment, so when they are changed, the deployment should be updated and the underlying POD restarted. It was a bug that they were not.

davidfestal:
> After first CR edit, I'm getting time="2019-08-06T14:44:00Z" level=info msg="Updating deployment che with new memory settings. Request: 512Mi, limit: 1Gi" every few seconds - the che deployment is edited every few seconds.

How many times is the deployment edited, and how long is it regularly edited ? Does it stop after some time ?

pkg/controller/che/che_controller.go
nickboldt: not a blocker. +1 for wider refactoring in another PR.

pkg/controller/che/che_controller.go
nickboldt: +1

davidfestal:
> Changing externalDevfileRegistry to false from true, but leaving devfileRegistryUrl not empty does nothing. Only deleting devfileRegistryUrl edits the CM and redeploys che-server.

That doesn't seem a problem to me: as soon as you have `devfileRegistryURL` set to a non-empty value, the value is used in the Che config map, even if `externalDevfileRegistry` is `false`.
We don't enfore consistency between those 2 properties, and the URL property setting has the priority.

pkg/controller/che/che_controller.go
nickboldt: yeah, save it for later refactoring. I'll assume you'll open an issue for externalizing params and refactoring "keycloak" to "identityProvider" and target it for Che 7.2 or later.

davidfestal:
> Changing externalDevfileRegistry to true from false does use external registry, but the local registry is not undeployed - I guess it should be?

Yes, it's a known limitation.

> This is not blocking 7.0 anyway

You're right, since implementation would have required more complex and impacting changes, I prefered letting this for after GA.

rhopp:
@davidfestal Here's what I did:
* `chectl server:start -a operator -p openshift -b crw.codeready.com -n rhopp-test --che-operator-image=quay.io/rhopp/che-operator:cr-update-bugs`  (che-operator:cr-update-bugs is image I built from this PR)
* I waited until the default deployment finishied successfully
* Then I changed two values in checluster CR:
* * `spec.server.tlsSupport: false` to `spec.server.tlsSupport: true`
* * `spec.server.selfSignedCert: false` to `spec.server.selfSignedCert: true`
* Then I let the operator do it's business.
* Once operator finished with creating routes, redeploying che etc (I don't exactly know what everything is it doing) - At this point I have usable "secure" che.

But when I look at che-operator logs, I can see, that every 5 seconds, it is doing something wich che deployment - this seems like a bug which needs to be fixed before merging.

**che-operator logs**: https://pastebin.com/jD31yHiQ

davidfestal:
@rhopp I fixed the bug you reported with the operator constantly deploying the Che server after a first CR update to TLS mode. I also fixed a similar issue with memory request / limit fields for the `che` and registry deployments.

I also rebased the PR on the last master, but the related commits are 5cee451, 9c32120 and 40c01b9.

@nickboldt FTR

rhopp:
I've tried `quay.io/dfestal/che-operator:fix-update-bugs` - LGTM.

PR #66, opened 2019-08-08
Test route url fix

sleshchenko:
The following PR has the following separated changes:
1. Update logrus from old `0.11.5` (Mar 14, 2017) to the latest v1.4.2 (May 18, 2019). It makes easier to debug che-operator if needed by enabling method name (more see https://github.com/sirupsen/logrus#logging-method-name)
3. Fix fetching test route URL.
There was a typo in test route assignment since `:=` declare new variable with limited scope, and `=` assign new value to exist variable.
Also, fetching is improved to make waiting for route host non-blocking.

PB logrus v1.4.2 https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20764


nickboldt:
This issue needs a milestone... is it blocker for 7.0? P1 for 7.1? 

Also I'm not sure I can review this. Might be better to let @davidfestal review when he's back from PTO next week.

nickboldt:
as to CQs, the rule is basically...

* if you depend on something that you SHIP (RUN-TIME), or something you need at BUILD-TIME only, you probably need a CQ.

You can look for existing CQs for current/previous versions of your deps here:

https://dev.eclipse.org/ipzilla/buglist.cgi?query_format=advanced&short_desc_type=allwordssubstr&short_desc=logrus&long_desc_type=substring&long_desc=&bug_file_loc_type=allwordssubstr&bug_file_loc=&keywords_type=allwords&keywords=&emailassigned_to1=1&emailtype1=substring&email1=&emailassigned_to2=1&emailreporter2=1&emailcc2=1&emailtype2=substring&email2=&bugidtype=include&bug_id=&chfieldfrom=&chfieldto=Now&chfieldvalue=&cmdtype=doit&order=Reuse+same+sort+as+last+time&field0-0-0=noop&type0-0-0=noop&value0-0-0=

pkg/controller/che/k8s_helpers.go
```diff
@@ -267,13 +267,16 @@ func (r *ReconcileChe) GetEndpointTlsCrt(instance *orgv1.CheCluster, url string)
 			return nil, err
 		}
 		// sometimes timing conditions apply, and host isn't available right away
-		if len(testRoute.Spec.Host) < 1 {
+		for i := 0; i < 10 && len(testRoute.Spec.Host) < 1; i++ {
```
davidfestal: I'm not a big fan of increasing the timeout from 1 second to 10 seconds. Afaik normally a operator controller should be as less blocking as possible.
I assume in our case it is acceptable until some wider refactoring is done.
However, since you return an error now when the route is not found, it might be interesting to test if the operator wouldn't simply trigger this request again after the error has been return, so that the process would resume without blocking the request loop.

sleshchenko: For me, it works even from the first try, I do not have an idea why testRoute.Spec.Host can be empty here, maybe it can be reproduced on some slow infrastructure...

So, timeout is set to 3 seconds to make operator less blocking.

> since you return an error now when the route is not found

Sorry but it's not an error when route is not found, but when route is found without host... And I'm not able to reproduce it on my local installation.

BTW Thanks that you caught my attention at that error, I figure it out that we must delete test route before returning an error.

sleshchenko: @davidfestal cc

sleshchenko: > However, since you return an error now when the route is not found, i

Let me clarify that I return an error if route is found but without host.

I've rewritten it in non-blocking way but I'm able to test only positive scenario when the route has host just after creation.
I do not have an idea why testRoute.Spec.Host can be empty here, maybe it can be reproduced on some slow infrastructure.

sleshchenko: The new approach is tested by simulating a case when test route does not have host immediately
```go
func (r *ReconcileChe) GetEndpointTlsCrt(instance *orgv1.CheCluster, url string) (certificate []byte, err error) {
	var requestURL string
	var testRoute *routev1.Route
	if len(url) < 1 {
		testRoute = r.GetEffectiveRoute(instance, "test")
		if testRoute == nil {
			testRoute = deploy.NewTlsRoute(instance, "test", "test", 8080)
			logrus.Infof("Creating a test route %s to extract router crt", testRoute.Name)
			if err := r.CreateNewRoute(instance, testRoute); err != nil {
				logrus.Errorf("Failed to create test route %s: %s", testRoute.Name, err)
				return nil, err
			}

                        // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
			return nil, errors.New("Fail the first time")
		}

		if len(testRoute.Spec.Host) < 1 {
			return nil, errors.New("Unable to get test route host for fetching certificate")
		}
 ...
```
And it seems to work as expected:
Screenshot_20190904_155704

@davidfestal, thanks for proposing this way!

pkg/controller/che/k8s_helpers.go
```diff
@@ -257,23 +256,24 @@ func (cl *k8s) GetDeploymentPod(name string, ns string) (podName string, err err
...
+				return nil, err
+			}
 		}
-		// sometimes timing conditions apply, and host isn't available right away
+
 		if len(testRoute.Spec.Host) < 1 {
-			time.Sleep(time.Duration(1) * time.Second)
-			testRoute := r.GetEffectiveRoute(instance, "test")
-			requestURL = "https://" + testRoute.Spec.Host
+			return nil, errors.New("Unable to get test route host for fetching certificate")
```
davidfestal: In order to be sure that the operator won't restart processing the custom resource immediately, you could also return the following:

```go
return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 1}, err
```

pkg/controller/che/k8s_helpers.go
```diff
@@ -257,23 +256,24 @@ func (cl *k8s) GetDeploymentPod(name string, ns string) (podName string, err err
...
+			testRoute = deploy.NewTlsRoute(instance, "test", "test", 8080)
+			logrus.Infof("Creating a test route %s to extract router crt", testRoute.Name)
+			if err := r.CreateNewRoute(instance, testRoute); err != nil {
+				logrus.Errorf("Failed to create test route %s: %s", testRoute.Name, err)
+				return nil, err
+			}
 		}
-		// sometimes timing conditions apply, and host isn't available right away
+
 		if len(testRoute.Spec.Host) < 1 {
```
davidfestal: So here it seems you'll always be in this case the first time since you don't get the effective route object after creating it in line [266](https://github.com/eclipse/che-operator/pull/66/files#diff-5899627ff2ee9c6f7afb34d27f2d3b3eR266) ?

Wouldn't it be a good idea to add

```go
testRoute = r.GetEffectiveRoute(instance, "test")
```

also just after line [266](https://github.com/eclipse/che-operator/pull/66/files#diff-5899627ff2ee9c6f7afb34d27f2d3b3eR266) ?

pkg/controller/che/k8s_helpers.go
davidfestal: Seems nice. Just added 2 comments about these changes:
- https://github.com/eclipse/che-operator/pull/66#discussion_r320832304
- https://github.com/eclipse/che-operator/pull/66#discussion_r320836623

pkg/controller/che/k8s_helpers.go
sleshchenko: I took a quick look on it and checked that 
```
r.client.Create(context.TODO(), route)
```
store the created object(response from server) into the second argument
See
https://github.com/eclipse/che-operator/blob/bc59de083a8be1bb3cdf53110bafc82cc23d3f84/vendor/sigs.k8s.io/controller-runtime/pkg/client/client.go#L103
https://github.com/eclipse/che-operator/blob/bc59de083a8be1bb3cdf53110bafc82cc23d3f84/vendor/sigs.k8s.io/controller-runtime/pkg/client/typed_client.go#L44
https://github.com/eclipse/che-operator/blob/bc59de083a8be1bb3cdf53110bafc82cc23d3f84/vendor/k8s.io/client-go/rest/request.go#L1094

Let me take a look closer and test it one more time

davidfestal: Thanks for checking !

davidfestal: Thanks for checking !

pkg/controller/che/k8s_helpers.go
sleshchenko: Here I'm not able to return `reconcile.Result` without modifying the signature. It seems to be a helper class and return here seems not good, at least there is no such any method in this class.

Do you think it is OK to send `reconcile.Result` back though `k8s_helper.go#GetEndpointTlsCrt` -> `create.go#CreateTLSSecret` -> `che_controller.go#Reconcile` ?

Or the following fragment in che_controller would be good enough?
```go
if err := r.CreateTLSSecret(instance, "", "self-signed-certificate"); err != nil {
  return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 1}, err
}
```
But it would mean that we wait one second on any error, like failed to get namespace, failed to create route, failed to fetch host... 
Then why we do not wait 1 second in other places like: when we failed to create service account https://github.com/eclipse/che-operator/blob/9682f3448fe240c216aa22d9d3f56cc056b01494/pkg/controller/che/che_controller.go#L315-L317

BTW Seems that even with `return reconcile.Result{}, err` there is ~ 1 second between `k8s_helper.go#GetEndpointTlsCrt` invocations
Screenshot_20190905_104935
the first one - is the first invocation
2nd and 3rd are the second invocation.

pkg/controller/che/k8s_helpers.go
sleshchenko: Screenshot_20190905_112140
As we can see here - route host is fetched from the first try

pkg/controller/che/k8s_helpers.go
davidfestal: > Here I'm not able to return reconcile.Result without modifying the signature.

Oh, yes, I didn't pay attention to the fact it was inside a utility function.
You might simply add a `retryLater` boolean return value, and use it in the `che_controller` reconcile function. 

> Then why we do not wait 1 second in other places like: when we failed to create service account 

Well the existing logic in the controller (that was already there when I started on it) seems to be the following:
- return **only an error** when it is an *unexpected error*, 
- return a **Result (with retry) + error** when it's an error that we expect in some scenarios and on which we explicitly want to restart the reconciliation after a while.

pkg/controller/che/k8s_helpers.go
davidfestal: Great, Thanks Sergii !

pkg/controller/che/k8s_helpers.go
sleshchenko: Do you think that we should retry if quite unexpected exception occurred like `failed to create route`? 
It's unexpected by maybe k8s/openshift API was temporary unavailable...
WDYT?

davidfestal: I think that if the `CreateTLSSecret` function returns `continueLater` with `true`, then the calling `reconcile` method should return `reconcile.Result{Requeue: true, RequeueAfter: time.Second * 1}, err` else it should return `nil, err` if an unexpected error occured.

sleshchenko: yeah, I understand it, let me rephrase my question:
CreateTLSSecret may returns error in different situations:
1. Failed to get an existing self-signed secret.
2. Failed to get existing route (like when API is not available)
3. Failed to create new route (like when API is not available)
4. The fetched route does not have host (it's clear that we should retry in such case)
5. Failed to request route host to retrieve TLS certificate
6. Failed to create a secret with fetched TLS self-signed certificate.

Which of this errors you consider as unexpected and which expected (like 4 about route) ?

davidfestal: To me you would only use `continueLater=true` in situation 4. We know that in some cases the host can take some time to be added to the route object. It's expected.

Possibly 5 could also be expected (and return `continueLater=true`, since sometimes the route may take some time to be setup.

But main point is the 4 IMO

Does it seem consistent to you ?

sleshchenko: I believe that this PR brings some fix and does not do code base worse.
Your proposed solution totally makes sense but it's more about refactoring the current architecture but not about fixing the issue I initially tried to use.
Sorry, but I'm not able to invest now in operator architecture.

@davidfestal If you agree that this PR is OK to merge - please approve.
If you think that it brings more issues than solves - feel free to close it.

tolusha:
Will done as part of refactoring
https://github.com/eclipse/che/issues/15127

PR #67, opened 2019-08-12
Inconsistent probes prevented deploying on RHPDS

davidfestal:
This PR aims at fixing an inconsistency in Che deployment readiness and liveness probes that used to prevent correct Che deployment on RHPDS Openshift 4 clusters.

The root cause was that Che start was quite long, and liveness probe was expired before the expiration of the readiness probe, which cause the Che POD to be constantly restarted.

@slemeur Did you have created an issue for your initial problem ? We could link it to the PR, and add it for triage to include it in 7.0.0 if you think it is critical.

### DIFF
--- a/pkg/deploy/deployment_che.go
+++ b/pkg/deploy/deployment_che.go
@@ -112,12 +112,13 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImage string, cheTag string, cmRe
 										Scheme: corev1.URISchemeHTTP,
 									},
 								},
+								// After POD start, the POD will be declared as either ready or failed after a minimum of 60 seconds and a maximum of 200 seconds
+								// 200 s = InitialDelaySeconds + TimeoutSeconds * FailureThreshold + PeriodSeconds * (FailureThreshold - 1)
 								InitialDelaySeconds: 60,
 								FailureThreshold:    10,
 								TimeoutSeconds:      5,
 								PeriodSeconds:       10,
-							},	// After POD start, the POD will be declared as either ready or failed after a minimum of 60 seconds and a maximum of 200 seconds
-									// 200 s = InitialDelaySeconds + TimeoutSeconds * FailureThreshold + PeriodSeconds * (FailureThreshold - 1)
+							},
 							LivenessProbe: &corev1.Probe{
 								Handler: corev1.Handler{
 									HTTPGet: &corev1.HTTPGetAction{
@@ -129,7 +130,8 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImage string, cheTag string, cmRe
 										Scheme: corev1.URISchemeHTTP,
 									},
 								},
-								InitialDelaySeconds: 200, // After POD start, don't initiate liveness probe before the POD has been declared ready or failed by the readiness probe
+								// After POD start, don't initiate liveness probe before the POD has been declared ready or failed by the readiness probe
+								InitialDelaySeconds: 200,
 								FailureThreshold:    3,
 								TimeoutSeconds:      3,
 								PeriodSeconds:       10,
--- a/pkg/deploy/deployment_che.go
+++ b/pkg/deploy/deployment_che.go
@@ -112,10 +112,10 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImage string, cheTag string, cmRe
 										Scheme: corev1.URISchemeHTTP,
 									},
 								},
-								// After POD start, the POD will be declared as either ready or failed after a minimum of 60 seconds and a maximum of 200 seconds
-								// 200 s = InitialDelaySeconds + TimeoutSeconds * FailureThreshold + PeriodSeconds * (FailureThreshold - 1)
-								InitialDelaySeconds: 60,
-								FailureThreshold:    10,
+								// After POD start, the POD will be seen as ready after a minimum of 15 seconds and we expect it to be seen as ready until a maximum of 200 seconds
+								// 200 s = InitialDelaySeconds + PeriodSeconds * (FailureThreshold - 1) + TimeoutSeconds
+								InitialDelaySeconds: 25,
+								FailureThreshold:    18,
 								TimeoutSeconds:      5,
 								PeriodSeconds:       10,
 							},
@@ -130,7 +130,7 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImage string, cheTag string, cmRe
 										Scheme: corev1.URISchemeHTTP,
 									},
 								},
-								// After POD start, don't initiate liveness probe before the POD has been declared ready or failed by the readiness probe
+								// After POD start, don't initiate liveness probe while the POD is still expected to be declared as ready by the readiness probe
 								InitialDelaySeconds: 200,
 								FailureThreshold:    3,
 								TimeoutSeconds:      3,
--- a/pkg/deploy/deployment_che.go
+++ b/pkg/deploy/deployment_che.go
@@ -116,7 +116,8 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImage string, cheTag string, cmRe
 								FailureThreshold:    10,
 								TimeoutSeconds:      5,
 								PeriodSeconds:       10,
-							}, // Will stop testing readiness after a minimum of 65 seconds and a maximum of 200 seconds (60 + 10*5 + 9*10)
+							},	// After POD start, the POD will be declared as either ready or failed after a minimum of 60 seconds and a maximum of 200 seconds
+									// 200 s = InitialDelaySeconds + TimeoutSeconds * FailureThreshold + PeriodSeconds * (FailureThreshold - 1)
 							LivenessProbe: &corev1.Probe{
 								Handler: corev1.Handler{
 									HTTPGet: &corev1.HTTPGetAction{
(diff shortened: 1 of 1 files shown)


slemeur:
Here: https://github.com/eclipse/che/issues/14209

pkg/deploy/deployment_che.go
```diff
@@ -112,10 +112,11 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImage string, cheTag string, cmRe
 										Scheme: corev1.URISchemeHTTP,
 									},
 								},
-								InitialDelaySeconds: 25,
-								FailureThreshold:    5,
+								InitialDelaySeconds: 60,
+								FailureThreshold:    10,
 								TimeoutSeconds:      5,
-							},
+								PeriodSeconds:       10,
```
ibuziuk: @davidfestal why `periodSeconds` is required for the `ReadinessProbe` ?

pkg/deploy/deployment_che.go
```diff
@@ -112,10 +112,11 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImage string, cheTag string, cmRe
...
+							}, // Will stop testing readiness after a minimum of 65 seconds and a maximum of 200 seconds (60 + 10*5 + 9*10)
```
ibuziuk: could you please clarify math?

> Will stop testing readiness after a minimum of 65 seconds

the readiness probe will continue to be called throughout the lifetime of the container, every periodSeconds, so that the container can make itself temporarily unavailable when one of its dependencies is unavailable, or while running a large batch job, performing maintenance, or something similar. [1] 

[1] https://blog.colinbreck.com/kubernetes-liveness-and-readiness-probes-how-to-avoid-shooting-y

pkg/deploy/deployment_che.go
```diff
@@ -127,9 +128,10 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImage string, cheTag string, cmRe
 										Scheme: corev1.URISchemeHTTP,
 									},
 								},
-								InitialDelaySeconds: 50,
+								InitialDelaySeconds: 200, // Start liveness probe after readiness probe has finished
```
ibuziuk: > Start liveness probe after readiness probe has finished

The comment is misleading IMO. `initialDelaySeconds: Number of seconds after the container has started before liveness or readiness probes are initiated.`

pkg/deploy/deployment_che.go
davidfestal: Well, I just added the default value to make it clear what is being used. This is the value effectively added by OpenShift, that you can still find in the effective deployment, even when not putting it here.

pkg/deploy/deployment_che.go
davidfestal: > could you please clarify math?

Sorry, it's a minimum of 60 seconds, not 65.

Here I'm considering the case of the POD start only:
- First, readiness probe will wait for 60 seconds after the start of the POD before checking anything
- Then the probe will start checking readiness. It might be that the first check will be successful immediately. In this case, POD will be seen as __*ready*__ after 60 seconds minimum. Else check will wait during a 5 seconds timeout.
- At most it will do it 10 times before declaring the POD as __*failed*__
- Between each check (10 checks at most, each of which can last 5 seconds at most), the probe will wait for 10 seconds (the period). That means that in our case this 10 seconds wait might occur at most 9 times (between the 10 checks).

So the math is:

Longest time (from POD start) after which the POD is declared __*ready*__ or __*failed*__ =
`InitialDelaySeconds` +
`FailureThreshold` * `TimeoutSeconds` +
(`FailureThreshold` - 1) * `PeriodSeconds`

Does it help .

> Will stop testing readiness after a minimum of 65 seconds

You're right, wording is misleading here: readiness probe wil continue its work. But at least the POD will, for the first time, be either declared __*ready*__ or __*failed*__ after this maximum delay. that's my main concern here.

pkg/deploy/deployment_che.go
davidfestal: You're right, the comment is misleading and not precise enough.

What I was meaning is that we don't want the liveness probe (that would *restart* the POD in case of failure) to start checking (just after POD start) before the POD is, for the first time, either declared __*ready*__ or __*failed*__ thanks to the readiness probe work.

That's important because liveness probe and readiness probe **use the same check**.

But **after** the POD is started and has been declared as __*ready*__ by the readiness probe, then the liveness probe will starts checking, and, since it is more requiring than the readiness probe, we can be sure that it will restart the POD if any unavailability occurs.

pkg/deploy/deployment_che.go
```diff
@@ -112,10 +112,12 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImage string, cheTag string, cmRe
...
 								},
-								InitialDelaySeconds: 25,
-								FailureThreshold:    5,
+								InitialDelaySeconds: 60,
+								FailureThreshold:    10,
 								TimeoutSeconds:      5,
-							},
+								PeriodSeconds:       10,
+							},	// After POD start, the POD will be declared as either ready or failed after a minimum of 60 seconds and a maximum of 200 seconds
+									// 200 s = InitialDelaySeconds + TimeoutSeconds * FailureThreshold + PeriodSeconds * (FailureThreshold - 1)
```
ibuziuk: not sure if this is the intentional offset for the comment ;-)

davidfestal: Well, it is tab-based indentation which, in VSCode at least, seems to align both comments.

davidfestal: Should be fixed in commit a77fc44

pkg/deploy/deployment_che.go
sleshchenko: I do not really follow the relation of liveness probe to readiness.
I mean I understand and agree on the following formulating
```
What I was meaning is that we don't want the liveness probe (that would restart the POD in case of failure) to start checking (just after POD start) before the POD is
```
before tomcat inside of POD is started. 
But I don't understand how `ready`/`failed` readiness probe influence liveness probe at all.

Also, note in case of 
```
FailureThreshold:    10,
TimeoutSeconds:      5,
```
if tomcat is hanging or crashed, K8s/OpenShift will continue set client traffic to it for 50 seconds (PeriodSeconds*FailureThreshold). Do we really need it?

What if we apply the following changes:
```
readiness:
  // Che can become available in minimum 25 seconds
  // If it's not available within 75 seconds, it will be marked as **Not Ready**
  // and check will continue to happen each 10 seconds
  // it will become ready after first successfull probe
  // P.S. I do know if it's too bad that we can have unready starting tomcat in such case instead 
  // of just starting but it allows Che Server process traffic faster if it's already ready, no need to wait 60 seconds.
  InitialDelaySeconds: 25, // maybe like 40 seconds
  FailureThreshold:    5,
  PeriodSeconds:       10,
liveness:
  // it's a normal case if tomcat start takes ~ 200 seconds
  // do not check liveness before to prevent killing Che Server that is still starting
  InitialDelaySeconds: 200,
  PeriodSeconds:       10,
  FailureThreshold:    3,
  ...
```

davidfestal: > Also, note in case of
> 
> ```
> FailureThreshold:    10,
> TimeoutSeconds:      5,
> ```
> 
> if tomcat is hanging or crashed(in the process of work, after successful start), K8s/OpenShift will continue set client traffic to it for 50 seconds (PeriodSeconds*FailureThreshold). Do we really need it?

No not when the liveness probe will be running afaict. So are you speaking of the case when Tomcat would be successfully started after 60 seconds, the container would be marked as `Ready` by the readiness probe, and the tomcat server would fail again just after ... Then we would still send traffic to it during 140 seconds (200 - 60), until liveness probe would be initiated and would detect the crashed tomcat and restart the POD ?

sleshchenko: > No not when the liveness probe will be running afaict. So are you speaking of the case when Tomcat would be successfully started after 60 seconds, the container would be marked as Ready by the readiness probe, and the tomcat server would fail again just after ... Then we would still send traffic to it during 140 seconds (200 - 60), until liveness probe would be initiated and would detect the crashed tomcat and restart the POD ?

Let me clarify the scenario I would like to highlight:
Che Server is started, it works for about 10 minutes, both of readiness and liveness probes are done successfully but then let's imagine connection pool is full and tomcat refused every request.
In our case, readiness and liveness probes are the same and with current values
1. liveness probe: period 10, FailureThreshold 3;  <- container will be restarted after 30 seconds;
2. readiness probe: period 10, FailureThreshold 10; <- container will receive all traffic until it won't be restarted, because 100 seconds it's needed for K8s to mark it as *Not Ready*.

In my understanding, readiness probe should fail sooner than lineness, it's a case then my application is temporary unavailable but let's give some time it to recover itself.

See https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-readiness-probes

davidfestal: >```
>  // If it's not available within 75 seconds, it will be marked as **Not Ready**
>  // and check will continue to happen each 10 seconds
>  // it will become ready after first successful probe
>```

Yes, you're right, But then after the 75 seconds we will have `Unhealthy` container events occurring in the POD events, which we probably don't want if we consider this time to be still acceptable to start Che.

davidfestal:
> Had a good discussion via BJ and agreed on changes that should be done before the merge

Yeah, thanks for your help @sleshchenko and detailed explanations about the aspects I had misundertood. I updated my previous  comments (answers to @ibuziuk ) accordingly to be more consistent with the right documentation interpretation.

Let me summarize the last change we agreed on related to the readiness probe:

- Lower `InitialDelaySeconds` to a low value (for developer's sake), say to 15 seconds
- Increase the `FailureThreshold` accordingly (=> to 13) to avoid `Unhealthy` events before reaching 200 seconds after the POD start.

PR #69, opened 2019-08-14
Release 7.0.0

davidfestal:
Release 7.0.0.

Changes done during the release should be merged to `master` because they contain new OLM files and new default GO values that will be used as the new basis for future master changes.

### DIFF
--- a/olm/prepare-community-operators-update.sh
+++ b/olm/prepare-community-operators-update.sh
@@ -59,7 +59,7 @@ do
   folderToUpdate="${communityOperatorsLocalGitFolder}/${platformSubFolder}/eclipse-che"
   destinationPackageFilePath="${folderToUpdate}/eclipse-che.package.yaml"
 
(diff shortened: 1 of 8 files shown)


olm/prepare-community-operators-update.sh
```diff
@@ -59,7 +59,7 @@ do
   folderToUpdate="${communityOperatorsLocalGitFolder}/${platformSubFolder}/eclipse-che"
   destinationPackageFilePath="${folderToUpdate}/eclipse-che.package.yaml"
 
-  lastPublishedPackageVersion=$(yq -r '.channels[] | select(.name == "final") | .currentCSV' "${destinationPackageFilePath}" | sed -e "s/eclipse-che.v//")
+  lastPublishedPackageVersion=$(yq -r '.channels[] | select(.name == "stable") | .currentCSV' "${destinationPackageFilePath}" | sed -e "s/eclipse-che.v//")
```
nickboldt: why change to stable from final ?

davidfestal: because it was decided to change the name of the channel to `stable` by @l0rd and @slemeur , at least for upstream Che operators.

PR #71, opened 2019-08-22
Add console link

sparkoo:
### What does this PR do?
It adds creating of ConsoleLink object if the corresponding API is available onto cluster.

This PR fixes https://github.com/eclipse/che/issues/14203

### Is it tested?
It's not fully tested yet but here you can find tested cases:

* OpenShift 4.2 (crc 4.1.11 on linux with consolelink CRD registered and console 4.2 used): TLS + Self-signed cert
Console link is created or update depending if it was created before
Screenshot_20190904_134436

<details>
<summary>Logs</summary>

Screenshot_20190904_132853

</details>
  
* OpenShift 4.2 (crc 4.1.11 on linux with consolelink CRD registered and console 4.2 used): HTTP
Console link is not created
<details>
<summary>Logs</summary>

Screenshot_20190904_133341

</details>

* OpenShift 4.1(crc 4.1.11 on linux)
Nothing failed, che-operator prints debug message that `consolelinks` is not supported by cluster
<details>
<summary>Logs</summary>
Screenshot_20190904_145325

</details>

* OpenShift 3.11 (minishift v1.34.1 on linux)
Nothing failed, che-operator prints debug message that `consolelinks` is not supported by cluster
<details>
<summary>Logs</summary>

Screenshot_20190903_170055

</details>

* Kubernetes v1.15.2 (minikube v1.3.1 on linux)
Nothing failed, che-operator prints debug message that `consolelinks` is not supported by cluster
<details>
<summary>Logs</summary>

Screenshot_20190903_162640

</details>

### How to test it?
Operator image to test here: `sleshchenko/che-operator:consoleLink`
To use the image apply following patch and run `./deploy.sh`:
```
diff --git a/deploy/operator.yaml b/deploy/operator.yaml
index f7c9070..3b1810e 100644
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -24,7 +24,7 @@ spec:
     spec:
       containers:
         - name: che-operator
-          image: quay.io/eclipse/che-operator:nightly
+          image: sleshchenko/che-operator:consoleLink
           ports:
           - containerPort: 60000
             name: metrics
```

Please note that consolelink supports just `https` (https://github.com/openshift/console-operator/blob/master/manifests/00-crd-extension-console-link.yaml#L52) so you have to set your deployment to use https. Applying this patch and running `./deploy.sh` should be enough.


pkg/controller/che/che_controller.go
```diff
@@ -608,6 +609,17 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 			}
 		}
 	}
+
+	// create ConsoleLink
+	spec := v1.ConsoleLinkSpec{Link: v1.Link{Href: "https://" + instance.Spec.Server.CheHost, Text: "Eclipse Che"},
```
skabashnyuk: is there a way to get protocol somehow?

pkg/controller/che/che_controller.go
```diff
@@ -608,6 +609,17 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+		Location: v1.ApplicationMenu,
+		ApplicationMenu: &v1.ApplicationMenuSpec{Section: "Red Hat Applications",
+			ImageURL: "https://raw.githubusercontent.com/eclipse/che-website/master/che/images/ico/96x96.png"}}
```
skabashnyuk: what about
- https://{che_host}/dashboard/assets/branding/loader.svg
- https://{che_host}/dashboard/assets/branding/che-logo.svg
?

pkg/controller/che/che_controller.go
sparkoo: yes, but consolelink support just https (https://github.com/openshift/console-operator/blob/master/manifests/00-crd-extension-console-link.yaml#L52). I'll add `if https`...

pkg/controller/che/che_controller.go
sparkoo: nice, I'll use that

sparkoo:
I've polished the code and updated PR description according to that. There's one new issue I've hit. `che-operator` role can't manage consolelinks as they are at cluster scope and `che-operator` is only in given namespace.
So current version get looping with this error:
`E0823 14:24:27.574932       1 reflector.go:134] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:196: Failed to list *v1.ConsoleLink: consolelinks.console.openshift.io is forbidden: User "system:serviceaccount:eclipse-che:che-operator" cannot list resource "consolelinks" in API group "console.openshift.io" at the cluster scope`
I didn't hit that before because I was running operator locally, directly from my local machine communicating with the cluster. However, when the operator is deployed and ran with `che-operator` role, it hits this.

sparkoo:
@davidfestal @sleshchenko @skabashnyuk  any ideas how to fix it ? https://github.com/eclipse/che-operator/pull/71#issuecomment-524336952

davidfestal:
@sparkoo just add the required cluster permission to the operator. That would mainly be adding your new permission in the operator [cluster role](https://github.com/eclipse/che-operator/blob/master/deploy/cluster_role.yaml#L28)

You can add it manually in the current cluster you're testing in (as kubeadmin).
Of course when you need to have these permissions setup automatically by the OperatorHub installation, there would changes to do in the synced CSV, and I assume at the time we would sync and I could help you on this.
But for now, in order to further test your changes, simply searching for this cluster role and modifying it (or adding a new one manually) should be sufficient.

cmd/manager/main.go
```diff
@@ -93,14 +92,23 @@ func main() {
...
-	logrus.Info("Registering Components")
+	logrus.Info("Registering Che Components Types")
 
 	// Setup Scheme for all resources
 	if err := apis.AddToScheme(mgr.GetScheme()); err != nil {
 		logrus.Error(err, "")
 		os.Exit(1)
 	}
 
+	logrus.Info("Registering Console Components Types")
```
davidfestal: Did you test this under OpenShift 3.11 or OpenShift 4.1 ?
It seems to me that at least you should register the console objects and console scheme only when it is supported and available.
So I assume that at least here would be the right place: https://github.com/eclipse/che-operator/blob/master/pkg/controller/che/che_controller.go#L75

deploy/cluster_role.yaml
```diff
@@ -27,6 +27,7 @@ rules:
       - watch
   - apiGroups:
       - config.openshift.io
+      - console.openshift.io
     resources:
       - infrastructures
```
davidfestal: Don't you add the `consolelinks` resource here ?

cmd/manager/main.go
sleshchenko: Yes, I tested it on `OpenShift 3.11` and `OpenShift 4.2`
In my understanding, we are not registering CRD on openshift by invoking it, but configuring our k8s client to knows about such kind of objects.

This is still draft and I'm going to test it on Openshift 3.11, 4.1, 4.2 before making it ready to review.

deploy/cluster_role.yaml
sleshchenko: I did not fail on OpenShift 3.11 but I did not investigate when this file is applied.
Will investigate and answer here why it should not fail or will rework it.

sleshchenko: Ah, sorry. I misunderstood your question! Good catch!

sleshchenko: BTW I quickly tried to test if putting extra items (which is not supported on a cluster) does not lead to errors. Seems that it does not
Screenshot_20190902_163815
Will retest it later

sleshchenko:
@davidfestal 

> Of course when you need to have these permissions setup automatically by the OperatorHub installation, there would changes to do in the synced CSV, and I assume at the time we would sync and I could help you on this.

Could you please go further and share a place where I should apply for consolelinks cluster permissions or you're going to handle that?

davidfestal:
@sleshchenko

This is mainly what is described in the following Readme https://github.com/eclipse/che-operator/blob/master/olm/README.md.

However since I changed the version of the required operator-sdk CLI command to version v0.1.0 in a recent PR, you might want to rebase your PR on master before applying the steps defined in the Readme.

We can also sync and discuss it if you want more details about how it works and why it is like that.

sleshchenko:
@davidfestal Seems I've updated mentioned CSV https://github.com/eclipse/che-operator/pull/71/commits/14d8e43b27a0a2bb55b2d58155666f6cc4d46d68
Could you explain more what is `v9.9.9`, it is right that I've generated `9.9.9-nightly.1567514659 
` version?

deploy/cluster_role.yaml
```diff
@@ -36,9 +36,10 @@ rules:
     resources:
       - consolelinks
     verbs:
-      - create
       - get
-      - delete
       - list
-      - patch
+      - watch
```
davidfestal: You should be able to skip the `watch` right by using the new [`nonCachedClient`](https://github.com/eclipse/che-operator/blob/9682f3448fe240c216aa22d9d3f56cc056b01494/pkg/controller/che/che_controller.go#L204) reconciler attribute.

It allows accessing the K8S API without trying to further cache / watch the related K8S objects, which doesn't seem to be necessary here.

deploy/cluster_role.yaml
davidfestal: Yes, you're right afaik.

deploy/cluster_role.yaml
sleshchenko: Current solution tries to get and update Che Cluster each time when che operator is invoked.
So, it makes sense to have `watch` here to avoid retrieving it each time
Screenshot_20190903_115130

Or maybe you would suggest another solution, like:
1. Do not update console link but only create it when it does not exist. But I see an error here, since existing consolelink may be outdated.
2. something else.

Thanks.

davidfestal:
> @davidfestal Seems I've updated mentioned CSV [14d8e43](https://github.com/eclipse/che-operator/commit/14d8e43b27a0a2bb55b2d58155666f6cc4d46d68)
> Could you explain more what is `v9.9.9`, it is right that I've generated `9.9.9-nightly.1567514659 ` version?

Sure. Here we're creating a new version (that would be automatically updated by the OperatorHub) **in the `nightly` channel** of our preview Quay applications.

Since it's a *nightly* channel that would *always contain the very last code on the `master` branch* of the `che-operator`, I simply used `9.9.9-nightly-<timestamp>` because OLM requires well-formed, ordered, semantic-versionning-compatible versions.

What you have to keep in mind is that real release-related CSVs will be later created by a script (during release process) from the last nightly version with proper version names and numbers.

deploy/cluster_role.yaml
davidfestal: Afaik the Che operator is not called if no change occurred in the `CheCluster` resource.
That's not going to happen that regularly, and I don't see any value / gain trgigerring regular watch of a `ConsoleLink` object. The only impact is increasing the range of required permissions.

cmd/manager/main.go
davidfestal: Could you still move your registration code here please, at least to be consistent with the rest of the code:

https://github.com/eclipse/che-operator/blob/9682f3448fe240c216aa22d9d3f56cc056b01494/pkg/controller/che/che_controller.go#L86

deploy/cluster_role.yaml
sleshchenko: OK, will get rid of watch verb here at least for the time being.

cmd/manager/main.go
sleshchenko: > at least to be consistent with the rest of the code

Will do

sleshchenko:
I believe this is good enough to be ready to review =)

deploy/cluster_role.yaml
sleshchenko: done

skabashnyuk:
@davidfestal can you take a look on this pr again, please?

pkg/controller/che/che_controller.go
```diff
@@ -92,6 +94,12 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		if err := userv1.AddToScheme(mgr.GetScheme()); err != nil {
 			logrus.Errorf("Failed to add OpenShift User to scheme: %s", err)
 		}
+        if err := consolev1.AddToScheme(mgr.GetScheme()); err != nil {
```
davidfestal: TBH I would have preferred this to be done *only when the console link is supported by the underlying OpenShift cluster* (simply inside a condition calling the `hasConsolelinkObject` function), even though I agree we might have noticed that unknown schemas/objects are ignored when unknown. Still seems cleaner (and clearer) to me.

@sleshchenko wdyt ?

sleshchenko: For me adding types to schema it's something as `init` operation, and `add` seems to be such a place.
Moving it to a condition calling the `hasConsolelinkObject` would mean that we register ( or try to register since we can add the corresponding `isAlreadyRegistered` method before ) each time when the operator is triggered (of course if consolelink is supported).
Personally I would leave as is to just configure client in one place.

But if you have a strong opinion that it must be done in the condition - it's acceptable for me, so I'll do.

davidfestal: Oh, I wasn't clear.
I was meaning let it here, but just as add a condition, just similarly as it is done in line [87](https://github.com/eclipse/che-operator/pull/71/files#diff-d9445e614f0d728fd140771b0fd29c50R87) to restrict these new registrations to the case when console links are supported.

sleshchenko: > I was meaning let it here, but just as add a condition, just similarly as it is done in line 87 to restrict these new registrations to the case when console links are supported.

Ah, sorry for the misunderstanding. 
Then it makes sense.

davidfestal:
Maybe we could wait a bit to merge, and merge this just after 7.1.0 is released ?
wdyt @l0rd ?

l0rd:
That's not a blocker for 7.1.0. We can wait 7.2.0 to merge it.

sleshchenko:
Marked PR as WIP to avoid occasional merging before 7.1.0 is released.

@l0rd @davidfestal Do you happen to know who is able to enable TODO bot this repo? Should I register an issue for WebMaster?

davidfestal:
> Do you happen to know who is able to enable TODO bot this repo?

I don't know. Maybe @nickboldt knows ?

l0rd:
@sleshchenko I think you should do that through a ticket. cc @benoitf

sleshchenko:
Registered an issue https://bugs.eclipse.org/bugs/show_bug.cgi?id=550803

pkg/controller/che/che_controller.go
```diff
@@ -1042,3 +1058,67 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+func createConsoleLink(isOpenShift4 bool, protocol string, instance *orgv1.CheCluster, r *ReconcileChe) error {
+	if !isOpenShift4 || !hasConsolelinkObject() {
+		logrus.Debug("Console link won't be created. It's not supported by cluster")
+		// console link is supported only on OpenShift >= 4.2
+		return nil
+	}
+
+	if protocol != "https" {
+		logrus.Debug("Console link won't be created. It's not supported when http connection is used")
+		// console link is supported only on OpenShift >= 4.2
```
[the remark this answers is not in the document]
sleshchenko: Nice catch! Fixed

pkg/controller/che/che_controller.go
```diff
@@ -92,6 +94,14 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		if err := userv1.AddToScheme(mgr.GetScheme()); err != nil {
 			logrus.Errorf("Failed to add OpenShift User to scheme: %s", err)
 		}
+		if hasConsolelinkObject() {
+			if err := consolev1.AddToScheme(mgr.GetScheme()); err != nil {
+				logrus.Errorf("Failed to add OpenShift ConsoleLink to scheme: %s", err)
+			}
+			// registration for ConsoleLinkList is missing in consoleApi
+			// this is temporary until https://github.com/openshift/api/pull/422 is not merged
```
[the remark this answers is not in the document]
sleshchenko: Good catch! Updated vendor and removed this comment!

PR #72, opened 2019-08-23
Update identityProvider and postgres container images to latest defaults on Operator update (Fix #13714)

davidfestal:
This Pull Request fixes issue https://github.com/eclipse/che/issues/13714.
It now avoids updating the `CheCluster` custom resource with default docker images of Keycloak, Postgres, and PVCJobs. This was, when the operator changes default values, they will be automatically picked up for an existing Custom Resource that initially used defaults, so that the corresponding Keycloak and Postgres deployments will be automatically updated.

This PR also detects if a custom resource is explicitly using the docker image values defined in the 7.0.0 GA operator, and if it's the case, is resetting the CR value to empty string (== default value), so that it will be able to upgrade it to the next default at next release.


pkg/deploy/defaults.go
```diff
@@ -30,21 +30,23 @@ const (
 	DefaultPvcClaimSize                 = "1Gi"
 	DefaultIngressStrategy              = "multi-host"
 	DefaultIngressClass                 = "nginx"
-	DefaultPluginRegistryImage          = "quay.io/eclipse/che-plugin-registry:7.0.0"
+	defaultPluginRegistryImage          = "registry.redhat.io/codeready-workspaces/che-plugin-registry:2.0"
+	defaultPluginRegistryUpstreamImage  = "quay.io/eclipse/che-plugin-registry:7.0.0"
 	DefaultPluginRegistryMemoryLimit    = "256Mi"
 	DefaultPluginRegistryMemoryRequest  = "16Mi"
-	DefaultDevfileRegistryImage         = "quay.io/eclipse/che-devfile-registry:7.0.0"
+	defaultDevfileRegistryImage         = "registry.redhat.io/codeready-workspaces/che-devfile-registry:2.0"
```
nickboldt: probably won't be called "che-" anything ... more likely.. 

`	defaultDevfileRegistryImage         = "registry.redhat.io/codeready-workspaces/devfile-registry:2.0"`

pkg/deploy/defaults.go
```diff
@@ -30,21 +30,23 @@ const (
...
 	DefaultKeycloakAdminUserName        = "admin"
 	DefaultCheLogLevel                  = "INFO"
 	DefaultCheDebug                     = "false"
-	DefaultPvcJobsImage                 = "registry.redhat.io/ubi8-minimal:8.0-159"
-	DefaultPvcJobsUpstreamImage         = "registry.access.redhat.com/ubi8-minimal:8.0-159"
-	DefaultPostgresImage                = "registry.redhat.io/rhscl/postgresql-96-rhel7:1-46"
-	DefaultPostgresUpstreamImage        = "centos/postgresql-96-centos7:9.6"
-	DefaultKeycloakImage                = "registry.redhat.io/redhat-sso-7/sso73-openshift:1.0-13"
-	DefaultKeycloakUpstreamImage        = "eclipse/che-keycloak:7.0.0"
+	defaultPvcJobsImage                 = "registry.redhat.io/ubi8-minimal:8.0-159"
```
nickboldt: the subtlety of upper and lower case vars here might be lost on people... is there no better way to define the distinction?

davidfestal: Well, lowercase is because these are not shared.
uppercase means that they are shared outside of the package.

PR #73, opened 2019-08-23
Bring che operator in line with the latest changes coming to Che server to support single-host mode.

metlos:
Che server will require different ingress annotations and a new config variable to support single-host mode. This brings the operator inline with what is going to be required once https://github.com/eclipse/che/pull/14134 is merged.


metlos:
@davidfestal could we please move this forward?

davidfestal:
@metlos are we still waiting for something of can we merge this now ?

metlos:
@davidfestal we can merge this now

PR #74, opened 2019-09-02
Don't allow installing Che with Openshift Oauth when no OAuth user exist

davidfestal:
This PR fixes the issue https://github.com/eclipse/che/issues/14013 : it prevents using Openshift OAuth if no real user exists in the OpenShift cluster.  

It also provides a way to show an error reason when the Che Operator cannot start a Che server, as well as a detailed explanation, and possibly a link to point to the related documentation.

The result in the OperatorHub shown in the following image:

image

### DIFF
--- a/deploy/cluster_role.yaml
+++ b/deploy/cluster_role.yaml
@@ -31,3 +31,9 @@ rules:
       - infrastructures
     verbs:
       - get
+  - apiGroups:
+      - user.openshift.io
+    resources:
+      - users
+    verbs:
+      - list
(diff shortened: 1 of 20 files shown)


davidfestal:
Useful notice: The main changes made to the OperatorHub files can be done through the followjng diff files:
- https://github.com/eclipse/che-operator/pull/74/files#diff-e2bb4d553ce81b2fa5cb9902f37eb438
- https://github.com/eclipse/che-operator/pull/74/files#diff-3bc5ce4cea4dc17ac406608d8eafe6a2

PR #75, opened 2019-09-03
Directly login through the `openshift` provider when OpenShift OAut is enabled

davidfestal:
This PR is the implementation of issue https://github.com/eclipse/che/issues/14012

When the Openshift OAuth is configured, it configure the Keycloak server so that it directly uses the OpenShift login page when connecting to Che.

### DIFF
--- a/pkg/controller/che/create.go
+++ b/pkg/controller/che/create.go
@@ -296,7 +296,11 @@ func (r *ReconcileChe) CreateIdentityProviderItems(instance *orgv1.CheCluster, r
 	}
 
 	if !tests {
-		openShiftIdentityProviderCommand := deploy.GetOpenShiftIdentityProviderProvisionCommand(instance, oAuthClientName, oauthSecret, keycloakAdminPassword, isOpenShift4)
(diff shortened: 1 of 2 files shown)


sleshchenko:
@davidfestal Is there a way how to walk over OAuth redirect and log in as Che Admin user?

davidfestal:
@sleshchenko You can easily disable automatic rediect to Openshift login in the Keycloak admin console. see https://github.com/eclipse/che/issues/14434#issuecomment-528349521

However I'm not sure that it would fix the problem you're trying to solve.
Even when logging as Che admin, how would the Che admin user have write access to the namespaces of Openshift users to clean the workspace K8S resources ?

Did we ever test the use of the Che admin user when running with the openshift OAuth integration enabled ? I don't have the answer to this question.

cc @l0rd

sleshchenko:
@davidfestal Thanks for quick response. I'm trying to remove existing user and don't need namespace access (in my case, since I can clean up namespace from OpenShift console directly).

> Did we ever test the use of the Che admin user when running with the openshift OAuth integration enabled ?

It's a good topic we might need to invest time to provide good UX for Che Admin

PR #85, opened 2019-09-27
Release 7.2.0

davidfestal:
Release 7.2.0.

Changes done during the release should be merged to master because they contain new OLM files and new default GO values that will be used as the new basis for future master changes.

Additionally, a minor fix has been done to a script file, in order to fix a minor issue encountered during release process.

### DIFF
--- a/pkg/deploy/defaults.go
+++ b/pkg/deploy/defaults.go
@@ -19,7 +19,7 @@ import (
 const (
 	defaultCheServerImageRepo           = "eclipse/che-server"
 	defaultCodeReadyServerImageRepo     = "registry.redhat.io/codeready-workspaces/server-rhel8"
-	defaultCheServerImageTag            = "7.1.0"
+	defaultCheServerImageTag            = "7.2.0"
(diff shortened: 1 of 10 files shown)


davidfestal:
Merging now that everything has been merged on the `community-operator` and propagated to the OperatorHub for both OpenShift and Kubernetes

pkg/deploy/defaults.go
```diff
@@ -19,7 +19,7 @@ import (
 const (
 	defaultCheServerImageRepo           = "eclipse/che-server"
 	defaultCodeReadyServerImageRepo     = "registry.redhat.io/codeready-workspaces/server-rhel8"
-	defaultCheServerImageTag            = "7.1.0"
+	defaultCheServerImageTag            = "7.2.0"
```
sleshchenko: @davidfestal Can we have nightly defaults in master to have nightly Che deployed along with nightly che-operator?

davidfestal: Not for now.
This was already discussed but would need a wider refactoring.
For now nightlies are managed differently.

PR #86, opened 2019-10-02
Sync default properties with upstream

skabashnyuk:
Removed some defaults to be closer to upstream configuration
- CHE_PREDEFINED_STACKS_RELOAD__ON__START - not used in Che 7
- CHE_WORKSPACE_AUTO_START - wrong parameter, has to be CHE_WORKSPACE_AUTO__START, upstream default value - ```true```
- CHE_INFRA_KUBERNETES_WORKSPACE__UNRECOVERABLE__EVENTS - upstream value ```FailedMount,FailedScheduling,MountVolume.SetUp failed,Failed to pull image,FailedCreate```
- CHE_LIMITS_WORKSPACE_IDLE_TIMEOUT - upstream value ```1800000```
- CHE_INFRA_KUBERNETES_SERVICE__ACCOUNT__NAME - Not changed since it has ```NULL``` value in upstream. To remove it we need to investigate if it's safe to do or not.

Refer to https://github.com/eclipse/che/issues/14614

### DIFF
--- a/pkg/deploy/che_configmap.go
+++ b/pkg/deploy/che_configmap.go
@@ -73,6 +73,7 @@ func GetCustomConfigMapData() (cheEnv map[string]string) {
 		"CHE_INFRA_KUBERNETES_SERVICE__ACCOUNT__NAME":           "che-workspace",
 		"CHE_WORKSPACE_AUTO_START":                              "true",
 		"CHE_INFRA_KUBERNETES_WORKSPACE__UNRECOVERABLE__EVENTS": "FailedMount,FailedScheduling,MountVolume.SetUp failed,Failed to pull image",
+		"CHE_LIMITS_WORKSPACE_IDLE_TIMEOUT":                     "1800000",
 	}
 	return cheEnv
 
--- a/pkg/deploy/che_configmap.go
+++ b/pkg/deploy/che_configmap.go
@@ -73,7 +73,6 @@ func GetCustomConfigMapData() (cheEnv map[string]string) {
 		"CHE_INFRA_KUBERNETES_SERVICE__ACCOUNT__NAME":           "che-workspace",
 		"CHE_WORKSPACE_AUTO_START":                              "true",
 		"CHE_INFRA_KUBERNETES_WORKSPACE__UNRECOVERABLE__EVENTS": "FailedMount,FailedScheduling,MountVolume.SetUp failed,Failed to pull image",
-		"CHE_LIMITS_WORKSPACE_IDLE_TIMEOUT":                     "-1",
 	}
 	return cheEnv
 
--- a/pkg/deploy/che_configmap.go
+++ b/pkg/deploy/che_configmap.go
@@ -69,11 +69,7 @@ type CheConfigMap struct {
 func GetCustomConfigMapData() (cheEnv map[string]string) {
 
 	cheEnv = map[string]string{
-		"CHE_PREDEFINED_STACKS_RELOAD__ON__START":               "true",
 		"CHE_INFRA_KUBERNETES_SERVICE__ACCOUNT__NAME":           "che-workspace",
-		"CHE_WORKSPACE_AUTO_START":                              "true",
-		"CHE_INFRA_KUBERNETES_WORKSPACE__UNRECOVERABLE__EVENTS": "FailedMount,FailedScheduling,MountVolume.SetUp failed,Failed to pull image",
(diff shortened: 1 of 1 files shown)


pkg/deploy/che_configmap.go
```diff
@@ -73,7 +73,6 @@ func GetCustomConfigMapData() (cheEnv map[string]string) {
 		"CHE_INFRA_KUBERNETES_SERVICE__ACCOUNT__NAME":           "che-workspace",
 		"CHE_WORKSPACE_AUTO_START":                              "true",
 		"CHE_INFRA_KUBERNETES_WORKSPACE__UNRECOVERABLE__EVENTS": "FailedMount,FailedScheduling,MountVolume.SetUp failed,Failed to pull image",
-		"CHE_LIMITS_WORKSPACE_IDLE_TIMEOUT":                     "-1",
```
sleshchenko: I'm afraid that this change won't be applied to already deployed Che where custom config map is already created.
If it won't - should we care about existing installation where the limit will be still -1?
Do we really need put any configuration here if we are not able to propagate new defaults? I predict that default `UNRECOVERABLE__EVENTS` will be changed soon as well... So, maybe it would be better to create empty custom config map?
cc @davidfestal

davidfestal:
In fact these values in the `custom` config are quite old and were put there by the operator long ago, at the beginning. In fact even the `custom` config map (which is a very weird way to allow customizing Che properties) would disappear after issue https://github.com/eclipse/che/issues/14635 is implemented.

And in the context of issue https://github.com/eclipse/che/issues/14635, the `custom` config map will be removed from all the namespaces managed by the `che-operator`, and the existing CM values would be added in the new `overrideCheProperties` field of the `CheCluster` custom resource.

So if some of the values that were added when creating the `custom` config map, please comment on issue https://github.com/eclipse/che/issues/14635 that they should *not* be pul into the `overrideCheProperties` field of the `CheCluster` custom resource.

davidfestal:
cc @tomgeorge who nearly finished implementing issue https://github.com/eclipse/che/issues/14635

skabashnyuk:
I think in context https://github.com/eclipse/che/issues/14614 we are talking about new installations

davidfestal:
> I think in context [eclipse/che#14614](https://github.com/eclipse/che/issues/14614) we are talking about new installations
> If it's important only for new installation - then LGTM

OK. But we should proably think about the update of existing installations made from previous operator releases, that will still hold old Che server properties in either their `custom` config map` or the new `overrideCheProperties` CR field.
IMO the automatic migration from the `custom` config map to the new `overrideCheProperties` CR field would be the best moment to automatically update old obsolete Che properties that are not expected anymore.

davidfestal:
Approving this PR in the meantime.

pkg/deploy/che_configmap.go
davidfestal: yes, it will only be effective for new installations, until issue https://github.com/eclipse/che/issues/14635 is implemented.

skabashnyuk:
СС @l0rd

l0rd:
I think that it may be a misunderstanding. When I said that the operator/helm chart should not override the defaults I meant that they should not change the default `che.properties` value. Hence I think that in this case it would be better to explicitly set 30 minutes.

skabashnyuk:
CC @sleshchenko @davidfestal I've made changes according to the latest @l0rd 's comment https://github.com/eclipse/che-operator/pull/86#issuecomment-538038245

davidfestal:
> I think that it may be a misunderstanding. When I said that the operator/helm chart should not override the defaults I meant that they should not change the default che.properties value. Hence I think that in this case it would be better to explicitly set 30 minutes.

> CC @sleshchenko @davidfestal I've made changes according to the latest @l0rd 's comment

@l0rd @skabashnyuk 

It seems to me that both changes would have the same result : if the `che.properties` file in the Che server contains a 30 minutes idle timeout, it will be automatically used if we remove the corresponding value in the `custom` config map. I don't see how setting it explicitly in the `custom` config map adds value here.

Ideally we should tend to have an empty `custom` config map by default. The upstream default properties contained in the `che.properties` should be sufficient to run without overriding anything.

IMO adding a non-empty `custom` config map by default at installation is a confusing design error from the beginning.

skabashnyuk:
@davidfestal @sleshchenko I've removed some defaults except one. This will make che-operator configuration is close to Che upstream. Are you ok with that?

davidfestal:
Perfect !

PR #87, opened 2019-10-03
Add CustomCheProperties field

tomgeorge:
This PR implements eclipse/che#14635

+ Adds a new field, under `spec.server` called `overrideCheProperties`.  Anything in `overrideCheProperties` will update the `che` ConfigMap with it's new value.
+ Anything that is found in the `custom` ConfigMap is merged into `overrideCheProperties`

Docs PR: https://github.com/eclipse/che-docs/pull/838


pkg/apis/org/v1/che_types.go
```diff
@@ -98,6 +100,15 @@ type CheClusterSpecServer struct {
 	ServerMemoryLimit string `json:"serverMemoryLimit"`
 }
 
+// ChePropertyOverride is a key-value pair of environment variables to
+// override for the default che configmap
```
davidfestal: override **or add** in the che config map

pkg/apis/org/v1/che_types.go
```diff
@@ -81,6 +81,8 @@ type CheClusterSpecServer struct {
 	// By default a dedicated plugin registry server is started.
 	// But if ExternalPluginRegistry is `true`, then no such dedicated server will be started by the operator
 	ExternalPluginRegistry bool `json:"externalPluginRegistry"`
+	// OverrideCheProperties is a list of EnvVars of Che properties to override from their default value
```
davidfestal: As a comment, I would put something like:

OverrideCheProperties is a list of additional environment variables that will be applied in the che config map, in addition to the values already generated from other fields of the custom resource (CR).
If OverrideCheProperties contains a property that would be normally generated in che config map from other CR fields, then the value in the OverrideCheProperties will be used.

pkg/controller/che/che_controller.go
```diff
@@ -842,19 +841,26 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
-			Kind: "ConfigMap"},
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "custom",
-			Namespace: instance.Namespace,
-			Labels:    cheLabels}}
-	if err := r.CreateNewConfigMap(instance, customCM); err != nil {
-		return reconcile.Result{}, err
+	// Get custom ConfigMap
+	// if it exists, add the data into OverrideCheProperties
+	customConfigMap := &corev1.ConfigMap{}
```
davidfestal: I would move this at the beginning of the reconcile method, to line https://github.com/eclipse/che-operator/pull/87/files#diff-d9445e614f0d728fd140771b0fd29c50L243

So it would be the first thing done.

pkg/controller/che/che_controller.go
```diff
@@ -842,19 +841,26 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+	err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: instance.Namespace, Name: "custom"}, customConfigMap)
+	if !errors.IsNotFound(err) {
```
davidfestal: Did you test what happens when `err == nil` ?
I would add a test here:
```go
if err != nil && !errors.IsNotFound(err) {
```

pkg/apis/org/v1/che_types.go
```diff
@@ -81,6 +81,8 @@ type CheClusterSpecServer struct {
...
+	OverrideCheProperties []ChePropertyOverride `json:"overrideCheProperties"`
```
davidfestal: Did you try using a `map[string]string` ?

That could:
- avoid the risk of adding several times the same properties with different values
- simplify some parts of your code.

pkg/controller/che/che_controller.go
tomgeorge: Do you want me to just move the declaration of the config map, or the whole reconciliation logic piece?

davidfestal: I was meaning move this whole block https://github.com/eclipse/che-operator/pull/87/files#diff-d9445e614f0d728fd140771b0fd29c50R846-R863 :

The maintenance of the `custom` config map.

davidfestal:
@rkratky @themr0c As soon as the main content of the linked docs PR is OK for you (not necessarily the final wording or formatting), could you approve this PR or comment on it ?

pkg/controller/che/che_controller.go
```diff
@@ -254,6 +253,34 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+	}
+	if err == nil {
+		logrus.Infof("Found legacy custom ConfigMap.  Adding those values to CheCluster.Spec.Server.CustomCheProperties")
+		if instance.Spec.Server.CustomCheProperties == nil {
+			instance.Spec.Server.CustomCheProperties = make(map[string]string)
+		}
+		for k, v := range customConfigMap.Data {
+			instance.Spec.Server.CustomCheProperties[k] = v
+		}
+		if err = r.client.Delete(context.TODO(), customConfigMap); err != nil {
```
davidfestal: I think I would delete *after* having successfully updated the custom resource with the `customConfigMap`.
Else we have a risk to loose the content of the old `custom` config map if an error occurs during the update just below

tomgeorge:
@nickboldt `chectl` with `--installer=operator --che-operator-cr-yaml` works

nickboldt:
@tomgeorge can you attach your cr.yaml file (or throw it in a gist) ? might be useful to turn it into an e2e test?

tomgeorge:
@nickboldt I used this gist https://gist.github.com/tomgeorge/0760d8d629919da0b7c3f08129fd2828.  It has throwaway custom values, but I also created a `custom` CM before installing, and those properties were overriden after everything was deployed.

PR #91, opened 2019-10-08
Add airgap mode

tomgeorge:
This PR addresses https://github.com/eclipse/che/issues/14734

- Add 3 CR fields, `airGapMode`, `airGapContainerRegistryHostname`, `airGapContainerRegistryRepository`
- If `airGapMode` is true, all of the images will be pulled from their airgapped repository location instead of `registry.redhat.io`, `docker.io`, etc.

We are thinking that this doesn't need a docs PR as this will mostly be covered in https://github.com/eclipse/che/issues/13886

### DIFF
--- a/pkg/apis/org/v1/che_types.go
+++ b/pkg/apis/org/v1/che_types.go
@@ -31,12 +31,10 @@ type CheClusterSpec struct {
 }
 
 type CheClusterSpecServer struct {
-	// AirGapMode is a flag to tell Che server that it is running in an air-gapped environment
-	AirGapMode bool `json:"airGapMode"`
 	// AirGapContainerRegistryHostname is the hostname to the internal registry to pull images from in the air-gapped environment
 	AirGapContainerRegistryHostname string `json:"airGapContainerRegistryHostname"`
-	// AirGapContainerRegistryRepository is the repository name in the registry to pull images from in the air-gapped environment
-	AirGapContainerRegistryRepository string `json:"airGapContainerRegistryRepository"`
+	// AirGapContainerRegistryOrganization is the repository name in the registry to pull images from in the air-gapped environment
+	AirGapContainerRegistryOrganization string `json:"airGapContainerRegistryOrganization"`
 	// CheImage is a server image used in Che deployment
 	CheImage string `json:"cheImage"`
 	// CheImageTag is a tag of an image used in Che deployment
@@ -254,7 +252,6 @@ func init() {
 }
 
 func (c *CheCluster) IsAirGapMode() bool {
-	return c.Spec.Server.AirGapMode &&
-		c.Spec.Server.AirGapContainerRegistryHostname != "" &&
-		c.Spec.Server.AirGapContainerRegistryRepository != ""
+	return c.Spec.Server.AirGapContainerRegistryHostname != "" ||
+		c.Spec.Server.AirGapContainerRegistryOrganization != ""
 }
--- a/pkg/apis/org/v1/che_types.go
+++ b/pkg/apis/org/v1/che_types.go
@@ -20,6 +20,7 @@ import (
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // CheClusterSpec defines the desired state of CheCluster
+// +k8s:openapi-gen=true
 type CheClusterSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
@@ -27,9 +28,10 @@ type CheClusterSpec struct {
 	Database CheClusterSpecDB      `json:"database"`
 	Auth     CheClusterSpecAuth    `json:"auth"`
 	Storage  CheClusterSpecStorage `json:"storage"`
-	K8SOnly  CheClusterSpecK8SOnly `json:"k8s"`
+	K8s      CheClusterSpecK8SOnly `json:"k8s"`
 }
 
+// +k8s:openapi-gen=true
 type CheClusterSpecServer struct {
 	// AirGapMode is a flag to tell Che server that it is running in an air-gapped environment
 	AirGapMode bool `json:"airGapMode"`
@@ -109,6 +111,7 @@ type CheClusterSpecServer struct {
 	ServerMemoryLimit string `json:"serverMemoryLimit"`
 }
 
+// +k8s:openapi-gen=true
 type CheClusterSpecDB struct {
 	// ExternalDB instructs the operator either to skip deploying Postgres,
 	// and passes connection details of existing DB to Che server (when set to true)
@@ -130,6 +133,7 @@ type CheClusterSpecDB struct {
 	PostgresImagePullPolicy corev1.PullPolicy `json:"postgresImagePullPolicy"`
 }
 
+// +k8s:openapi-gen=true
 type CheClusterSpecAuth struct {
 	// ExternalKeycloak instructs operator on whether or not to deploy Keycloak/RH SSO instance. When set to true provision connection details
 	ExternalKeycloak bool `json:"externalIdentityProvider"`
@@ -162,6 +166,7 @@ type CheClusterSpecAuth struct {
 	KeycloakImagePullPolicy corev1.PullPolicy `json:"identityProviderImagePullPolicy"`
 }
 
+// +k8s:openapi-gen=true
 type CheClusterSpecStorage struct {
 	// PvcStrategy is a persistent volume claim strategy for Che server. Can be common (all workspaces PVCs in one volume),
 	// per-workspace (one PVC per workspace for all declared volumes) and unique (one PVC per declared volume). Defaults to common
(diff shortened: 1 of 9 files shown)


pkg/apis/org/v1/che_types.go
```diff
@@ -20,16 +20,18 @@ import (
...
+// +k8s:openapi-gen=true
 type CheClusterSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	Server   CheClusterSpecServer  `json:"server"`
 	Database CheClusterSpecDB      `json:"database"`
 	Auth     CheClusterSpecAuth    `json:"auth"`
 	Storage  CheClusterSpecStorage `json:"storage"`
-	K8SOnly  CheClusterSpecK8SOnly `json:"k8s"`
+	K8s      CheClusterSpecK8SOnly `json:"k8s"`
```
davidfestal: Don't you need to change something in the rest of the code ta adapt to this change ?

BTW do you think it would be possible to a distinct PR for openapi support ?

tomgeorge: oh whoops, I will change that back

pkg/deploy/che_configmap.go
```diff
@@ -65,6 +65,11 @@ type CheConfigMap struct {
 	PluginRegistryUrl                    string `json:"CHE_WORKSPACE_PLUGIN__REGISTRY__URL,omitempty"`
 	DevfileRegistryUrl                   string `json:"CHE_WORKSPACE_DEVFILE__REGISTRY__URL,omitempty"`
 	WebSocketEndpointMinor               string `json:"CHE_WEBSOCKET_ENDPOINT__MINOR"`
+	CheWorkspacePluginBrokerInitImage    string `json:"CHE_WORKSPACE_PLUGIN__BROKER_INIT_IMAGE,omitempty"`
+	CheWorkspacePluginBrokerUnifiedImage string `json:"CHE_WORKSPACE_PLUGIN__BROKER_UNIFIED_IMAGE,omitempty"`
+	CheServerSecureExposerJwtProxyImage  string `json:"CHE_SERVER_SECURE__EXPOSER_JWTPROXY_IMAGE,omitempty"`
+	CheWorkspaceSidecarImagePullPolicy   string `json:"CHE_WORKSPACE_SIDECAR_IMAGE__PULL__POLICY,omitempty"`
```
davidfestal: > CheWorkspaceSidecarImagePullPolicy
> CheDockerAlwaysPullImage

@l0rd @nickboldt Do you know if Airgap should override these properties to force a given value ?

pkg/deploy/che_configmap.go
```diff
@@ -154,6 +159,8 @@ func GetConfigMapData(cr *orgv1.CheCluster) (cheEnv map[string]string) {
 		CheWebSocketEndpoint:                 wsprotocol + "://" + cheHost + "/api/websocket",
 		WebSocketEndpointMinor:               wsprotocol + "://" + cheHost + "/api/websocket-minor",
 		CheDebugServer:                       cheDebug,
+		CheDockerAlwaysPullImage:             cheDockerAlwaysPullImage,
+		CheWorkspaceSidecarImagePullPolicy:   cheWorkspaceSidecarImagePullPolicy,
```
davidfestal: > CheWorkspaceSidecarImagePullPolicy
> CheDockerAlwaysPullImage

IMO we shouldn't override it by default with a constant, only when airgap is enabled, and possibly with a different value than the default one provided by `che.properties`

I'm not even sure that we have to override them for airgap

pkg/deploy/che_configmap.go
```diff
@@ -225,3 +235,14 @@ func NewCheConfigMap(cr *orgv1.CheCluster, cheEnv map[string]string) *corev1.Con
 		Data: cheEnv,
 	}
 }
+
+func extraImagesConfig(cr *orgv1.CheCluster) map[string]string {
+	extraImages := map[string]string{
+		"CHE_WORKSPACE_PLUGIN__BROKER_INIT_IMAGE":    patchDefaultImageName(cr, cheWorkspacePluginBrokerInitImage),
+		"CHE_WORKSPACE_PLUGIN__BROKER_UNIFIED_IMAGE": patchDefaultImageName(cr, cheWorkspacePluginBrokerUnifiedImage),
+		"CHE_SERVER_SECURE__EXPOSER_JWTPROXY_IMAGE":  patchDefaultImageName(cr, cheServerSecureExposerJwtProxyImage),
+		"CHE_WORKSPACE_SIDECAR_IMAGE__PULL__POLICY":  cheWorkspaceSidecarImagePullPolicy,
```
davidfestal: > CheWorkspaceSidecarImagePullPolicy
> CheDockerAlwaysPullImage

it seems you don't do anything with these values here since you already added them in the main config map building code.
Either we given them a specific value for airgap, or we can just remove them completely and the default will be used by the Che server from the `che.properties`.

pkg/deploy/che_configmap.go
nickboldt: yes, otherwise it'll try to pull from dockerhub/quay/rhoi repos.

pkg/deploy/che_configmap.go
nickboldt: yeah we might be OK here since the assumption is that the newer images will also be inside the airgap environment. 

As long as the image r.r.io/crw/server:2.0 -> myquay.bigcorp.com/crw-stuff/crw-server:2.0 and newer ones are managed under the same tag... then no policy change should be needed.

I mean pullpolicy of Always pull might annoy teh devops team at bigcorp as it's extra network traffic on every startup but that'd be up to them to decide when installing Che/CRW in teh first place... right?

davidfestal: yeah. So let not override those at all for now

davidfestal:
I would say LGTM unless you @l0rd and @nickboldt still want to approve before.

pkg/deploy/extra_images.go
```diff
@@ -0,0 +1,8 @@
+// This file is generated, and contains the latest versions of certain properties from che.properties
+package deploy
+
+const (
+	cheWorkspacePluginBrokerInitImage    = "eclipse/che-init-plugin-broker:v0.21"
```
nickboldt: I would prefer if these contained the registry in the URL (docker.io) as a reminder of things we need to migrate to quay.

Also don't love that jwtproxy does SHA releases, not numbered ones (7.3.0). But those are out of scope for this issue. :)

pkg/deploy/che_configmap.go
```diff
@@ -67,7 +67,9 @@ type CheConfigMap struct {
 	WebSocketEndpointMinor               string `json:"CHE_WEBSOCKET_ENDPOINT__MINOR"`
 	CheWorkspacePluginBrokerInitImage    string `json:"CHE_WORKSPACE_PLUGIN__BROKER_INIT_IMAGE,omitempty"`
 	CheWorkspacePluginBrokerUnifiedImage string `json:"CHE_WORKSPACE_PLUGIN__BROKER_UNIFIED_IMAGE,omitempty"`
-	CheServerSecureExposesJwtProxyImage  string `json:"CHE_SERVER_SECURE__EXPOSER_JWTPROXY_IMAGE,omitempty"`
+	CheServerSecureExposerJwtProxyImage  string `json:"CHE_SERVER_SECURE__EXPOSER_JWTPROXY_IMAGE,omitempty"`
+	CheWorkspaceSidecarImagePullPolicy   string `json:"CHE_WORKSPACE_SIDECAR_IMAGE__PULL__POLICY,omitempty"`
+	CheDockerAlwaysPullImage             string `json:"CHE_DOCKER_ALWAYS__PULL__IMAGE,omitempty"`
```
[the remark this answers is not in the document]
sleshchenko: And probably override for [plugin_broker.pull_policy](https://github.com/eclipse/che/blob/36fe587e28fba1e37d9f6b38da673cf917a5132f/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties#L633) should be added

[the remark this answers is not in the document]
sleshchenko: maybe imagePullPolicy for jwtproxy should be overridden as well https://github.com/eclipse/che/blob/154d668505cc86763f16ad8868e0a1fcf4bccc28/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/secure/jwtproxy/JwtProxyProvisioner.java#L287

Some of my messages may be out of the topic since I'm not familiar with airgap topic but I have seen familiar words and thought maybe by shared knowledge can be useful here =)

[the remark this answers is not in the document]
davidfestal: After speaking with @nickboldt we had concluded that the airgap scenario doesn't need any overriding of the default pull policies. Even a `Always` pull policy would work in airgap mode, since we still point to a valid container registry.

So I'd better leave everything non-airgap-related out of the scope of this issue / PR.

Do you agree @sleshchenko ?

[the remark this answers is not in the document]
davidfestal: BTW both properties:
```
CHE_WORKSPACE_SIDECAR_IMAGE__PULL__POLICY
CHE_DOCKER_ALWAYS__PULL__IMAGE
```
Have already removed by Tom.

l0rd:
Good job @tomgeorge. Would you mind updating the description of this PR? I think that you are not using the `airgapMode` attribute anymore. For future references it's better if the description corresponds to what has been merged.

tomgeorge:
@l0rd I updated the description.  Thanks!

PR #93, opened 2019-10-11
Configure the operator to deploy airgapped versions of the devfile and plugin registries

tomgeorge:
This PR changes the operator to configure the devfile and plugin registries to work with airgap mode as per eclipse/che#14734.

It creates two configmaps, devfile-registry and plugin-registry, and adds the values of `airGapContainerRegistryHostname` and `airGapContainerRegistryOrganization` to it, and sets the deployments to optionally reference those configmaps.


pkg/apis/org/v1/zz_generated.deepcopy.go
```diff
@@ -1,22 +1,6 @@
...
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
-
-// Code generated by deepcopy-gen. DO NOT EDIT.
+// Code generated by operator-sdk. DO NOT EDIT.
```
davidfestal: Could you add the license back ?

pkg/controller/che/che_controller.go
```diff
@@ -796,6 +812,27 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 	// Create devfile registry resources unless an external registry is used
 	externalDevfileRegistry := instance.Spec.Server.ExternalDevfileRegistry
 	if !externalDevfileRegistry {
+		if instance.IsAirGapMode() {
```
davidfestal: Don't you undate the config map in case it already exists ?
It seems we should, right ?

pkg/deploy/registry_configmap.go
```diff
@@ -0,0 +1,97 @@
...
+		Data: GetPluginRegistryConfigMapData(cr),
+	}
+}
+
+func GetDevfileRegistryConfigMapData(cr *orgv1.CheCluster) map[string]string {
+	devfileRegistryEnv := make(map[string]string)
+	data := &DevFileRegistryConfigMap{
+		CheDevfileImagesRegistryURL:          cr.Spec.Server.AirGapContainerRegistryHostname,
+		CheDevfileImagesRegistryOrganization: cr.Spec.Server.AirGapContainerRegistryOrganization,
+		CheDevfileImagesRegistryTag:          "latest",
```
davidfestal: why do you add the `latest` ? We should remove it afail.

PR #94, opened 2019-10-11
Update custom resource doc, and make it available in OpenApi spec

tomgeorge:
#### Description

This PR :
- adds an openapi schema to the `CheCluster` CRD, allowing for completion and validation in the OpenShift 4.2 console.
- updates the comments of each Custom Resource field so that they can be the basis of the official `CheCluster` custom resource documentation.  

####

This is the implementarion part of issue https://github.com/eclipse/che/issues/13886#issuecomment-534171945

#### NOTE related to docs:

@rkratky 

The updated documentation comments added in file [che-types.go](https://github.com/eclipse/che-operator/pull/94/files#diff-293dad4e0a11c8e239f117f58789c653) contain the description of the content that should also be documented in the Che documentation (by the docs team ?).


pkg/apis/org/v1/che_types.go
```diff
@@ -19,29 +19,34 @@ import (
...
 }
 
+// +k8s:openapi-gen=true
 type CheClusterSpecServer struct {
 	// AirGapContainerRegistryHostname is the hostname to the internal registry to pull images from in the air-gapped environment
 	AirGapContainerRegistryHostname string `json:"airGapContainerRegistryHostname"`
 	// AirGapContainerRegistryOrganization is the repository name in the registry to pull images from in the air-gapped environment
 	AirGapContainerRegistryOrganization string `json:"airGapContainerRegistryOrganization"`
 	// CheImage is a server image used in Che deployment
+	// +optional
```
davidfestal: in addition to the `// +optional` comment, you can add the `,omitempty` marker in the json tag, like that:
```
	CheImage string `json:"cheImage,omitempty"`
```
Of course, only on values that will have defauts, which is the case for most of them.

PR #100, opened 2019-10-21
Append `/v3` to pluginRegistryURL in all cases.

rhopp:
In history it was not appended, when flavor was codeready (backwards compatibility). I think this is not needed anymore.

Should fix https://issues.jboss.org/browse/CRW-430

Signed-off-by: Radim Hopp <rhopp@redhat.com>

### DIFF
--- a/pkg/controller/che/che_controller.go
+++ b/pkg/controller/che/che_controller.go
@@ -941,9 +941,7 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 		if err != nil {
 			return reconcile.Result{}, err
 		}
-		if cheFlavor != "codeready" {
-			guessedPluginRegistryURL += "/v3"
-		}
+		guessedPluginRegistryURL += "/v3"
 		if pluginRegistryURL == "" {
 			pluginRegistryURL = guessedPluginRegistryURL
 		}


ibuziuk:
@rhopp do we know why initially the `v3` was not appended to the codeready? In the PR description you wrote about backwards compatibility, but I'm not really getting how was it achieved

rhopp:
@ibuziuk My suspicion is, that CRW 1.2 was still expecting the old version of plugin-registry format (served at `/`). So this if was there to support both upstream che (which expects new version) and CRW 1.2 (which expected old format).
Now, when we have CRE 2.0, based on Che 7, it expects new format of plugin registry (`/v3`), but because of this `if` statement, the `/v3` is not appended to URL).

I hope, that this version of operator doesn't have to support CRW 1.x as well. If yes, the refatoring would have to be bigger imo.

ibuziuk:
> CRW 1.2 was still expecting the old version of plugin-registry format (served at /). So this if was there to support both upstream che (which expects new version) and CRW 1.2 (which expected old format).

I'm getting lost. CRW 1.2 is Che 6 based. Why would it expect plugin-registry at all (devfile / plugin registries are Che 7 concepts which do not exist in the Che 6 world).

How this PR was tested?

rhopp:
@ibuziuk Even though it was not supported, "che 7" workspaces could be started in CRW 1.2 (the same way they could be started in Che 6.19). That's why the plugin/devfile registry was there I think.

How was it tested?
I've build my own image (quay.io/rhopp/che-operator:crw2) and started CRW 2.0 with it:
```
bin/run server:start -a operator -p openshift -b rhopp-qa.devcluster.openshift.com -n crw2-fix --che-operator-image=quay.io/rhopp/che-operator:crw2 --che-operator-cr-yaml=/home/rhopp/temp/crw2/crw2-cr.yaml
```

Where `crw2-cr.yaml` is this: https://gist.github.com/rhopp/474e28a729b1afccf221b62e0444fc4f (with one difference... I've changed cheImageTag to be older, because current one has bug and the server doesn't start - https://github.com/redhat-developer/codeready-workspaces/pull/181)

Once this CRW was deployed, I tried to start workspace.
The startup failed, but on different error, than before. And I think this error in unrelated to this change:
image
From this screenshot it's obvious it successfully started downloading/processing plugins from correct URL.
If you compare it with original error from https://issues.jboss.org/browse/CRW-430, where it's trying to download the plugin meta.yaml from URL without `/v3` in it.

ibuziuk:
> Even though it was not supported, "che 7" workspaces could be started in CRW 1.2 (the same way they could be started in Che 6.19). That's why the plugin/devfile registry was there I think.

Yeah, maybe it makes sense and back to 6.19 there was only the initial version of plugins, but it is better to double-check it in order to have a clear understanding why the `v3` was skipped initially for the codeready flavour

> The startup failed, but on different error, than before. And I think this error in unrelated to this change

from the error message, It looks like the outdated version of plugin brokers is used - `v0.23.0` is the latest one. Which one is currently used by CRW ?

amisevsk:
>Once this CRW was deployed, I tried to start workspace.
The startup failed, but on different error, than before. And I think this error in unrelated to this change:

The error message in the screenshot is because the plugin broker used in CRW was forked from plugin broker `v0.19.0`; offline support (the `relative:extension/` format) was added in `v0.20.0`.

rhopp:
@amisevsk Yes... But that's another problem I'm not trying to solve in this PR.
The original problem this PR is for is https://issues.jboss.org/browse/CRW-430.

Solution to the plugin broker issue is rebuilding of pluginbroker image (the correct sources are already in place), but I don't know how to do that. I'm pinging Nick to respin that image.

PR #101, opened 2019-10-22
che #14809 Enable CodeReady branding on the ConsoleLink elements created by the che operator with 'codeready' flavor

ibuziuk:
Che issue - https://github.com/eclipse/che/issues/14809

ConsoleLink functionality works just fine

image

Verified against crc:

`./run server:start -a operator -p crc -b console-openshift-console.apps-crc.testing -n crw-link --che-operator-image=quay.io/ibuziuk/che-operator:che-14809-new --che-operator-cr-yaml=crw.yaml`

crw.yaml:

```
kind: CheCluster 
metadata:
  name: eclipse-che
spec:
  server:
    # server image used in Che deployment
    cheImage: 'quay.io/crw/server-rhel8'
    # tag of an image used in Che deployment
    cheImageTag: '2.0-377'
    # image:tag used in Devfile registry deployment
    devfileRegistryImage: 'quay.io/crw/devfileregistry-rhel8:2.0-159'
    # image:tag used in plugin registry deployment
    pluginRegistryImage: 'quay.io/crw/pluginregistry-rhel8:2.0-203'
    # defaults to `che`. When set to `codeready`, CodeReady Workspaces is deployed
    # the difference is in images, labels, exec commands
    cheFlavor: 'codeready'
    # when set to true the operator will attempt to get a secret in OpenShift router namespace
    # to add it to Java trust store of Che server. Requires cluster-admin privileges for operator service account
    selfSignedCert: true
    # TLS mode for Che. Make sure you either have public cert, or set selfSignedCert to true
    tlsSupport: true
    # protocol+hostname of a proxy server. Automatically added as JAVA_OPTS and https(s)_proxy
    # to Che server and workspaces containers
    # proxyURL: ''
    # port of a proxy server
    # proxyPort: ''
    # username for a proxy server
    # proxyUser: ''
    # password for a proxy user
    # proxyPassword: ''
    # a list of non-proxy hosts. Use | as delimiter, eg localhost|my.host.com|123.42.12.32
    # nonProxyHosts: ''
    # overrides for https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
    customCheProperties:
      CHE_WORKSPACE_SIDECAR_IMAGE__PULL__POLICY: 'Always'
      CHE_DOCKER_ALWAYS__PULL__IMAGE: 'true'
      # set by operator, probably don't need to override. Latest in https://github.com/eclipse/che-operator/blob/master/pkg/deploy/defaults.go#L47-L51
      # CHE_INFRA_KUBERNETES_PVC_JOBS_IMAGE: 'registry.access.redhat.com/ubi8-minimal:8.0-213'
      CHE_WORKSPACE_PLUGIN__BROKER_INIT_IMAGE: 'quay.io/crw/pluginbrokerinit-rhel8:2.0-6'
      CHE_WORKSPACE_PLUGIN__BROKER_UNIFIED_IMAGE: 'quay.io/crw/pluginbroker-rhel8:2.0-5'
      CHE_SERVER_SECURE__EXPOSER_JWTPROXY_IMAGE: 'quay.io/crw/jwtproxy-rhel8:2.0-4'
      CHE_FACTORY_DEFAULT__EDITOR: 'eclipse/che-theia/7.3.0'
      CHE_FACTORY_DEFAULT__PLUGINS: 'eclipse/che-machine-exec-plugin/7.3.0'
      CHE_WORKSPACE_DEVFILE_DEFAULT__EDITOR: 'eclipse/che-theia/7.3.0'
      CHE_WORKSPACE_DEVFILE_DEFAULT__EDITOR_PLUGINS: 'eclipse/che-machine-exec-plugin/7.3.0'
    # when set to true, the operator skips deploying Postgres, and passes connection details of existing DB to Che server
    # otherwise a Postgres deployment is created
    # externalDb: false
    # Postgres deployment in format image:tag. Defaults to registry.redhat.io/rhscl/postgresql-96-rhel7 (see pkg/deploy/defaults.go for latest tag)
    # set by operator, probably don't need to override. Latest in https://github.com/eclipse/che-operator/blob/master/pkg/deploy/defaults.go#L47-L51
    # postgresImage: 'registry.redhat.io/rhscl/postgresql-96-rhel7:1-47'
  # storage:
    # persistent volume claim strategy for Che server. Can be common (all workspaces PVCs in one volume),
    # per-workspace (one PVC per workspace for all declared volumes) and unique (one PVC per declared volume). Defaults to common
    # pvcStrategy: 'per-workspace'
    # size of a persistent volume claim for workspaces. Defaults to 1Gi
    # pvcClaimSize: '1Gi'
    # instruct Che server to launch a special pod to precreate a subpath in a PV
    # preCreateSubPaths: true
    # image:tag for preCreateSubPaths jobs
    # set by operator, probably don't need to override. Latest in https://github.com/eclipse/che-operator/blob/master/pkg/deploy/defaults.go#L47-L51
    # pvcJobsImage: 'registry.access.redhat.com/ubi8-minimal:8.0-213'
  # auth:
    # instructs operator on whether or not to deploy Keycloak/RH SSO instance. When set to true provision connection details
    # externalIdentityProvider: false
    # retrieved from respective route/ingress unless explicitly specified in CR (when ExternalKeycloak is true)
    # instructs an Operator to enable OpenShift v3 identity provider in Keycloak,
    # as well as create respective oAuthClient and configure Che configMap accordingly
    # openShiftoAuth: false
    # image:tag used in Keycloak deployment
    # set by operator, probably don't need to override. Latest in https://github.com/eclipse/che-operator/blob/master/pkg/deploy/defaults.go#L47-L51
    # identityProviderImage: 'registry.redhat.io/redhat-sso-7/sso73-openshift:1.0-15'
```

### DIFF
--- a/pkg/deploy/defaults.go
+++ b/pkg/deploy/defaults.go
@@ -71,11 +71,19 @@ const (
 	OldDefaultPostgresUpstreamImageToDetect = "centos/postgresql-96-centos7:9.6"
 
 	// ConsoleLink default
-	DefaultConsoleLinkName        = "che"
-	DefaultConsoleLinkImage       = "/dashboard/assets/branding/loader.svg"
-	DefaultConsoleLinkDisplayName = "Eclipse Che"
-	DefaultConsoleLinkSection     = "Red Hat Applications"
+	DefaultConsoleLinkName                = "che"
+	DefaultConsoleLinkSection             = "Red Hat Applications"
+	DefaultConsoleLinkImage               = "/dashboard/assets/branding/loader.svg"
+	defaultConsoleLinkUpstreamDisplayName = "Eclipse Che"
+	defaultConsoleLinkDisplayName         = "CodeReady Workspaces"
 )
+
+func DefaultConsoleLinkDisplayName(cheFlavor string) string {
+	if cheFlavor == "codeready" {
+		return defaultConsoleLinkDisplayName
+	}
+	return defaultConsoleLinkUpstreamDisplayName
+}
 
 func DefaultCheServerImageTag(cheFlavor string) string {
 	if cheFlavor == "codeready" {
--- a/pkg/controller/che/che_controller.go
+++ b/pkg/controller/che/che_controller.go
@@ -1044,6 +1044,7 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 
 	// we can now try to create consolelink, after che instance is available
 	if err := createConsoleLink(isOpenShift4, protocol, instance, r); err != nil {
+		logrus.Errorf("An error occurred during console link provisioning: %s", err)
 		return reconcile.Result{}, err
 	}
 
@@ -1200,7 +1201,7 @@ func createConsoleLink(isOpenShift4 bool, protocol string, instance *orgv1.CheCl
 		// console link is supported only with https
 		return nil
 	}
-
+	cheFlavor := instance.Spec.Server.CheFlavor
 	cheHost := instance.Spec.Server.CheHost
 	preparedConsoleLink := &consolev1.ConsoleLink{
 		ObjectMeta: metav1.ObjectMeta{
@@ -1209,7 +1210,7 @@ func createConsoleLink(isOpenShift4 bool, protocol string, instance *orgv1.CheCl
 		Spec: consolev1.ConsoleLinkSpec{
 			Link: consolev1.Link{
 				Href: protocol + "://" + cheHost,
-				Text: deploy.DefaultConsoleLinkDisplayName},
+				Text: deploy.DefaultConsoleLinkDisplayName(cheFlavor)},
 			Location: consolev1.ApplicationMenu,
 			ApplicationMenu: &consolev1.ApplicationMenuSpec{
 				Section:  deploy.DefaultConsoleLinkSection,


pkg/deploy/defaults.go
```diff
@@ -71,12 +71,28 @@ const (
...
-	DefaultConsoleLinkName        = "che"
-	DefaultConsoleLinkImage       = "/dashboard/assets/branding/loader.svg"
-	DefaultConsoleLinkDisplayName = "Eclipse Che"
-	DefaultConsoleLinkSection     = "Red Hat Applications"
+	DefaultConsoleLinkSection              = "Red Hat Applications"
+	DefaultConsoleLinkImage                = "/dashboard/assets/branding/loader.svg"
+	defaultConsoleLinkCheName              = "che"
+	defaultConsoleLinkCodeReadyName        = "codeready"
+	defaultConsoleLinkCheDisplayName       = "Eclipse Che"
+	defaultConsoleLinkCodeReadyDisplayName = "CodeReady Workspaces"
```
nickboldt: -1, for a few reasons...

a) "CodeReady" is a brand, not a product, so should NOT be used here. Options are "Red Hat CodeReady Workspaces", "CodeReady Workspaces", "Workspaces" or if an internal variable not exposed to the end user in the UI/UX, then CRW is ok. But...

b) the existing convention used in https://github.com/eclipse/che-operator/blob/master/pkg/deploy/defaults.go#L40-L41 to differentiate upstream/downstream versions of variables is with "Upstream" prefix on the variable name, so we could keep using that, eg., `defaultConsoleLinkUpstreamName` and `defaultConsoleLinkName`

where does the `ConsoleLinkName` appear? if it's UI, it shouldn't be "codeready" but rather "workspaces" or "Workspaces" as that's the product's shortname. "CodeReady" includes Studio, Builder, Containers, and more, so we can't just slap that in Workspaces and OCP and assume people will think we mean the Che-based product.

pkg/deploy/defaults.go
```diff
@@ -71,12 +71,28 @@ const (
 	OldDefaultPostgresUpstreamImageToDetect = "centos/postgresql-96-centos7:9.6"
 
 	// ConsoleLink default
-	DefaultConsoleLinkName        = "che"
-	DefaultConsoleLinkImage       = "/dashboard/assets/branding/loader.svg"
-	DefaultConsoleLinkDisplayName = "Eclipse Che"
-	DefaultConsoleLinkSection     = "Red Hat Applications"
+	DefaultConsoleLinkSection              = "Red Hat Applications"
+	DefaultConsoleLinkImage                = "/dashboard/assets/branding/loader.svg"
```
nickboldt: is this the Che image? or the CRW image? I'm guessing if we've built the dashboard correctly in CRW it'll use the CRW one, but just wanted to ask.

ibuziuk: @nickboldt variables names already contain CodeReady, so I just follow the standard that is in place e.g. `defaultCodeReadyServerImageRepo`  https://github.com/eclipse/che-operator/blob/master/pkg/deploy/defaults.go#L24 or `defaultCodeReadyServerImageTag` - https://github.com/eclipse/che-operator/blob/master/pkg/deploy/defaults.go#L24

and I do not plan to do the refactoring. The rule of thumb is either to follow the convention that is already in place (even if you don not like) or proceed with the major refactoring

ibuziuk: > is this the Che image? or the CRW image?

I suppose it should use CRW image but I can not confirm since I can not verify

nickboldt:
Please remove the "WIP" marker before you merge this.

davidfestal:
I plan to review tomorrow morning.

ibuziuk:
@nickboldt well, I plan to verify it first before merging ;-)

pkg/deploy/defaults.go
davidfestal: cc @slemeur

davidfestal: @slemeur Could you provide some guidance here ?

pkg/deploy/defaults.go
davidfestal: @nickboldt doesn't the CRW assembly replace the image with a CRW one during packaging ? And in this case is the *relative path* of the image still the same ?
If it's the case we wouldn't need to change the loader image.

pkg/deploy/defaults.go
davidfestal: Related to comment b, it's right that the most general case for variables is to have the *raw property name* for CRW, and *property name + "upstream"* for Eclipse Che. So to increase consistency we could use this naming scheme here as well.

As for the consoleLinkName, I'm don't think it is displayed to the user, and changing it might break the OS 4.2 integration if they would be searching for the ConsoleLink by name.

In fact looking into the OpenShift 4.2 console, it is clear that **we should not** change the ConsoleLinkName:

https://github.com/openshift/console/blob/f5a2a8b0358398748f301af0c11bd74ac33badce/frontend/packages/dev-console/src/components/topology/topology-utils.ts#L27

davidfestal:
> > and I don't like the variable names
> 
> variables names already contain CodeReady, so I just follow the standard that is in place e.g. defaultCodeReadyServerImageRepo https://github.com/eclipse/che-operator/blob/master/pkg/deploy/defaults.go#L24 or defaultCodeReadyServerImageTag - https://github.com/eclipse/che-operator/blob/master/pkg/deploy/defaults.go#L24
> 
> and I do not plan to do the refactoring. The rule of thumb is either to follow the convention that is already in place (even if you do not like) or proceed with the major refactoring

In fact, afaik, the use of `CodeReady` in the default property names is more the exception (only for the che server component). At least as far as I'm concerned,  I used only the `raw vs xxxUpstreamxxx` convention to add new defaults since I took ownership of the operator code.

davidfestal:
> PR is WIP since I have problems verifying it. Basically, deployment on OCP 4.2 with tls / self-signed certs is failing with: 
> 
> > Using embedded assembly.
> > Found a custom cert. Adding it to java trust store based on /usr/lib/jvm/java-1.8.0/jre/lib/security/cacerts
> > chmod: changing permissions of '/home/jboss/cacerts': Operation not permitted
>
> I tried to deploy via:
> ```
> chectl server:start -a operator -p openshift -b rhopp-qa.devcluster.openshift.com -n my-crw-branding --che-operator-image=quay.io/ibuziuk/che-operator:che-14809 --che-operator-cr-yaml=crw.yaml
> ```
>

@ibuziuk Did you try to deploy it simply through the OperatorHub (from the UI), and then change the docker image to your docker image in the deployed CSV, and operator deployment ?

Not sure CheCtl is the recommended install mode for Openhsift 4.2 that has OperatorHub inside it.

ibuziuk:
@nickboldt @davidfestal PR has been updated and verified against crc.

ConsoleLink functionality works just fine, but the CodeReady image was not applied:

image
Peek 2019-10-24 13-48

@nickboldt will the logo issue be addressed as part of the https://issues.jboss.org/browse/CRW-315?focusedCommentId=13803743&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13803743

slemeur:
The display name should be "CodeReady Workspaces". 
Agree with @ibuziuk we need to apply the proper icon as well.

nickboldt:
> @nickboldt doesn't the CRW assembly replace the image with a CRW one during packaging ? And in this case is the relative path of the image still the same ?
> If it's the case we wouldn't need to change the loader image.

I believe so, yes. @evidolob is investigating the branding problem now, hopefully fixed tomorrow. See https://github.com/eclipse/che/issues/14982

I have no idea which icon is used for the devconsole dropdown but if it's one from the CRW assembly or the crw-theia assembly then we should be OK once branding is fixed.

ibuziuk:
> I have no idea which icon is used for the devconsole dropdown

The icon is `/dashboard/assets/branding/loader.svg`

@rhopp @davidfestal may I get your reviews ;-)

nickboldt:
Ah, well for CRW 2 we're using https://github.com/redhat-developer/codeready-workspaces/blob/master/assembly/codeready-workspaces-assembly-dashboard-war/src/main/webapp/assets/branding/CodeReady_icon_loader.svg not `/assets/branding/loader.svg` as per https://github.com/redhat-developer/codeready-workspaces/blob/master/assembly/codeready-workspaces-assembly-dashboard-war/src/main/webapp/assets/branding/product.json.template#L7 but if we need to rename the file we can do so. :)

ibuziuk:
@nickboldt would be great if you could do the image renaming ;-)
should I create an issue for that?

nickboldt:
No need. Here's a PR: https://github.com/redhat-developer/codeready-workspaces/pull/189/files

See also eclipse/che#14982 and https://issues.jboss.org/browse/CRW-315?focusedCommentId=13804077&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13804077

ibuziuk:
@nickboldt awesome, lemme know against which image I can re-test

nickboldt:
Should be able to check new image names in 2.0-377 or later. Building now, should be there in an hour or so, unless something goes wacky.

https://quay.io/repository/crw/server-rhel8?tag=2.0&tab=tags

ibuziuk:
@nickboldt tested against `2.0-377` and now the image is blank :/

image

sleshchenko:
@ibuziuk If you use self-signed cert - it may be caused by an untrusted host. You can easily figure out it by discovering Developer Tools of browser.

ibuziuk:
ah, that was it - works now

image

PR #102, opened 2019-10-23
Set CHE_DEVFILE_REGISTRY_URL to allow airgap operation

tomgeorge:
This PR adds the `CHE_DEVFILE_HTTPS_ENDPOINT` variable when configuring the registries in airgap mode.  It uses the public-facing route of the devfile registry, and places it in the `devfile-registry` config map, and redeploys a new version of the registry based on changes to the route.

Enables https://github.com/eclipse/che/issues/14880

### DIFF
--- a/pkg/controller/che/che_controller.go
+++ b/pkg/controller/che/che_controller.go
@@ -827,17 +827,16 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 
 	devfileRegistryURL := instance.Spec.Server.DevfileRegistryUrl
 
-	guessedDevfileRegistryURL, err := addRegistryRoute("devfile")
-	if err != nil {
-		return reconcile.Result{}, err
-	}
-	if devfileRegistryURL == "" {
-		devfileRegistryURL = guessedDevfileRegistryURL
-	}
-
 	// Create devfile registry resources unless an external registry is used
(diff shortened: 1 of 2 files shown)


skabashnyuk:
Can this variable contain http link? 
CC @sleshchenko

tomgeorge:
yes, it just needs to be whatever the external URL of the devfile registry you are going to use for airgap mode

tomgeorge:
I marked this as WIP because it is waiting for https://github.com/eclipse/che/pull/14964 to be merged

skabashnyuk:
this https part in name confuses a bit. What if we change it to CHE_DEVFILE_REGISTRY_URL?

tomgeorge:
env var name has been changed to CHE_DEVFILE_REGISTRY_URL

davidfestal:
@skabashnyuk @tomgeorge  isn't this env variable name already used by the devfile registry ? Did you really plan to change the devfile registry to adapt to the a new name ?

ibuziuk:
IMO, it is a good idea to get rid of `HTTPS` even if the changes for devfile registry are required

ibuziuk:
@tomgeorge I believe we can remove WIP and ask for @davidfestal @skabashnyuk review

tomgeorge:
@davidfestal @skabashnyuk removed the WIP tag.  When you get a chance can you please review?

davidfestal:
LGTM

PR #105, opened 2019-10-28
Fix typo in example CR (`chePostgresHostName`)

davidfestal:


nickboldt:
re-read PR title and realized WHY you were asking for this change. Seems legit.

PR #108, opened 2019-10-29
Change tests so that they do not depend on hard coded image names

tomgeorge:
Signed-off-by: Tom George <tg82490@gmail.com>

The previous version of these tests depended on the already-existing image names and tags in `defaults.go`, but these can change with releases.

They have been changed to figure out what the expected images should be based on the variables in `defaults.go`

### DIFF
--- a/pkg/deploy/defaults_test.go
+++ b/pkg/deploy/defaults_test.go
@@ -1,8 +1,10 @@
 package deploy
 
 import (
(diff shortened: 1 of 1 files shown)


davidfestal:
cc @nickboldt @rhopp That will help for future changes in defaults

tomgeorge:
Could just be the author ;)

nickboldt:
nope(it's(the(nested(brackets)))).

PR #115, opened 2019-11-08
Correctly manage `che.properties`-located docker images for CRW (jira crw-468) 

davidfestal:
This PR is a fix in the operator for the issue https://issues.jboss.org/browse/CRW-468

It enables correctly getting CRW-specific images for the 3 containers whose docker images is specified in the upstream Che server `che.properties` file.

### DIFF
--- a/pkg/deploy/defaults.go
+++ b/pkg/deploy/defaults.go
@@ -63,6 +63,15 @@ const (
 	DefaultSecurityContextFsGroup   = "1724"
 	DefaultSecurityContextRunAsUser = "1724"
 
+	// CRW images for that are mentioned in the Che server che.properties
+	// For CRW these should be synced by hand with images stored in RH registries
+	// instead of being synced by script with the content of the upstream `che.properties` file
+	// NB:
+	// The upstream equivent are stored in the generated `extra_images.go` source file.
+	defaultCheWorkspacePluginBrokerInitImage    = "quay.io/crw/pluginbrokerinit-rhel8:2.0-6"
+	defaultCheWorkspacePluginBrokerUnifiedImage = "quay.io/crw/pluginbroker-rhel8:2.0-5"
+	defaultCheServerSecureExposerJwtProxyImage  = "quay.io/crw/jwtproxy-rhel8:2.0-4"
+
 	// This is only to correctly  manage defaults during the transition
 	// from Upstream 7.0.0 GA to the next version
 	// That fixed bug https://github.com/eclipse/che/issues/13714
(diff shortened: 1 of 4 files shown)


pkg/deploy/defaults.go
```diff
@@ -139,6 +148,51 @@ func DefaultDevfileRegistryImage(cr *orgv1.CheCluster, cheFlavor string) string
 	}
 }
 
+func DefaultCheWorkspacePluginBrokerInitImage(cr *orgv1.CheCluster, cheFlavor string) string {
+	if cheFlavor == "codeready" {
+		// In the CRW case, we should always set the plugin broker image in the Che config map
+		return patchDefaultImageName(cr, defaultCheWorkspacePluginBrokerInitImage)
```
dmytro-ndp: There could be `IsAirGapMode` in CRW as well, couldn't it?

pkg/deploy/defaults.go
```diff
@@ -139,6 +148,51 @@ func DefaultDevfileRegistryImage(cr *orgv1.CheCluster, cheFlavor string) string
...
+		if cr.IsAirGapMode() {
+			return patchDefaultImageName(cr, defaultCheWorkspacePluginBrokerInitUpstreamImage)
+		}
+		return ""
+	}
+}
+
+func DefaultCheWorkspacePluginBrokerUnifiedImage(cr *orgv1.CheCluster, cheFlavor string) string {
+	if cheFlavor == "codeready" {
+		// In the CRW case, we should always set the plugin broker image in the Che config map
```
dmytro-ndp: There could be `IsAirGapMode` in CRW as well, couldn't it?

pkg/deploy/defaults.go
```diff
@@ -139,6 +148,51 @@ func DefaultDevfileRegistryImage(cr *orgv1.CheCluster, cheFlavor string) string
...
+			return patchDefaultImageName(cr, defaultCheWorkspacePluginBrokerUnifiedUpstreamImage)
+		}
+		return ""
+	}
+}
+
+func DefaultCheServerSecureExposerJwtProxyImage(cr *orgv1.CheCluster, cheFlavor string) string {
+	if cheFlavor == "codeready" {
+		// In the CRW case, we should always set the jwt-proxy image in the Che config map
+		return patchDefaultImageName(cr, defaultCheServerSecureExposerJwtProxyImage)
```
dmytro-ndp: There could be `IsAirGapMode` in CRW as well, couldn't it?

pkg/deploy/defaults.go
davidfestal: `IsAirGapMode()` is also tested inside the `patchDefaultImageName` method

The difference is that we patch the image name *in all cases* for CRW and *only in case of airgap mode* for upstream Che

dmytro-ndp: Will that method `patchDefaultImageName()` be run after `DefaultCheWorkspacePluginBrokerInitImage()` called?

davidfestal: it is called by it :-)

dmytro-ndp: thank you for explanation

PR #122, opened 2019-11-14
Fix the Openshift OAuth provisioning for CRW, and move provisioning files

davidfestal:
This PR:
- Replaces `jq` tool calls by equivalent `sed` calls
- Moves the `xxx_provision` files out of `deploy` (more for yamls) (should fix new issue https://issues.jboss.org/browse/CRW-482 at the same time) 
- Replace a hard-coded value of the clientId (`che-public`) in the `token-exchange` feature on CRW
- Correctly enables the RHSSO preview features required for `token-exchange` (the way to activate them slightly differs from the upstream Che case) 


templates/oauth_provision
```diff
@@ -25,7 +25,7 @@ default_to_openshift_login() {
 }
 
 enable_openshift_token-exchange() {
-  IDENTITY_PROVIDER_ID=$({{ .Script }} get -r {{ .KeycloakRealm }} identity-provider/instances/{{ .ProviderId }} | jq --raw-output '.internalId')
```
dmytro-ndp: what about external keycloaks with non-default realms?

davidfestal: fixed (copy-paste error)

PR #124, opened 2019-11-18
Update to 7.3.2

dmytro-ndp:
Need to update defaults to point to release of Eclipse Che 7.3.2 and new bits on https://quay.io/organization/crw

(it could also require further update of pluginregistry image as well)


dmytro-ndp:
@davidfestal: that you for notice. 
It has been done.

dmytro-ndp:
@davidfestal: indeed, there is `quay.io/crw/pluginbrokerinit-rhel8:2.0-6` only https://github.com/eclipse/che-operator/pull/124

PR #126, opened 2019-11-20
Add environment variable for JGroups labels

mshaposhnik:
JGroups is using labels to detect pods where replication should be enabled (to avoid pinging all containers in deployment). Since CRW using different value for `app` and `component`  labels, they should be set via environment variable.


davidfestal:
Merging since we tested it together with @mshaposhnik on the QE cluster

PR #128, opened 2019-11-22
Align JGroups labels property name with original one in KUBE_PING

mshaposhnik:
Align JGroups labels property name with original one in `KUBE_PING`
(see https://github.com/jgroups-extras/jgroups-kubernetes/blob/master/src/main/java/org/jgroups/protocols/kubernetes/KUBE_PING.java#L81) 
So possible naming confusions may be avoided in future.
Issue to `KUBE_PING` upstream to align their documentation: https://github.com/jgroups-extras/jgroups-kubernetes/issues/79


nickboldt:
Does this typo affect anything? Blocker for CRW 2.0 ?

mshaposhnik:
not really a blocker but better to have it to avoid possible confusions

davidfestal:
I just discussed with @m and there is a related Che server PR: https://github.com/eclipse/che/pull/15282 whoich I assume will **not** be included in CRW 2.0 since it would require a respin of the Che server for something that is not critical.

PR #131, opened 2019-11-22
Fix the JIRA issue CRW-514 (constant Che server rollout) 

davidfestal:
This PR fixes the bug https://issues.jboss.org/browse/CRW-514 that was introduced by PR https://github.com/eclipse/che-operator/pull/126

In order to fix this bug, it mainly orders map keys in alphabetic order when converting a map to string

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -12,6 +12,7 @@
 package util
 
 import (
+	"sort"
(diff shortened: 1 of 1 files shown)


davidfestal:
cc @mshaposhnik @skabashnyuk

PR #134, opened 2019-11-28
add github ci

sparkoo:
Signed-off-by: Michal Vala <mvala@redhat.com>


.github/workflows/go.yml
```diff
@@ -0,0 +1,12 @@
+name: Docker
```
benoitf: I would rename the file to build.yml as with go we don't really know what is being done

FYI it's missing copyright headers

.github/workflows/go.yml
```diff
@@ -0,0 +1,12 @@
...
+on: [push]
+jobs:
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Build the Docker image
+      run: docker build . --file Dockerfile
```
benoitf: AFAIK `--file` is optional no ?

benoitf:
Should the event be on `pull_request` ? because here you assume contribs are always pushed on this repository ?

sparkoo:
> Should the event be on `pull_request` ? because here you assume contribs are always pushed on this repository ?

@benoitf yes, fixed that and your other comments. Thanks for the review

davidfestal:
I'm not against it at all, but there is already a nightly build for the `che-operator` for every commit on master, on the centos CI afair. @ibuziuk do you confirm ?
And the idea was that at some we should extend this to also support PR flow.
Another point: I'm not sure that the general question about what CI server should be used (Centos or Eclipse CI) has been answered for now. @nickboldt @rhopp Any news about that ?

So what is exactly the purpose of using GitHub actions for this now ? Does this PR correspond to any opened and planned issue ?

sparkoo:
@davidfestal it is not part of any planned issue. I'm currently working on few che-operator tasks and I miss CI feedback on pull request so I've added this simple check. When proper (whatever it means) CI will be set, I'm happy to remove this. But currently there is nothing.

davidfestal:
I understand and will approve, though I don't see what it brings in addition to building the docker image locally, which I assume we already do when we develop.

PR #136, opened 2019-11-29
Support workspace namespace configuration

sparkoo:
Adds 2 properties `cheNamespaceDefault` and `cheNamespaceAllowUserDefined` that match che server ones 
cheNamespaceDefault - https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties#L246 
cheNamespaceAllowUserDefined - https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties#L253

issue: https://github.com/eclipse/che/issues/15300


deploy/crds/org_v1_che_cr.yaml
```diff
@@ -50,6 +50,10 @@ spec:
     serverMemoryRequest: ''
     # sets mem limit for server deployment. Defaults to 1Gi
     serverMemoryLimit: ''
+    # sets default namespace where new workspaces will be created
+    cheNamespaceDefault: ''
```
metlos: `workspaceNamespaceDefault`?

deploy/crds/org_v1_che_cr.yaml
```diff
@@ -50,6 +50,10 @@ spec:
...
+    # defines if user is able to specify namespace different from the default
+    cheNamespaceAllowUserDefined: false
```
metlos: `allowUserDefinedWorkspaceNamespaces`?

deploy/crds/org_v1_che_crd.yaml
```diff
@@ -220,7 +220,7 @@ spec:
                 used by the Che installation.
               properties:
                 enable:
-                  description: Enables `metrics` Che server endpoint.
+                  description: Enables `metrics` Che server endpoint. Default to `false`.
```
metlos: Does this change belong in this PR?

davidfestal:
I just have 2 points:

- Just FYI, now that we provide an OpenAPI schema for the `CheCluster` Custom Resource, we might want to upgade the version of the CRD in order to introduce new fields while ensuring backward-compatibilty with already-existing CRDs in a cluster, and proper updates of CSVs.

I already looked into the process of managing multi-version CRDs, and I think we should go for it at some point. But that's not a trivial thing and I think it requires some exploration / tests first.

Not sure this prevents this PR to be merged, but for sure as soon as it, we would have to deal with multi-version CRDs before releasing to the OperatorHub.

- You didn't update the CSVs (in the `olm` sub-directory) to reflect thet changes in the CRD. That's somehow related to the first point. Here we can probably keep it like that for now, but that will have to be fixed before the next release.

deploy/crds/org_v1_che_cr.yaml
sparkoo: yes, better

deploy/crds/org_v1_che_cr.yaml
sparkoo: yes, better

deploy/crds/org_v1_che_crd.yaml
sparkoo: no it doesn't. It's generated code. I've forgot to generate latest changes in previous PR. As it is just description, I've leave it here. I can do it in separate PR if you think it would be better.

sparkoo:
@davidfestal if there is no established way how to update CRD version, it is out of the scope of this issue. If it needs to be handled first, let's do it, but please don't block it longer than necessary.
I've updated olm.

davidfestal:
> @davidfestal
> 
> * if there is no established way how to update CRD version,

There is an established way to do this: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#specify-multiple-versions.

The thing is that we didn't have the need and the opportunity to set it up and test it for Che *until now*.

> it is out of the scope of this issue.

Well, it relates to correctly updating the operator to include the changes you add in this PR.
Generally speaking it seems to me that ensuring successful automatic update of the operator is in the scope of any PR that might impact update.
That said I know that for now it is *not* that easy to test it.   

> If it needs to be handled first, let's do it,

As I said it could be managed in a next PR, but it should at least be planned as a task to achieve *before publishing a next release on OperatorHub*.

> but please don't block it longer than necessary.

I'm not trying to block anything. I just commented without requesting any change. And I'm not the only code owner here.
However I'm trying to point out the fact that, if we merge this PR *and release* without correctly managing the CRD versionning, users will face the following problems:
- Automatic update of Che 7 through OperatorHub will *not* update the CRD (since it already exists in the cluster and it is a cluster-wide resource). So users won't be able to use the new properties.
- If ever the CRD would be updated manually, then the new fields would be available in Openshift 4 console doc and yaml completion, even for previous versions of installed Che that would not have been automatically upgraded (this second case is not critical since this PR only adds new fields).

> * I've updated olm.

Thanks !

davidfestal:
@sparkoo 
After discussing the point with @l0rd , I created a distinct issue for CRD versioning: https://github.com/eclipse/che/issues/15389

sparkoo:
@davidfestal great, thanks.

PR #137, opened 2019-12-03
🚧 <username>-che as default namespace

sparkoo:
Signed-off-by: Michal Vala <mvala@redhat.com>

**What does this PR do?**
 - don't create che-workspace SA
 - check what namespace we're using for workspaces (`workspaceNamespaceDefault`)
   - if it is different than che-operator is deployed to, create all ClusterRoles and ClusterRoleBindings needed to create new namespaces and to create workspaces in them
   - if it is same as che-operator's namespace, create just Role and RoleBinding needed to manage workspaces in same namespace
 - if oauth is enabled, just create `che` sa with no granted permissions

TODO:
 - [ ] `deploy.sh` now works only on `che` namespace, because `ClusterRoleBinding` needs SA namespace. fix that (how?)
 - [x] test and fix permissions for k8s (currently it works for openshift)
 - [x] test with OS Oauth. PR does not grant any permissions for `che` sa
 - [ ] check chectl

issue: https://github.com/eclipse/che/issues/15493


pkg/controller/che/che_controller.go
```diff
@@ -376,49 +376,9 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
-		return reconcile.Result{}, err
-	}
-	viewRole := deploy.NewRole(instance, "view", []string{"pods"}, []string{"list"})
-	if err := r.CreateNewRole(instance, viewRole); err != nil {
-		return reconcile.Result{}, err
-	}
-	// create RoleBindings for created (and existing ClusterRole) roles and service accounts
-	cheRoleBinding := deploy.NewRoleBinding(instance, "che", cheServiceAccount.Name, "edit", "ClusterRole")
-	if err := r.CreateNewRoleBinding(instance, cheRoleBinding); err != nil {
+	if err := createServiceAccounts(instance, r); err != nil {
```
metlos: Could we make the creation of the service account optional based on the enablement of openshift oauth? E.g. oauth disabled => create sa, oauth enabled => don't create the sa.

sparkoo: are you sure that we don't need `che` sa at all when using oauth? it's used to run che pods https://github.com/eclipse/che-operator/blob/master/pkg/deploy/deployment_che.go#L73

metlos: Hmm.. ok, that didn't occur to me :) I'm just worried about the "power" of that service account and would like to minimize the privs if the admin configures the server appropriately - namely, I don't want us to require namespace CRUD only when absolutely necessary.

sparkoo: yes, I will try to just create `che` sa, but don't give it any privileges when oauth.

deploy/cluster_role_che.yaml
```diff
@@ -0,0 +1,142 @@
...
+    resources:
+      - secrets
+  - verbs:
+      - list
+      - create
+      - delete
+    apiGroups:
+      - ''
+    resources:
+      - secrets
```
sleshchenko: Could you elaborate on why we have two `secrets` with different `apiGroups` and `verbs`?

deploy/cluster_role_binding_che.yaml
```diff
@@ -0,0 +1,22 @@
...
+#    which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+#  SPDX-License-Identifier: EPL-2.0
+#
+#  Contributors:
+#    Red Hat, Inc. - initial API and implementation
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: che-operator-che
```
sleshchenko: Is it cluster_role for Che Server? 
Maybe it's better to show it in its name...

Also, is OLM going to create it for Che Server? Is it the right thing to do? Maybe Che Operator should create it only when Che Server needs it?

sparkoo: yes, it is for `che` sa. But operator, to be able to grant these permissions, has to have it on it's own. So I'm creating the ClusterRole now. Do you see anny better approach here?

tolusha:
@sparkoo 
Are you going to work on it?
If not do you mind if @AndrienkoAleksandr finish the task?

sparkoo:
@tolusha @AndrienkoAleksandr feel free to take it, we have currently other priorities.

There is issue with permissions. Current che-operator can handle only all in one namespace strategy. Read https://www.eclipse.org/lists/che-dev/msg03498.html). Otherwise it should be pretty straight-forward. Please let me know if you need any help.

deploy/cluster_role_binding_che.yaml
sleshchenko: > yes, it is for che sa. But operator, to be able to grant these permissions, has to have it on it's own. So I'm creating the ClusterRole now. Do you see anny better approach here?

it's fine for me, I just found that `che-operator-che` RB name tells nothing. Maybe it would be better to name it as `che-operator-cluster-role`.

AndrienkoAleksandr: Ok

AndrienkoAleksandr:
Development and discussions moved to the https://github.com/eclipse/che-operator/pull/156/commits/e5dbf1e1d6127b314ccd3d1d52c62bd15d679adb

PR #140, opened 2019-12-16
Fix a CRD change that is incompatible with OS 3.11

davidfestal:
This PR fixes issue https://github.com/eclipse/che/issues/15396 :
*Invalid CRD openAPIV3Schema on OpenShift 3.11*

It seems that the last CRD generation that was applied on this repository was probably done with an `operator-sdk` > 0.10.0.
It added a `type: object` field  (and description) at the root of the `spec.validation.openAPIV3Schema` field.

This makes the CRD invalid for Kubernetes 1.11 (on which Openshift 3.11 is based), since this version of Kubernetes had the following bug: https://github.com/kubernetes/kubernetes/issues/65293.

This PR removes the added optional fields that made the CRD invalid for OS 3.11

### DIFF
--- a/deploy/crds/org_v1_che_crd.yaml
+++ b/deploy/crds/org_v1_che_crd.yaml
@@ -24,8 +24,6 @@ spec:
     status: {}
   validation:
     openAPIV3Schema:
-      description: The `CheCluster` custom resource allows defining and managing a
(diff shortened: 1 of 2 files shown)


davidfestal:
cc @tolusha

PR #142, opened 2019-12-21
Apply git self-signed SSL certs

vinokurig:
Add new boolean property `gitSelfSignedCert`that applies environment variables that contain SSL certificate and git host written from `che-git-self-signed-cert` config map.
The same logic is applied for the helm deployment: https://github.com/eclipse/che/pull/15218

fixes https://github.com/eclipse/che/issues/15285

Docs PR: https://github.com/eclipse/che-docs/pull/1001

### DIFF
--- a/pkg/deploy/deployment_che.go
+++ b/pkg/deploy/deployment_che.go
@@ -32,7 +32,14 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImage string, cheTag string, cmRe
 		Name:  "CHE_SELF__SIGNED__CERT",
 		Value: "",
 	}
-
+	gitSelfSignedCertEnv := corev1.EnvVar{
+        Name:  "CHE_GIT_SELF__SIGNED__CERT",
+        Value: "",
+    }
+    gitSelfSignedCertHostEnv := corev1.EnvVar{
+        Name:  "CHE_GIT_SELF__SIGNED__CERT__HOST",
+        Value: "",
+    }
 	if cr.Spec.Server.SelfSignedCert {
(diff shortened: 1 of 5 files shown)


tolusha:
@vinokurig 
Should be tested against this doc https://github.com/eclipse/che-docs/pull/1001 ?

vinokurig:
@tolusha 
Yes

pkg/deploy/deployment_che.go
```diff
@@ -47,6 +54,33 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImage string, cheTag string, cmRe
...
+                },
+            },
+        }
+        gitSelfSignedCertHostEnv = corev1.EnvVar{
+            Name: "CHE_GIT_SELF__SIGNED__CERT__HOST",
+            ValueFrom: &corev1.EnvVarSource{
+                ConfigMapKeyRef: &corev1.ConfigMapKeySelector{
+                    Key: "githost",
+                    LocalObjectReference: corev1.LocalObjectReference{
+                        Name: "che-git-self-signed-cert",
```
tolusha: It is the same name as above.
Is it correct?

vinokurig: yes, it uses the same config-map, but another key

vinokurig:
Add new boolean property gitSelfSignedCert that applies environment variables that contain SSL certificate and git host written from che-git-self-signed-cert config map.
The same logic is applied for the helm deployment: eclipse/che#15218

fixes eclipse/che#15285

Docs PR: eclipse/che-docs#1001

PR #144, opened 2020-01-02
Add ability configure host volume path.

AndrienkoAleksandr:
### Referenced issue:
https://github.com/eclipse/che/issues/15065

## Proposed changes

Add ability to configure host volume path and  storage class on the cluster node. Needed to store postgres volume to non `tmp` location in the cluster node. See more concept provided by @gattytto: https://github.com/eclipse/che/issues/15065#issuecomment-550616300


AndrienkoAleksandr:
@gattytto Join to review, please.

gattytto:
confirmed working for the case of postgres-data persistence and also checked che-server pod comes back up after rebooting the minikube node. 

thank you very much, can this behauvoir be replicated to workspace volumes with a "workspacePVCHostVolumePath:" option in the cr.yaml?

AndrienkoAleksandr:
> @gattytto  confirmed working for the case of postgres-data persistence and also checked che-server pod comes back up after rebooting the minikube node.

Thanks a lot.

> "workspacePVCHostVolumePath:" option in the cr.yaml?

@gattytto , Yes, sure. I think it should be second step and it's require modification che-server. I will start work on it soon.

PR #154, opened 2020-01-23
Created minishift e2e tests with minishift.

flacatus:
Updated the `run-okd-tests.sh` script. This script will be a part of PR check.

### DIFF
--- a/e2e/run-okd-tests.sh
+++ b/e2e/run-okd-tests.sh
@@ -1,6 +1,6 @@
-#!/bin/bash -e
+#!/bin/bash
 #
-# Copyright (c) 2012-2018 Red Hat, Inc.
+# Copyright (c) 2012-2020 Red Hat, Inc.
 # This program and the accompanying materials are made
 # available under the terms of the Eclipse Public License 2.0
 # which is available at https://www.eclipse.org/legal/epl-2.0/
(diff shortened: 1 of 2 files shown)


e2e/run-okd-tests.sh
```diff
@@ -10,63 +10,86 @@
...
+  
+  echo "[INFO] Build operator docker image and load in to minishift VM..."
+  cd ${OPERATOR_REPO} && docker build -t che/operator -f Dockerfile . && docker save che/operator > operator.tar
+  eval $(./tmp/minishift docker-env) && docker load -i operator.tar && rm operator.tar
+  
+  echo "[INFO] Run tests..."
+  ./tmp/run-tests
+  
+  echo "[INFO] Deleting minishift VM..."
+  yes | ./tmp/minishift delete && rm -rf ~/.minishift ${OPERATOR_REPO}/tmp
```
tolusha: Since `set -e` when any command fails then minishift won't be deleted.
Is it better to move this to exit trap?

e2e/tests.go
```diff
@@ -95,26 +96,6 @@ func main() {
 		logrus.Info("Installation succeeded")
 	}
 
-	// reconfigure CR to enable TLS support
-	logrus.Info("Patching CR with TLS enabled. This should cause a new Che deployment")
```
tolusha: I guess the code was removed because you covered that with bash script, correct?

e2e/run-okd-tests.sh
flacatus: Yeah. Your right if any commands fails before  `./tmp/minishift delete` minishift will not delete.

e2e/tests.go
flacatus: Yes. But in the future probably need to improve tls tests coverage.

e2e/run-okd-tests.sh
tolusha: It is better to keep `set -e` and delete minishift using bash exit trap

flacatus: Ok. I keep `set -e` and I add bash exit trap.

PR #158, opened 2020-01-28
Create new script to install dependencies like KVM driver for minishift

flacatus:
Created new script to install dependencies like KVM docker machine to start minishift on centos clean machine
Signed-off-by: Flavius Lacatusu <flacatus@redhat.com>


flacatus:
@rhopp I'm agree with you to create for example a new folder for example `.ci` and put there the CI scripts.

PR #160, opened 2020-01-30
Disable readiness probe in case of debug Eclipse Che Server

tolusha:

### Reference issue
https://github.com/eclipse/che/issues/14976

### DIFF
--- a/pkg/deploy/deployment_che.go
+++ b/pkg/deploy/deployment_che.go
@@ -83,6 +83,7 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImage string, cheTag string, cmRe
 
 	memLimit := util.GetValue(cr.Spec.Server.ServerMemoryLimit, DefaultServerMemoryLimit)
 	pullPolicy := corev1.PullPolicy(util.GetValue(string(cr.Spec.Server.CheImagePullPolicy), DefaultPullPolicyFromDockerImage(cheImageAndTag)))
+	cheDebug := util.GetValue(cr.Spec.Server.CheDebug, "false")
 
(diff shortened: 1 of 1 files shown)


pkg/deploy/deployment_che.go
```diff
@@ -83,6 +83,7 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImage string, cheTag string, cmRe
 
 	memLimit := util.GetValue(cr.Spec.Server.ServerMemoryLimit, DefaultServerMemoryLimit)
 	pullPolicy := corev1.PullPolicy(util.GetValue(string(cr.Spec.Server.CheImagePullPolicy), DefaultPullPolicyFromDockerImage(cheImageAndTag)))
+	cheDebug := util.GetValue(cr.Spec.Server.CheDebug, "false")
```
sleshchenko: Please consider moving "false" into some constants file, like we have with others above- - DefaultServerMemoryLimit, DefaultPullPolicyFromDockerImage....

pkg/deploy/deployment_che.go
```diff
@@ -197,6 +179,30 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImage string, cheTag string, cmRe
 			},
 		},
 	}
+
+	// configure readiness probe if debug isn't set
+	cheDebug := util.GetValue(cr.Spec.Server.CheDebug, DefaultCheDebug)
+	if cheDebug == "false" {
```
davidfestal: Maybe :
if cheDebug != true 
?

Since the field is a string, and thus could be empty (which means false) ?

PR #163, opened 2020-02-03
Release 7.8.0

tolusha:
Changes done during the release should be merged to master because they contain new OLM files and new default GO values that will be used as the new basis for future master changes.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -10,31 +10,35 @@
 
 Start a cluster using `cluster-bot` application.
 
-To be able to test update it is needed to created some user before. Login as `kubeadmin`. At the middle of the dashboard click `Update OAuth configuration`. Then `Identity providers` -> `Add` -> `HTPassword` and upload a htpassword file (can be created with HTPassword utility). Login using HTPassword, logout and login again as `kubeadmin`. Go to `kube:admin` menu -> `Copy Login Command` -> `Display Token` and launch to showing command in the terminal. Now it is possible to test update:
+To be able to test update it is needed to created some user before. Login as `kubeadmin`. Click `Update the cluster OAuth configuration` at the middle of the dashboard, then `Identity providers` -> `Add` -> `HTPassword` and upload a htpassword file (can be created with HTPassword utility). Logout and login using HTPassword, then logout and login as `kubeadmin`. Go to `kube:admin` -> `Copy Login Command` -> `Display Token` and launch showing command in the terminal. Now it is possible to test update:
 
 ```bash
-olm/testUpdates.sh openshift stable
+olm/testUpdate.sh openshift stable
 ```
 
-Open Eclipse Che dashboard in the anonymous to validate that the correct version is installed and workspace can be created:
+Open Eclipse Che dashboard in an anonymous tab:
 
 ```bash
-echo http://$(kubectl get ingress -n che | grep ^che | awk -F ' ' '{ print $2 }')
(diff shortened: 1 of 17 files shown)


RELEASE.md
```diff
@@ -10,32 +10,36 @@
...
-olm/testUpdates.sh openshift stable
+olm/testUpdate.sh openshift stable
 ```
 
-Open Eclipse Che dashboard in the anonymous to validate that the correct version is installed and workspace can be created:
+Open Eclipse Che dashboard in an anonymous tab:
 
 ```bash
-echo http://$(kubectl get ingress -n che | grep ^che | awk -F ' ' '{ print $2 }')
+echo http://$(oc get route -n eclipse-che-preview-test | grep ^che | awk -F ' ' '{ print $2 }')
```
davidfestal: The even simpler solution is to retrieve the URL from the status of the `CheCluster` CR when it is in the `Available` state.

make-release.sh
```diff
@@ -37,17 +42,17 @@ check() {
   command -v operator-sdk >/dev/null 2>&1 || { echo -e $RED"operator-sdk is not installed. Aborting."$NC; exit 1; }
 
   local operatorVersion=$(operator-sdk version)
-  [[ ! $operatorVersion =~ .*v0.10.0.* ]] || { echo -e $RED"operator-sdk v0.10.0 is required"$NC; exit 1; }
+  [[ $operatorVersion =~ .*v0.10.0.* ]] || { echo -e $RED"operator-sdk v0.10.0 is required"$NC; exit 1; }
```
davidfestal: why this change ?

pkg/deploy/defaults.go
```diff
@@ -69,10 +69,10 @@ const (
 	// For CRW these should be synced by hand with images stored in RH registries
 	// instead of being synced by script with the content of the upstream `che.properties` file
 	// NB:
-	// The upstream equivent are stored in the generated `extra_images.go` source file.
-	defaultCheWorkspacePluginBrokerInitImage    = "quay.io/crw/pluginbrokerinit-rhel8:2.0-6"
-	defaultCheWorkspacePluginBrokerUnifiedImage = "quay.io/crw/pluginbroker-rhel8:2.0-5"
-	defaultCheServerSecureExposerJwtProxyImage  = "quay.io/crw/jwtproxy-rhel8:2.0-4"
+	// The upstream equivalent are stored in the generated `extra_images.go` source file.
+	defaultCheWorkspacePluginBrokerMetadataImage    = "quay.io/crw/pluginbroker-rhel8:2.1"
```
davidfestal:
```suggestion
	defaultCheWorkspacePluginBrokerMetadataImage    = "quay.io/crw/pluginbroker-metadata-rhel8:2.1"
```

pkg/deploy/defaults.go
```diff
@@ -69,10 +69,10 @@ const (
...
+	defaultCheWorkspacePluginBrokerArtifactsImage = "quay.io/crw/pluginbrokerinit-rhel8:2.1"
```
davidfestal:
```suggestion
	defaultCheWorkspacePluginBrokerArtifactsImage = "quay.io/crw/pluginbroker-artifacts-rhel8:2.1"
```

make-release.sh
tolusha: '||' uses to execute the code when the previous command failed so
`[[ $operatorVersion =~ .*v0.10.0.* ]] || {...` will be run when version isn't `v0.10.0`

RELEASE.md
tolusha: Agree, I will update the script and use it in the next release.

PR #166, opened 2020-02-05
CHE-15493: <username>-che as default namespace

AndrienkoAleksandr:
**Referenced issue**
issue: https://github.com/eclipse/che/issues/15493

**What does this PR do?**
Continue work started https://github.com/eclipse/che-operator/pull/137 :
 - don't create che-workspace SA
 - check what namespace we're using for workspaces (`workspaceNamespaceDefault`)
   - if it is different than che-operator is deployed to, create all ClusterRoles and ClusterRoleBindings needed to create new namespaces and to create workspaces in them
   - if it is same as che-operator's namespace, create just Role and RoleBinding needed to manage workspaces in same namespace
 - if oauth is enabled, just create `che` sa with no granted permissions
 - finalize cluster roles and cluster role bindings created by che-operator for Che server
 - make working Che server property CHE_INFRA_KUBERNETES_CLUSTER__ROLE__NAME

**Related pr(s)**
https://github.com/che-incubator/chectl/pull/469

TODO:
 - [x] test and fix permissions for k8s (currently it works for openshift)
 - [x] test with OS Oauth. PR does not grant any permissions for `che` sa
 - [x] check chectl
 - [x] test installation with olm


deploy/cluster_role.yaml
```diff
@@ -48,3 +48,42 @@ rules:
...
+      - get
+      - create
+  - apiGroups:
+      - authorization.openshift.io
+      - rbac.authorization.k8s.io
+    resources:
+      - rolebindings
+    verbs:
+      - get
+      - update
```
davidfestal: It is really necessary that the operator would be able to *update any rolebinding in any namespace* ?

deploy/cluster_role_binding_che.yaml
```diff
@@ -0,0 +1,22 @@
...
+#  Contributors:
+#    Red Hat, Inc. - initial API and implementation
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: che-operator-che
+subjects:
+  - kind: ServiceAccount
+    name: che-operator
+    namespace: che
```
davidfestal: I don't understand how this can fit the OLM and OperatorHub use-cases, since in these cases we don't control at all the namespace in which the operator will be installed and the `CheCluster` be created to trigger the Che installation.
But even `chectl`-based installation allows configuring the namespace where operator and Che server will be installed.
So I don't understand how a namespaced subject in this cluster role binding makes sense here.

Shouldn't this role binding be configured by the operator itself since knows in which namespace it is installed, and also controls where a given CheCluster will be installed, and with which workspace creation strategy it is configured.

deploy/cluster_role_binding_createns.yaml
```diff
@@ -0,0 +1,22 @@
...
+#  Contributors:
+#    Red Hat, Inc. - initial API and implementation
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: che-operator-create-namespaces
+subjects:
+  - kind: ServiceAccount
+    name: che-operator
+    namespace: che
```
davidfestal: I don't understand how this can fit the OLM and OperatorHub use-cases, since in these cases we don't control at all the namespace in which the operator will be installed and the `CheCluster` be created to trigger the Che installation.
But even `chectl`-based installation allows configuring the namespace where operator and Che server will be installed.
So I don't understand how a namespaced cluster role binding makes sense here.

Shouldn't this role binding be configured by the operator itself since knows in which namespace it is installed, and also controls where a given CheCluster will be installed, and with which workspace creation strategy it is configured.

deploy/cluster_role_che.yaml
```diff
@@ -0,0 +1,159 @@
...
+#  SPDX-License-Identifier: EPL-2.0
+#
+#  Contributors:
+#    Red Hat, Inc. - initial API and implementation
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: che-manage-namespaces
+  labels:
+    app: che
```
davidfestal: Is it a role that is to be applied to the operator it self (=> also integrated into the OLM CSV)  or the the Che server when it is created by the Che operator.
To me *only the permissions granted to the operator itself* should be set by such yaml files.
On the contrary the *permissions granted by the operator to the Che server* (or other K8S objects created by the operator from a `CheCluster)` should be created / bound by the operator when installing a Che server.

deploy/cluster_role_createns.yaml
```diff
@@ -0,0 +1,32 @@
...
+#  SPDX-License-Identifier: EPL-2.0
+#
+#  Contributors:
+#    Red Hat, Inc. - initial API and implementation
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: che-create-namespaces
+  labels:
+    app: che
```
davidfestal: Is it a role that is to be applied to the operator it self (=> also integrated into the OLM CSV)  or the the Che server when it is created by the Che operator.
To me *only the permissions granted to the operator itself* should be set by such yaml files.
On the contrary the *permissions granted by the operator to the Che server* (or other K8S objects created by the operator from a `CheCluster)` should be created / bound by the operator when installing a Che server.

deploy_k8s.sh
```diff
@@ -0,0 +1,41 @@
...
+kubectl apply -f "${BASE_DIR}"/deploy/service_account.yaml -n="${NAMESPACE}"
+kubectl apply -f "${BASE_DIR}"/deploy/role.yaml -n="${NAMESPACE}"
+kubectl apply -f "${BASE_DIR}"/deploy/role_binding.yaml -n="${NAMESPACE}"
+
+kubectl apply -f "${BASE_DIR}"/deploy/cluster_role.yaml
+kubectl apply -f "${BASE_DIR}"/deploy/cluster_role_che.yaml
+kubectl apply -f "${BASE_DIR}"/deploy/cluster_role_createns.yaml
+
+kubectl apply -f "${BASE_DIR}"/deploy/cluster_role_binding.yaml -n="${NAMESPACE}"
+kubectl apply -f "${BASE_DIR}"/deploy/cluster_role_binding_che.yaml -n="${NAMESPACE}"
```
davidfestal: creating, in a configurable namespace, this cluster role binding that refers to the `che-operator` subject in the fixed `che` namespace seems inconsistent to me.

olm/eclipse-che-preview-kubernetes/deploy/olm-catalog/csv-config.yaml
```diff
@@ -1,3 +1,7 @@
-role-paths: [ "generated/roles/role.yaml" ]
+role-paths:
+  - "generated/roles/role.yaml"
+  - "../../deploy/cluster_role.yaml"
+  - "../../deploy/cluster_role_che.yaml"
```
davidfestal: Why do we need to add 3 separate cluster-role.yaml files ?
Why was `cluster_role.yaml` added here while it wasn't necessary previously for the `K8S` use-case ?

pkg/controller/che/che_controller.go
```diff
@@ -1191,6 +1155,179 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+					Resources: []string{"namespaces"},
+					Verbs: []string{"get"},
+				},
+			})
+			if err := r.CreateNewClusterRole(instance, cheManageNameSpacesRole); err != nil {
+				return err
+			}
+
+			// this binding is needed to manage che workspaces out of che namespace
+			// `${namespace}-clusterrole-manage-namespaces` ClusterRole should be created during che-operator deploy
```
davidfestal: What do you mean ? That it is just created 5 lines above ?

pkg/controller/che/che_controller.go
```diff
@@ -1191,6 +1155,179 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+					Resources: []string{"namespaces"},
+					Verbs: []string{"create", "update"},
+				},
+			})
+			if err := r.CreateNewClusterRole(instance, cheCreateNameSpacesRole); err != nil {
+				return err
+			}
+
+			// this binding is needed to create new namespaces for workspaces
+			// `${namespace}-clusterrole-create-namespaces` ClusterRole should be created during che-operator deploy
```
davidfestal: What do you mean ? That it is just created 5 lines above ?

pkg/controller/che/che_controller.go
```diff
@@ -376,50 +380,10 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
-	if err := r.CreateNewRoleBinding(instance, viewRoleBinding); err != nil {
+	if err := createServiceAccounts(instance, r); err != nil {
 		return reconcile.Result{}, err
 	}
 
-	// If the user specified an additional cluster role to use for the Che workspace, create a role binding for it
-	// Use a role binding instead of a cluster role binding to keep the additional access scoped to the workspace's namespace
-	workspaceClusterRole := instance.Spec.Server.CheWorkspaceClusterRole
-	if workspaceClusterRole != "" {
-		customRoleBinding := deploy.NewRoleBinding(instance, "che-workspace-custom", workspaceServiceAccount.Name, workspaceClusterRole, "ClusterRole")
```
davidfestal: Why is this code removed ? Was the logic of this transferred to the Che server itself ... which I assumed according to line [158](https://github.com/eclipse/che-operator/pull/166/files#diff-1ab1e66612216169071c95a0f5767785R158) ?

davidfestal (dismissed):

pkg/controller/che/che_controller.go
AndrienkoAleksandr: Code moved to che-server https://github.com/eclipse/che/blob/b2b6f5f488ea523bac0a9313f0f65a23ac98a16a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesWorkspaceServiceAccount.java#L68

deploy/cluster_role_binding_createns.yaml
AndrienkoAleksandr: OLM generate own clusterroles and roles using package yaml and set up namespace defined in the group yaml: https://github.com/eclipse/che-operator/blob/master/olm/olm.sh#L127

AndrienkoAleksandr:
@davidfestal all cluster roles was join to the on file, like we discussed.

deploy/cluster_role_binding_che.yaml
AndrienkoAleksandr: file removed

deploy/cluster_role_binding_createns.yaml
AndrienkoAleksandr: file removed

deploy/cluster_role_che.yaml
AndrienkoAleksandr: file removed

deploy/cluster_role_createns.yaml
AndrienkoAleksandr: file removed

olm/eclipse-che-preview-kubernetes/deploy/olm-catalog/csv-config.yaml
AndrienkoAleksandr: Fixed, we use only one cluster role binding

pkg/controller/che/che_controller.go
AndrienkoAleksandr: We delegate some part permissions to the che server

pkg/controller/che/che_controller.go
AndrienkoAleksandr: We delegate some part permissions to the che server

deploy/cluster_role.yaml
AndrienkoAleksandr: Che server for replace us update: https://github.com/eclipse/che/blob/master/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesWorkspaceServiceAccount.java#L94

deploy/cluster_role_binding_createns.yaml
davidfestal: Thanks @AndrienkoAleksandr !

deploy/cluster_role_che.yaml
davidfestal: Thanks @AndrienkoAleksandr !

deploy/cluster_role_createns.yaml
davidfestal: Thanks @AndrienkoAleksandr !

olm/eclipse-che-preview-kubernetes/deploy/olm-catalog/csv-config.yaml
davidfestal: Thanks @AndrienkoAleksandr !

pkg/controller/che/che_controller.go
davidfestal: OK

pkg/controller/che/che_controller.go
davidfestal: Thanks for the detail

davidfestal:
> @davidfestal all cluster roles was join to the one file, like we discussed.

Thanks @AndrienkoAleksandr !

tolusha:
[test]

tolusha:
I believe it should be in 7.9.0,
@nickboldt @davidfestal ?

davidfestal:
> I believe it should be in 7.9.0,
> @nickboldt @davidfestal ?

To me the main question is the following: is the ability to configure the workspace namespace, **in the case when OpenShift OAuth is disabled** really a requirement for CRW 2.1 ?

Let me explain: this new feature involves the addition of [highly-requiring cluster-wide permissions to the Che operator](https://github.com/eclipse/che-operator/pull/166/files#diff-991bb55ef23f18e7ffa537df2221d7bcR51).

But on OpenShift, the Eclipse Che `OpenShift Oauth` mode, enabled by default, **already** allows creating workspace K8S objects **under the account of the end-user**, so that the end-user can create new namespaces and manage them **without requiring these additional cluster-wide permissions**.

So requiring those new permissions **systematically** for OpenShift (OCP, OSD) use-cases that would not need it, seems something we should take a bit more time to think about IMO.
In addition, we can easily envision that in some infrastructural contexts, there may be some reluctance to grant such cluster-wide permissions to an operator that still does run in a user namespace.

So I'm not sure that, for 7.9 (and CRW 2.1), we should include this change, unless it's a hard functional requirement.

cc @l0rd @nickboldt @slemeur

sparkoo:
Remaining this thread where I wanted to start this exact discussion https://www.eclipse.org/lists/che-dev/msg03498.html

If we want to have `<username>-che` as default namespace and want to provide the smoothest experience with this setup, we have imho no other option than this PR proposes. In other cases we need to step back of easy "one-click" install or `<username>-che` as default namespace.

davidfestal:
@sparkoo
For non-Openshift Kubernetes contexts, I fully agree with you.

But I was more considering the upcoming CRW 2.1 (and related Che 7.9) release. And with this in mind, what I was trying to say is that, afaik, in the default OpenShift use-case, which is **with Openshift OAuth enabled**, there is no need for this PR, in order to be able to create a namespace for a workspace (it has been working since months), even to be able to use the `<username>-che` namespace for a user by default.

sparkoo:
@davidfestal yes, that's true. I currently can't think of the default universal setup that will not have big drawback for namespace/auth strategies we support. This smells with having more installation paths or configure installation of che-operator.

davidfestal:
> I currently can't think of the default universal setup that will not have big drawback for namespace/auth strategies we support. This smells with having more installation paths or configure installation of che-operator.

Maybe that's something we should quickly take the time to think about *after the 7.9.0* (and CRW 2.1) release, while avoiding adding all these new permissions in the common operator installation *just before the release*.

@AndrienkoAleksandr had already had some ideas about providing several levels of permissions to the che operator (according to the installation context) ,and let the operator logic accommodate these differences to enable or disable some installation options.
But that's not something we can conclude on before the release obviously.

deploy/crds/org_v1_che_cr.yaml
```diff
@@ -59,7 +59,7 @@ spec:
     # sets mem limit for server deployment. Defaults to 1Gi
     serverMemoryLimit: ''
     # sets default namespace where new workspaces will be created
-    workspaceNamespaceDefault: ''
+    workspaceNamespaceDefault: "che-workspace-<username>"
```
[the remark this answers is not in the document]
AndrienkoAleksandr: Yes, I will create corresponding pr.

cmd/manager/signal/signal_handler.go
```diff
@@ -0,0 +1,94 @@
...
+	cheFlavor := os.Getenv("CHE_FLAVOR")
+	if cheFlavor == "" {
+		cheFlavor = "che"
+	}
+	defaultTerminationGracePeriodSeconds := int64(20)
+
+	deployment := &appsv1.Deployment{}
+	namespacedName := types.NamespacedName{Namespace: namespace, Name: cheFlavor + "-operator"}
+	if err := k8sClient.Get(context.TODO(), namespacedName, deployment); err != nil {
+		logrus.Warnf("Unable to find '%s' deployment in namespace '%s', err: %s", cheFlavor+"-operator", namespace, err.Error())
```
tolusha, line 83:
```suggestion
		logrus.Warnf("Unable to find '%s' deployment in namespace '%s', err: %v", cheFlavor+"-operator", namespace, err.Error())
```

pkg/controller/che/che_controller.go
```diff
@@ -549,49 +560,52 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
 		if err != nil {
-			logrus.Error(err)
-		}
-		if !tests {
 			return reconcile.Result{RequeueAfter: time.Second}, err
 		}
-	}
+		// fall back to the "narrower" workspace namespace strategy
+		if len(deniedRules) > 0 {
+			if _, ok := instance.Spec.Server.CustomCheProperties["CHE_INFRA_KUBERNETES_NAMESPACE_DEFAULT"]; ok {
```
tolusha, line 573: Do we really need this check?

pkg/controller/che/workspace_namespace_permission.go
```diff
@@ -0,0 +1,362 @@
...
+//    workspace components.
+// Notice: After permission delegation che-server will create service account "che-workspace" ITSELF with
+//         "exec" and "view" roles for each new workspace.
+func (r *ReconcileChe) delegateWorkspacePermissionsInTheDifferNamespaceThanChe(instance *orgv1.CheCluster, deployContext *deploy.DeployContext) (reconcile.Result, error) {
+	logrus.Info("Configure permissions for Che workspaces. Workspaces will be executed in the differ namespace than Che namespace")
+	tests := r.tests
+
+	cheCreateNamespacesName := fmt.Sprintf(CheWorkspacesNamespaceClusterRoleNameTemplate, instance.Namespace)
+	// Create clusterrole "<workspace-namespace/project-name>-clusterrole-manage-namespaces" to manage namespace/projects for Che workspaces.
+	roleCNsynchronized, err := deploy.SyncClusterRoleToCheCluster(deployContext, cheCreateNamespacesName, getCheCreateNamespacesPolicy())
```
tolusha, line 147:
```suggestion
	provisioned, err := deploy.SyncClusterRoleToCheCluster(deployContext, cheCreateNamespacesName, getCheCreateNamespacesPolicy())
```

pkg/controller/che/workspace_namespace_permission.go
```diff
@@ -0,0 +1,362 @@
...
+			logrus.Error(err)
+		}
+		if !tests {
+			return reconcile.Result{RequeueAfter: time.Second}, err
+		}
+	}
+
+	cheManageNamespacesName := fmt.Sprintf(CheWorkspacesClusterRoleNameTemplate, instance.Namespace)
+	// Create clusterrole "<workspace-namespace/project-name>-clusterrole-workspaces" to create k8s components for Che workspaces.
+	roleMNSynchronized, err := deploy.SyncClusterRoleToCheCluster(deployContext, cheManageNamespacesName, getCheManageNamespacesPolicy())
```
tolusha, line 170:
```suggestion
	provisioned, err = deploy.SyncClusterRoleToCheCluster(deployContext, cheManageNamespacesName, getCheManageNamespacesPolicy())
```

pkg/controller/che/che_controller.go
```diff
@@ -549,49 +560,52 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+				delete(instance.Spec.Server.CustomCheProperties, "CHE_INFRA_KUBERNETES_NAMESPACE_DEFAULT")
```
tolusha, line 574:
```suggestion
instance.Spec.Server.CustomCheProperties = delete(instance.Spec.Server.CustomCheProperties, "CHE_INFRA_KUBERNETES_NAMESPACE_DEFAULT")
```

AndrienkoAleksandr: delete removes elements directly in the map and returns void. Thanks I will remove

```
if _, ok := instance.Spec.Server.CustomCheProperties["CHE_INFRA_KUBERNETES_NAMESPACE_DEFAULT"]; ok {
```

pkg/controller/che/che_controller.go
```diff
@@ -549,49 +560,52 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+			}
 		}
 	}
 
-	cheRoleBinding, err := deploy.SyncRoleBindingToCluster(deployContext, "che", "che", "edit", "ClusterRole")
-	if cheRoleBinding == nil {
-		logrus.Info("Waiting on role binding 'che' to be created")
+	//  Openshift oAuth client handles permissions for workspaces itself.
+	if !util.IsOAuthEnabled(instance) && util.IsWorkspaceInSameNamespaceWithChe(instance) {
+		reconcile, err := r.delegateWorkspacePermissionsInTheSameNamespaceWithChe(deployContext)
```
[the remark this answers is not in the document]
AndrienkoAleksandr: We have this logic. For now it's located in the method reconcileWorkspacePermissionsFinalizer . But Ok, I moved all this stuff about clean up previous permissions set in the same place.

pkg/controller/che/cluster_permissions_finalizer.go
```diff
@@ -0,0 +1,101 @@
...
+		if err := r.client.Update(context.Background(), instance); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *ReconcileChe) removeClusterRole(clusterRoleName string, cheClusterName string) error {
+	logrus.Infof("Custom resource is being deleted. Deleting Cluster role %s.", clusterRoleName)
```
tolusha, line 70:
```suggestion
	logrus.Infof("Deleting Cluster role %s.", clusterRoleName)
```

pkg/controller/che/cluster_permissions_finalizer.go
```diff
@@ -0,0 +1,101 @@
...
+	} else if !errors.IsNotFound(err) {
+		logrus.Errorf("Failed to get %s clusterRole: %s", clusterRoleName, err)
+		return err
+	}
+
+	return nil
+}
+
+func (r *ReconcileChe) removeClusterRoleBinding(clusterRoleBindingName string, cheClusterName string) error {
+	logrus.Infof("Custom resource %s is being deleted. Deleting Cluster rolebinding %s.", cheClusterName, clusterRoleBindingName)
```
tolusha, line 87:
```suggestion
	logrus.Infof("Deleting Cluster rolebinding %s.", clusterRoleBindingName)
```

pkg/controller/che/cluster_permissions_finalizer.go
```diff
@@ -0,0 +1,101 @@
...
+		instance.ObjectMeta.Finalizers = util.DoRemoveString(instance.ObjectMeta.Finalizers, clusterPermissionsFinalizerName)
+		if err := r.client.Update(context.Background(), instance); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *ReconcileChe) removeClusterRole(clusterRoleName string, cheClusterName string) error {
```
tolusha, line 69: Move to `clusterole.go` ?

pkg/controller/che/cluster_permissions_finalizer.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+	} else if !errors.IsNotFound(err) {
+		logrus.Errorf("Failed to get %s clusterRole: %s", clusterRoleName, err)
+		return err
+	}
+
+	return nil
+}
+
+func (r *ReconcileChe) removeClusterRoleBinding(clusterRoleBindingName string, cheClusterName string) error {
```
tolusha, line 86: Move to `clusterrolebinding.go` ?

pkg/controller/che/workspace_namespace_permission.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	if err := deploy.DeleteRoleBinding(EditRoleBindingName, instance.Namespace, cli); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *ReconcileChe) reconcileWorkspacePermissionsFinalizer(instance *orgv1.CheCluster, deployContext *deploy.DeployContext) error {
```
tolusha, line 216: Add some comments.

pkg/controller/che/workspace_namespace_permission.go
```diff
@@ -0,0 +1,362 @@
...
+// - "<workspace-namespace/project-name>-clusterrole-manage-namespaces" - cluster role to mange namespace(for Kubernetes platform)
+//    or project(for Openshift platform) for new workspace.
+// - "<workspace-namespace/project-name>-clusterrole-workspaces" - cluster role to create and manage k8s objects required for
+//    workspace components.
+// Notice: After permission delegation che-server will create service account "che-workspace" ITSELF with
+//         "exec" and "view" roles for each new workspace.
+func (r *ReconcileChe) delegateWorkspacePermissionsInTheDifferNamespaceThanChe(instance *orgv1.CheCluster, deployContext *deploy.DeployContext) (reconcile.Result, error) {
+	tests := r.tests
+
+	cheCreateNamespacesName := fmt.Sprintf(CheWorkspacesNamespaceClusterRoleNameTemplate, instance.Namespace)
```
tolusha, line 143:
```suggestion
	сheWorkspacesNamespaceClusterRoleName := fmt.Sprintf(CheWorkspacesNamespaceClusterRoleNameTemplate, instance.Namespace)
        сheWorkspacesNamespaceClusterRoleBindingName := сheWorkspacesNamespaceClusterRoleName
```

pkg/controller/che/workspace_namespace_permission.go
```diff
@@ -0,0 +1,362 @@
...
+	if !provisioned {
+		logrus.Infof("Waiting on clusterrole '%s' to be created", cheCreateNamespacesName)
+		if err != nil {
+			logrus.Error(err)
+		}
+		if !tests {
+			return reconcile.Result{RequeueAfter: time.Second}, err
+		}
+	}
+	cheCreateNamespacesRolebinding, err := deploy.SyncClusterRoleBindingToCluster(deployContext, cheCreateNamespacesName, CheServiceAccountName, cheCreateNamespacesName)
```
tolusha, line 155:
```suggestion
	сheWorkspacesNamespaceClusterRoleBinding, err := deploy.SyncClusterRoleBindingToCluster(deployContext, cheCreateNamespacesName, CheServiceAccountName, cheCreateNamespacesName)
```

pkg/controller/che/workspace_namespace_permission.go
```diff
@@ -0,0 +1,362 @@
...
+		logrus.Infof("Waiting on clusterrolebinding '%s' to be created", cheCreateNamespacesName)
+		if err != nil {
+			logrus.Error(err)
+		}
+		if !tests {
+			return reconcile.Result{RequeueAfter: time.Second}, err
+		}
+	}
+
+	cheManageNamespacesName := fmt.Sprintf(CheWorkspacesClusterRoleNameTemplate, instance.Namespace)
```
tolusha, line 166:
```suggestion
	сheWorkspacesClusterRoleName := fmt.Sprintf(CheWorkspacesClusterRoleNameTemplate, instance.Namespace)
        сheWorkspacesClusterRoleBindingName := сheWorkspacesClusterRoleName
```

pkg/controller/che/workspace_namespace_permission.go
```diff
@@ -0,0 +1,362 @@
...
+			if err := r.RemoveCheWorkspacesClusterPermissions(instance); err != nil {
+				logrus.Errorf("workspace permissions finalizers was not removed from CR, cause %s", err.Error())
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func getDeleteClusterPermissionsPolicy() []rbac.PolicyRule {
```
tolusha, line 244: Remove pls

pkg/controller/che/workspace_namespace_permission.go
```diff
@@ -0,0 +1,362 @@
...
+		return err
+	}
+	if err := deploy.DeleteRoleBinding(ExecRoleBindingName, instance.Namespace, cli); err != nil {
+		return err
+	}
+
+	if err := deploy.DeleteRole(deploy.ViewRoleName, instance.Namespace, cli); err != nil {
+		return err
+	}
+	if err := deploy.DeleteRoleBinding(ExecRoleBindingName, instance.Namespace, cli); err != nil {
```
tolusha, line 205:
```suggestion
	if err := deploy.DeleteRoleBinding(ViewRoleBindingName, instance.Namespace, cli); err != nil {
```

pkg/controller/che/workspace_namespace_permission.go
```diff
@@ -0,0 +1,362 @@
...
+		if !tests {
+			return reconcile.Result{RequeueAfter: time.Second}, err
+		}
+	}
+	return reconcile.Result{}, nil
+}
+
+// DeleteWorkspacesInSameNamespaceWithChePermissions - removes workspaces in same namespace with Che role and rolebindings.
+func (r *ReconcileChe) DeleteWorkspacesInSameNamespaceWithChePermissions(instance *orgv1.CheCluster, cli client.Client) error {
+	logrus.Info("Delete workspaces in the same namespace with Che permissions.")
```
tolusha, line 193:
```suggestion
```

pkg/controller/che/workspace_namespace_permission.go
```diff
@@ -0,0 +1,362 @@
...
+		}
+		if !tests {
+			// Add workspaces cluster permission finalizer to the CR if deletion timestamp is 0.
+			// Or delete workspaces cluster permission set and finalizer from CR if deletion timestamp is not 0.
+			if err := r.ReconcileCheWorkspacesClusterPermissionsFinalizer(instance); err != nil {
+				logrus.Errorf("unable to add workspace permissions finalizers to the CR, cause %s", err.Error())
+				return err
+			}
+		}
+	} else {
```
tolusha, line 232:
```suggestion
	} else if !util.IsOAuthEnabled(instance) && util.IsWorkspaceInSameNamespaceWithChe(instance) {
```

tolusha (approved):

AndrienkoAleksandr:
CI errors on start workspaces again due image hash for che-theia: https://github.com/eclipse/che/issues/18931

pkg/controller/che/che_controller.go
```diff
@@ -549,49 +561,47 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+			logrus.Warnf("Not enough permissions to start a workspace in dedicated namespace. Denied policies: %v", deniedRules)
+			logrus.Warnf("Fall back to '%s' namespace for workspaces.", instance.Namespace)
+			delete(instance.Spec.Server.CustomCheProperties, "CHE_INFRA_KUBERNETES_NAMESPACE_DEFAULT")
+			instance.Spec.Server.WorkspaceNamespaceDefault = instance.Namespace
+			r.UpdateCheCRSpec(instance, "Default namespace for workspaces", instance.Namespace);
+			if err != nil && !tests {
+				logrus.Error(err)
+				return reconcile.Result{RequeueAfter: time.Second * 1}, err
+			}
+		} else if !util.IsOAuthEnabled(instance) {
```
tolusha, line 586:
```suggestion
		} else {
```

pkg/controller/che/che_controller.go
```diff
@@ -549,49 +561,47 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+			if reconcileResult.Requeue && !tests {
+				return reconcileResult, err
+			}
 		}
 	}
 
-	cheRoleBinding, err := deploy.SyncRoleBindingToCluster(deployContext, "che", "che", "edit", "ClusterRole")
-	if cheRoleBinding == nil {
-		logrus.Info("Waiting on role binding 'che' to be created")
+	if util.IsOAuthEnabled(instance) || util.IsWorkspaceInSameNamespaceWithChe(instance) {
```
tolusha, line 598:
```suggestion
	} else {
```

pkg/controller/che/che_controller.go
```diff
@@ -549,49 +561,47 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+		if len(deniedRules) > 0 {
+			logrus.Warnf("Not enough permissions to start a workspace in dedicated namespace. Denied policies: %v", deniedRules)
+			logrus.Warnf("Fall back to '%s' namespace for workspaces.", instance.Namespace)
+			delete(instance.Spec.Server.CustomCheProperties, "CHE_INFRA_KUBERNETES_NAMESPACE_DEFAULT")
+			instance.Spec.Server.WorkspaceNamespaceDefault = instance.Namespace
+			r.UpdateCheCRSpec(instance, "Default namespace for workspaces", instance.Namespace);
+			if err != nil && !tests {
+				logrus.Error(err)
+				return reconcile.Result{RequeueAfter: time.Second * 1}, err
+			}
```
tolusha, line 585:
```suggestion
			err = r.UpdateCheCRSpec(instance, "Default namespace for workspaces", instance.Namespace);
			return reconcile.Result{RequeueAfter: time.Second * 1}, err
```

pkg/controller/che/workspace_namespace_permission.go
```diff
@@ -0,0 +1,383 @@
...
+
+	return nil
+}
+
+func (r *ReconcileChe) reconcileWorkspacePermissionsFinalizer(instance *orgv1.CheCluster, deployContext *deploy.DeployContext) error {
+	tests := r.tests
+
+	if !util.IsOAuthEnabled(instance) {
+		if util.IsWorkspaceInSameNamespaceWithChe(instance) {
+			if !tests {
```
tolusha, line 222: ?

pkg/controller/che/workspace_namespace_permission.go
```diff
@@ -0,0 +1,383 @@
...
+					return err
+				}
+			}
+		} else {
+			// Delete permission set for configuration "same namespace for Che and workspaces".
+			if err := r.DeleteWorkspacesInSameNamespaceWithChePermissions(instance, deployContext.ClusterAPI.Client); err != nil {
+				logrus.Errorf("unable to delete workspaces in same namespace permission set, cause %s", err.Error())
+				return err
+			}
+			if !tests {
```
tolusha, line 235: ?

tolusha (approved):

pkg/controller/che/che_controller.go
```diff
@@ -549,49 +561,48 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 		}
 	}
 
-	cheWorkspaceSA, err := deploy.SyncServiceAccountToCluster(deployContext, "che-workspace")
-	if cheWorkspaceSA == nil {
-		logrus.Info("Waiting on service account 'che-workspace' to be created")
+	if !util.IsOAuthEnabled(instance) && !util.IsWorkspaceInSameNamespaceWithChe(instance) {
+		clusterRole, err := deploy.GetClusterRole(CheWorkspacesClusterRoleNameTemplate, deployContext.ClusterAPI.Client)
 		if err != nil {
-			logrus.Error(err)
```
tolusha, line 556: Add log pls

sparkoo (commented):
Just taking my approval down as today is it's 1st birthday and a lot of changed since then :)

PR #171, opened 2020-02-14
Propagate custom public certificate to env variable

vinokurig:
Add a property to configure propagation of custom public certificate from config-map to environment variable

https://github.com/eclipse/che/issues/16002

### DIFF
--- a/deploy/crds/org_v1_che_cr.yaml
+++ b/deploy/crds/org_v1_che_cr.yaml
@@ -33,10 +33,10 @@ spec:
     # when set to true the operator will attempt to get a secret in OpenShift router namespace
     # to add it to Java trust store of Che server. Requires cluster-admin privileges for operator service account
     selfSignedCert: false
-    ## If enabled then the certificates from `che-public-certs` config map will be added to Java trust store of the Che server.
-    customPublicCerts: false
-    ## If enabled then the certificate from `che-git-self-signed-cert` config map
-    ## will be propagated to the Che components and provide particular configuration for Git.
+    # Name of the config-map with public certificates to add to Java trust store of the Che server.
+    customPublicCertsConfigMapName: ''
+    # If enabled then the certificate from `che-git-self-signed-cert` config map
+    # will be propagated to the Che components and provide particular configuration for Git.
     gitSelfSignedCert: false
     # TLS mode for Che. Make sure you either have public cert, or set selfSignedCert to true
     tlsSupport: false
--- a/deploy/crds/org_v1_che_cr.yaml
+++ b/deploy/crds/org_v1_che_cr.yaml
@@ -33,8 +33,8 @@ spec:
     # when set to true the operator will attempt to get a secret in OpenShift router namespace
     # to add it to Java trust store of Che server. Requires cluster-admin privileges for operator service account
     selfSignedCert: false
-    ## If enabled then the certificate from `custom-public-cert` config map will be added to Java trust store of Che server.
-    customPublicCert: false
+    ## If enabled then the certificates from `che-public-certs` config map will be added to Java trust store of the Che server.
+    customPublicCerts: false
     ## If enabled then the certificate from `che-git-self-signed-cert` config map
     ## will be propagated to the Che components and provide particular configuration for Git.
     gitSelfSignedCert: false
--- a/deploy/crds/org_v1_che_cr.yaml
+++ b/deploy/crds/org_v1_che_cr.yaml
@@ -33,6 +33,8 @@ spec:
     # when set to true the operator will attempt to get a secret in OpenShift router namespace
     # to add it to Java trust store of Che server. Requires cluster-admin privileges for operator service account
     selfSignedCert: false
+    ## If enabled then the certificate from `custom-public-cert` config map will be added to Java trust store of Che server.
+    customPublicCert: false
     ## If enabled then the certificate from `che-git-self-signed-cert` config map
     ## will be propagated to the Che components and provide particular configuration for Git.
     gitSelfSignedCert: false
--- a/deploy/crds/org_v1_che_cr.yaml
+++ b/deploy/crds/org_v1_che_cr.yaml
@@ -34,7 +34,7 @@ spec:
     # to add it to Java trust store of Che server. Requires cluster-admin privileges for operator service account
     selfSignedCert: false
     # Name of the config-map with public certificates to add to Java trust store of the Che server.
-    customPublicCertsConfigMapName: ''
+    serverTrustStoreConfigMapName: ''
     # If enabled then the certificate from `che-git-self-signed-cert` config map
     # will be propagated to the Che components and provide particular configuration for Git.
     gitSelfSignedCert: false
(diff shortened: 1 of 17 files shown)


vinokurig:
Docs PR: https://github.com/eclipse/che-docs/pull/1078

tolusha:
@vinokurig 
You shouldn't manually update cr/crd file.
If you update `che_types.go` then follow the instruction:

```
// Important: You should regenerate some generated code after modifying this file. At the root o fthe project:
// - Run "operator-sdk generate k8s": this will perform required changes in the "pkg/apis/org/v1/zz_generatedxxx" files
// - Run "operator-sdk generate openapi": this will generate the "deploy/crds/org_v1_checluster_crd.yaml" file
// - In the updated "deploy/crds/org_v1_checluster_crd.yaml": Delete all the `required:` openAPI rules in the CRD OpenApi schema.
// - Rename the new "deploy/crds/org_v1_checluster_crd.yaml" to "deploy/crds/org_v1_che_crd.yaml" to override it.
// IMPORTANT These 2 last steps are important to ensure backward compatibility with already existing `CheCluster` CRs that were created when no schema was provided.
```

vinokurig:
@tolusha 
> You shouldn't manually update cr/crd file.
If you update che_types.go then follow the instruction:

Fixed

vinokurig:
@tolusha
> You shouldn't manually update cr/crd file.
If you update che_types.go then follow the instruction:

Done

deploy/crds/org_v1_che_cr.yaml
```diff
@@ -33,6 +33,8 @@ spec:
     # when set to true the operator will attempt to get a secret in OpenShift router namespace
     # to add it to Java trust store of Che server. Requires cluster-admin privileges for operator service account
     selfSignedCert: false
+    ## If enabled then the certificate from `custom-public-cert` config map will be added to Java trust store of Che server.
+    customPublicCert: false
```
sleshchenko: I know you did in the same way as we have for selfSignedCert and gitSelfSignedCert but it would be much more flexible if user configures configMap/secretMap name instead of specifying `true`/`false` and store their certs in hard-coded `custom-public-cert` config map. WDYT?

pkg/apis/org/v1/che_types.go
```diff
@@ -117,6 +117,13 @@ type CheClusterSpecServer struct {
 	// This is disabled by default.
 	// +optional
 	SelfSignedCert bool `json:"selfSignedCert"`
+	// If enabled, then the certificate from `custom-public-cert`
+	// config map will be added to Java trust store of Che server.
+	// This is usually required when adding the OpenShift OAuth provider
```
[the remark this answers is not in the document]
sleshchenko: Also, it's not clear that che-operator uses `ca.crt` config map key but not any other.

pkg/deploy/deployment_che.go
```diff
@@ -54,6 +58,20 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImage string, cheTag string, cmRe
...
+	if cr.Spec.Server.CustomPublicCert {
+        customPublicCertEnv = corev1.EnvVar{
+            Name: "CHE_CUSTOM_PUBLIC_CERT",
+            ValueFrom: &corev1.EnvVarSource{
+                ConfigMapKeyRef: &corev1.ConfigMapKeySelector{
+                    Key: "ca.crt",
+                    LocalObjectReference: corev1.LocalObjectReference{
+                        Name: "custom-public-cert",
+                    },
+                    Optional: &optionalEnv,
```
sleshchenko: I really don't get motivation why this env var should be optional. It may lead to an unpredictable situation when user configures `true` in CheCluster CR but does not create config map... Che Server will be redeployed but custom-public cert won't be propagated.
In my opinion, it's better to fail a reconciliation loop and tells that config map is missing

pkg/apis/org/v1/che_types.go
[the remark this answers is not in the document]
vinokurig: fixed

pkg/deploy/deployment_che.go
vinokurig: fixed

deploy/crds/org_v1_che_cr.yaml
vinokurig: done

pkg/deploy/deployment_che.go
```diff
@@ -158,6 +179,9 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImageAndTag string, cmRevision st
 									},
 								},
 							},
+							VolumeMounts: []corev1.VolumeMount{
+                                customPublicCertsVolumeMount,
```
sleshchenko: It seems to be wrong formatted

vinokurig: fixed the formatting

tolusha:
@vinokurig 
One more step is required

```bash
cd olm
./update-nightly-olm-files.sh
```

And include the generated files into PR

vinokurig:
@tolusha 
>One more step is required
>
>```
>cd olm
>./update-nightly-olm-files.sh
>```
>And include the generated files into PR

Done

pkg/controller/che/che_controller.go
```diff
@@ -1160,13 +1160,16 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+	desiredCustomPublicCerts := instance.Spec.Server.ServerTrustStoreConfigMapName != ""
 	desiredGitSelfSignedCert := instance.Spec.Server.GitSelfSignedCert
 	effectiveSelfSignedCert := r.GetDeploymentEnvVarSource(effectiveCheDeployment, "CHE_SELF__SIGNED__CERT") != nil
+	effectiveCustomPublicCerts := r.GetDeploymentVolume(effectiveCheDeployment, "che-public-certs").ConfigMap != nil
 	effectiveGitSelfSignedCert := r.GetDeploymentEnvVarSource(effectiveCheDeployment, "CHE_GIT_SELF__SIGNED__CERT") != nil
 	if desiredMemRequest.Cmp(effectiveMemRequest) != 0 ||
 		desiredMemLimit.Cmp(effectiveMemLimit) != 0 ||
 		effectiveImagePullPolicy != desiredImagePullPolicy ||
 		effectiveSelfSignedCert != desiredSelfSignedCert ||
+		effectiveCustomPublicCerts != desiredCustomPublicCerts ||
```
davidfestal: It seems that a change in the content of the config map will not trigger a redeploy of the impacted components (mainly the Che server).
You might want to also support this case by doing the same way as what is done for the `che` config map (with the current config map resource version being added as a metadata in the Che deployment itself).

@tolusha Do we want to support automatic rolling update of the Che server when the custom public certs config map content is changed ?

tolusha: yes. it make sense.

vinokurig: The certificates are represented in files, so we create the certificates config-map with the help of  the `--from-file` key. It is impossible to add/edit a single property using files, user can just edit the whole content of the config map, but I don't think it will be the case.

vinokurig: I have no idea how to check the certificates config-map changed states. In case of the `che config-map` there is a predefined list of values which is compared with the actual values, but the  certificates config map hasn't any predefined values. @tolusha @AndrienkoAleksandr @mmorhun @davidfestal WDYT?

vinokurig:
[ci-test]

flacatus:
[test]

PR #180, opened 2020-02-26
Add script to update cr/crd files

tolusha:

### Reference issue
https://github.com/eclipse/che/issues/14782


davidfestal:
Great @tolusha ! And sorry for not reviewing quickly enough.

PR #182, opened 2020-03-02
Enable to debug operator locally

tolusha:
Enable to debug operator locally

### DIFF
--- a/local-debug.sh
+++ b/local-debug.sh
@@ -6,11 +6,13 @@ if [ $# -ne 1 ]; then
     exit 1
 fi
 
+command -v delv >/dev/null 2>&1 || { echo "operator-sdk is not installed. Aborting."; exit 1; }
+command -v operator-sdk >/dev/null 2>&1 || { echo -e $RED"operator-sdk is not installed. Aborting."$NC; exit 1; }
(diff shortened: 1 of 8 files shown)


local-debug.sh
```diff
@@ -0,0 +1,16 @@
...
+    exit 1
+fi
+
+CHE_NAMESPACE=che
+
+kubectl create namespace $CHE_NAMESPACE
+kubectl apply -f deploy/crds/org_v1_che_crd.yaml
+kubectl apply -f $1 -n che
+
+command -v delv >/dev/null 2>&1 || { echo "operator-sdk is not installed. Aborting."; exit 1; }
```
AndrienkoAleksandr: "operator-sdk is not installed. Aborting." -> "dlv is not installed. Aborting."

pkg/controller/che/create.go
```diff
@@ -294,6 +294,7 @@ func (r *ReconcileChe) CreateIdentityProviderItems(instance *orgv1.CheCluster, r
 			return err
 		}
 	}
+	//
```
AndrienkoAleksandr: We can remove it.

PR #185, opened 2020-03-04
Support single-user mode

tolusha:

### Reference issue
https://github.com/eclipse/che/issues/13420

### How to deploy single user:
CR resource should contains:
```yaml
spec:
  server:
    customCheProperties:
       CHE_MULTIUSER: "false"
```

### DIFF
--- a/pkg/controller/che/che_controller.go
+++ b/pkg/controller/che/che_controller.go
@@ -434,22 +434,22 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 
 	if cheMultiUser == "false" {
 		cheLabels := deploy.GetLabels(instance, cheFlavor)
-		pvc := deploy.NewPvc(instance, deploy.DefaultCheVolumeName, "1Gi", cheLabels)
+		pvc := deploy.NewPvc(instance, deploy.DefaultCheVolumeClaimName, "1Gi", cheLabels)
 		if err := r.CreatePVC(instance, pvc); err != nil {
 			return reconcile.Result{}, err
 		}
-		if !tests {
-			err = r.client.Get(context.TODO(), types.NamespacedName{Name: pvc.Name, Namespace: instance.Namespace}, pvc)
-			if pvc.Status.Phase != "Bound" {
-				k8sclient.GetPVCStatus(pvc, instance.Namespace)
-			}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Name: pvc.Name, Namespace: instance.Namespace}, pvc)
+		if pvc.Status.Phase != "Bound" {
+			k8sclient.GetPVCStatus(pvc, instance.Namespace)
 		}
-		if k8sclient.IsPVCExists(deploy.DefaultPostgresVolumeName, instance.Namespace) {
-			k8sclient.DeletePVC(deploy.DefaultPostgresVolumeName, instance.Namespace)
+		if k8sclient.IsPVCExists(deploy.DefaultPostgresVolumeClaimName, instance.Namespace) {
+			k8sclient.DeletePVC(deploy.DefaultPostgresVolumeClaimName, instance.Namespace)
 		}
 	} else {
-		if k8sclient.IsPVCExists(deploy.DefaultCheVolumeName, instance.Namespace) {
-			k8sclient.DeletePVC(deploy.DefaultCheVolumeName, instance.Namespace)
+		if !tests {
+			if k8sclient.IsPVCExists(deploy.DefaultCheVolumeClaimName, instance.Namespace) {
+				k8sclient.DeletePVC(deploy.DefaultCheVolumeClaimName, instance.Namespace)
+			}
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/deploy/deployment_che.go
```diff
@@ -179,6 +180,24 @@ func NewCheDeployment(cr *orgv1.CheCluster, cheImageAndTag string, cmRevision st
...
 	}
 
+	cheMultiUser := GetCheMultiUser(cr)
+	if cheMultiUser == "false" {
+		cheDeployment.Spec.Template.Spec.Volumes = []corev1.Volume{
+			{
+				Name: DefaultCheVolumeName,
+				VolumeSource: corev1.VolumeSource{
+					PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+						ClaimName: DefaultCheVolumeName,
```
AndrienkoAleksandr: I think it should be called DefaultCheVolumeClaimName, because we don't create volume itself. che-operator creates volumeClaim and infra provide real volume for us.

AndrienkoAleksandr: Ok I see, we are using the same name for volume and claim, so let it be.

pkg/deploy/che_configmap.go
```diff
@@ -198,6 +193,15 @@ func GetConfigMapData(cr *orgv1.CheCluster) (cheEnv map[string]string) {
 		CheMetricsEnabled:                      cheMetrics,
 	}
 
+	if cheMultiUser == "true" {
```
davidfestal: You use the test :
```go
if cheMultiUser == "true" {
```
and in other cases the alternate:
```go
if cheMultiUser == "false" {
```
It seems to me it can lead to errors, since the `cheMultiuser` value is a string, and not a real `boolean`.
What will happen if the value added in the `customCheProperties` is neither `"false"` or `"true"` (think of case errors for example). Afaict this would lead to an inconsistent configuration.

I assume we should always use the test in the same way (compare with the same literal), or maybe the `GetCheMultiUser()` function should force the value to `"true"`, when it is not exactly `"false"` ?

tolusha: Actually `GetCheMultiUser()` returns default value which is `true`  when no value is set in CR

tolusha: But I do agree with you.

pkg/controller/che/che_controller.go
```diff
@@ -574,125 +602,131 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+						logrus.Errorf("An error occurred: %s", err)
+					}
+					if err := r.client.Update(context.TODO(), newKeycloakDeployment); err != nil {
+						logrus.Errorf("Failed to update Keycloak deployment: %s", err)
+					}
 					return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 5}, err
 				}
+				keycloakRealmClientStatus := instance.Status.KeycloakProvisoned
+				if !keycloakRealmClientStatus {
+					if err := r.CreateKyecloakResources(instance, request, keycloakDeployment.Name); err != nil {
```
AndrienkoAleksandr: Old typo: CreateKyecloakResources -> CreateKeycloakResources

sleshchenko:
```yaml
spec:
  server:
    customCheProperties:
       CHE_MULTIUSER: "false"
```
Single/Multi-user mode is not a custom property but kind of fundamental settings because it influences not the only `server` but also `database` and `auth` sections, actually singleUser mode makes that sections ignored, right?
So, please consider moving multi-user property to server level, or even on root spec.

pkg/deploy/che_configmap.go
davidfestal: > Actually GetCheMultiUser() returns default value which is true when no value is set in CR

Sure. I was more thinking of the case where an invalid value (not `false` nor `true`) would be set in the CR.

tolusha:
@sleshchenko 
I disagree with you. 
We already have che property to switch to multi-user mode and introducing a new one is n't good idea.

AndrienkoAleksandr:
> So, please consider moving multi-user property to server level, or even on root spec.

@sleshchenko do we save back compatibility in the custom resource if we will do it?..

AndrienkoAleksandr:
@tolusha Do we need OLM artifacts for this pr?

tolusha:
@AndrienkoAleksandr 
no

tolusha:
@sleshchenko One more point
Single user mode isn't supported

sleshchenko:
> Single user mode isn't supported

@tolusha If we don't want to support SingleUser Che officially by operator - then I'm totally OK with hiding such an ability in customCheProperties.

pkg/deploy/che_configmap.go
davidfestal: Thanks for your fix in commit https://github.com/eclipse/che-operator/pull/185/commits/0e55f83db05c29c8a92e98c5e6993e343ef71bdd

tolusha:
@l0rd 
This pr allows to install Che in single user mode by operator.
The main goal of the PR is to deprecate completely `deploy.sh` script. 
That's why I would like to "hide" this functionality in `customCheProperties` like this and not document at all.

```yaml
spec:
  server:
    customCheProperties:
       CHE_MULTIUSER: "false"
```
wdyt?

PR #186, opened 2020-03-06
Set tlsSupport to true by default

mmorhun:

Sets `tlsSupport` field of Che CR to `true`:
```yaml
spec:
  server:
    tlsSupport: true
```

To set TLS support back to `false` (which is not recommended),  one need to run `chectl` with `--che-operator-cr-patch-yaml=patch.yaml` flag (create the patch file from the snippet above, but change the value to `false`).

### DIFF
--- a/deploy/crds/org_v1_che_cr.yaml
+++ b/deploy/crds/org_v1_che_cr.yaml
@@ -39,7 +39,7 @@ spec:
     # will be propagated to the Che components and provide particular configuration for Git.
     gitSelfSignedCert: false
     # TLS mode for Che. Make sure you either have public cert, or set selfSignedCert to true
-    tlsSupport: false
+    tlsSupport: true
     # protocol+hostname of a proxy server. Automatically added as JAVA_OPTS and https(s)_proxy
     # to Che server and workspaces containers
     proxyURL: ''
--- a/olm/eclipse-che-preview-kubernetes/deploy/olm-catalog/eclipse-che-preview-kubernetes/9.9.9-nightly.1583509666/eclipse-che-preview-kubernetes.crd.yaml
+++ b/olm/eclipse-che-preview-kubernetes/deploy/olm-catalog/eclipse-che-preview-kubernetes/9.9.9-nightly.1583509666/eclipse-che-preview-kubernetes.crd.yaml
@@ -1,0 +1,516 @@
+#
+#  Copyright (c) 2012-2019 Red Hat, Inc.
+#    This program and the accompanying materials are made
+#    available under the terms of the Eclipse Public License 2.0
+#    which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+#  SPDX-License-Identifier: EPL-2.0
+#
+#  Contributors:
+#    Red Hat, Inc. - initial API and implementation
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: checlusters.org.eclipse.che
+spec:
+  group: org.eclipse.che
+  names:
+    kind: CheCluster
+    listKind: CheClusterList
+    plural: checlusters
+    singular: checluster
+  scope: Namespaced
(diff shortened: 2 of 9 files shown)


davidfestal:
You should also update the OLM nightlies (running the script) and then manually change the ALM example annotation in the new nightly CSVs.

tolusha:
> Please add the same changes in OLM files

What does it mean?

AndrienkoAleksandr:
I think @mmorhun should generate new nightly CVS

davidfestal:
> > Please add the same changes in OLM files
> 
> What does it mean?

Sorry I was referring to my [previous comment](https://github.com/eclipse/che-operator/pull/186#issuecomment-595769277):

> You should also update the OLM nightlies (running the script) and then manually change the ALM example annotation in the new nightly CSVs.

erlansan:
> Signed-off-by: Mykola Morhun [mmorhun@redhat.com](mailto:mmorhun@redhat.com)
> 
> Sets `tlsSupport` field of Che CR to `true`:
> 
> ```yaml
> spec:
>   server:
>     tlsSupport: true
> ```
> 
> To set TLS support back to `false` (which is not recommended), one need to run `chectl` with `--che-operator-cr-patch-yaml=patch.yaml` flag (create the patch file from the snippet above, but change the value to `false`).

Hello, thank you for the possible solution! Would it be possible to describe in more detail, where this option is located and how to change it? I'm simply running the commands and getting all of the images from the internet. After reading your solution i'm still yet to understand how to do it. Hope it wont be much trouble, thanks in advance!

AndrienkoAleksandr:
For latest Che `tlsSupport: true` is always true and field is deprecated. See more https://github.com/eclipse/che-operator/blob/master/deploy/crds/org_v1_che_crd.yaml#L445

erlansan:
> @erlansan For latest Che `tlsSupport: true` is always true and field is deprecated. See more https://github.com/eclipse/che-operator/blob/master/deploy/crds/org_v1_che_crd.yaml#L445

I see that. I believe i need to do one of 2 things. Either turn off tsl, or create my own sertificates. I'm currently running into the issue of not being able to start eclipse che on my minishift cluster because i don't have TSL sertificates configured, and minishifts sertificates are not valid(i belive some of developers mentioned that minishift is deprecated at the time)
So therefore i'm not able to start Eclipse Che pod and getting a timeout error. 
I tried figuring out how to create selft signed certificates in Windows, but so far with no success because i'm lacking the expertise to configure it properly(since all of existing commands are tailored for linux based systems)
Therefore i'm asking for some kind of direction, how can i bypass that issue of not being able to run Eclipse Che. If i could turn off tsl support, than i believe i could be able to start Eclipse Che on minishift withoud the pod timing out. I would really appreciate if you could help me out with that.

tolusha:
@eriawan 
Pls follow this guide to install Eclipse Che on Minishift with tls enabled

https://www.eclipse.org/che/docs/che-7/installing-che-in-tls-mode-with-self-signed-certificates/#deploying-che-with-self-signed-tls-certificates-on-openshift-3_installing-che-in-tls-mode-with-self-signed-certificates

AndrienkoAleksandr:
I guess we need write some docs about generation certificates on the Windows...

erlansan:
@tolusha  and @AndrienkoAleksandr  thank you for quick really! I really appreciate that
@tolusha i tried using the https://www.eclipse.org/che/docs/che-7/installing-che-in-tls-mode-with-self-signed-certificates/#generating-self-signed-tls-certificates_installing-che-in-tls-mode-with-self-signed-certificates docu on windows, finding the equivalents of the commands for windows, but no luck, can't get it to work for me. I'm aware of that guide and therefore asked if there is some guidance for windows users, because i lack the knowledge of reqired commands on both OS(windows and linux).

tolusha:
@erlansan 
I see. Despite the flag `tlsSupport` is deprecated we still have working api.
Try set it to false in your CR.

PR #190, opened 2020-03-09
Add minikube installation without vm-driver

flacatus:
In order to install minikube in ci.centos.org needed to install with vm-driver=none flag.


tolusha:
@flacatus 
I am curious, why do we need to test on minikube without vm driver?

flacatus:
> @flacatus
> I am curious, why do we need to test on minikube without vm driver?

Basically, Jobs in ci.centos.org runs under root user and minikube is not able to use VM driver like kvm, virtualbox etc.. The only way to run minikube in ci.centos is with non driver...

PR #191, opened 2020-03-12
Fallback to default Che secret name if none is given in case of turned on tls

mmorhun:

### What does this PR do?
Uses default `che-tls` secret name for securing Che ingresses if TLS is turned on and `tlsSecretName` is not provided in Che CR.

https://github.com/eclipse/che/issues/16280

### DIFF
--- a/pkg/deploy/ingress.go
+++ b/pkg/deploy/ingress.go
@@ -12,6 +12,8 @@
 package deploy
 
 import (
+	"strconv"
+
 	orgv1 "github.com/eclipse/che-operator/pkg/apis/org/v1"
 	"github.com/eclipse/che-operator/pkg/util"
 	"k8s.io/api/extensions/v1beta1"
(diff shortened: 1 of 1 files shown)


pkg/deploy/ingress.go
```diff
@@ -28,11 +30,18 @@ func NewIngress(cr *orgv1.CheCluster, name string, serviceName string, port int)
...
-	tlsSecretName := cr.Spec.K8s.TlsSecretName
+
 	tls := "false"
 	if tlsSupport {
 		tls = "true"
 	}
+	tlsSecretName := cr.Spec.K8s.TlsSecretName
+	// If TLS is turned on but the secret is not set, try to use Che default value as k8s cluster defaults will not work.
+	tlsFlag, _ := strconv.ParseBool(tls)
+	if tlsFlag && tlsSecretName == "" {
```
AndrienkoAleksandr: It can also use util.GetValue(stringFromCR, defaultStringValue)

 Than code could be:
```
tlsSecretName := cr.Spec.K8s.TlsSecretName
tlsFlag, _ := strconv.ParseBool(tls)
if tlsFlag {
        tlsSecretName = util.getValue(tlsSecretName, "che-tls")
}
```

But it optional, up to you.

PR #194, opened 2020-03-13
Set default Che secret name in Che server config map if tlsSecretName is not specified in CR

mmorhun:

### What does this PR do?
Sets default Che TLS secret name in Che config map if Che CR doesn't specify it when TLS is turned on.

### Referenced issues
https://github.com/eclipse/che/issues/16280

### DIFF
--- a/pkg/deploy/che_configmap.go
+++ b/pkg/deploy/che_configmap.go
@@ -128,6 +128,9 @@ func GetConfigMapData(cr *orgv1.CheCluster) (cheEnv map[string]string) {
 
 	ingressDomain := cr.Spec.K8s.IngressDomain
 	tlsSecretName := cr.Spec.K8s.TlsSecretName
+	if tlsSupport && tlsSecretName == "" {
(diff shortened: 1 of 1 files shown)


mmorhun:
[test]

pkg/deploy/che_configmap.go
```diff
@@ -128,6 +128,9 @@ func GetConfigMapData(cr *orgv1.CheCluster) (cheEnv map[string]string) {
 
 	ingressDomain := cr.Spec.K8s.IngressDomain
 	tlsSecretName := cr.Spec.K8s.TlsSecretName
+	if tlsSupport && tlsSecretName == "" {
```
tolusha: `len(tlsSecretName) < 1`

mmorhun: What is the benefit of `len(tlsSecretName) < 1` over `tlsSecretName == ""`?

tolusha: no benefits)

PR #197, opened 2020-03-15
update tests and local yaml to...

nickboldt:
update tests and local yaml to latest

Change-Id: I23b99eece63db52f5aa82fb8566714a8a2e255b7

### DIFF
--- a/deploy/operator-local.yaml
+++ b/deploy/operator-local.yaml
@@ -55,23 +55,22 @@ spec:
             - name: OPERATOR_NAME
               value: "che-operator"
             - name: CHE_VERSION
-              value: "7.8.0"
+              value: 7.9.1
             - name: IMAGE_default_che_server
-              value: "quay.io/eclipse/che-server:7.8.0"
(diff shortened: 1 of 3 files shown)


nickboldt:
Not sure if these files actually need to be updated... but if they do, then the release process ought to be updating them too, as they're terribly out of date. WDYT, @tolusha @davidfestal  ?

(Context is that I've just updated che-operator branch `crw-2.1` so we can use the latest in the `7.9.x` branch in the CRW operator. [CRW-579](https://issues.redhat.com/browse/CRW-579?focusedCommentId=13998978&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13998978)

tolusha:
@nickboldt 
Release does not update the files above.
I will include them in the next release process.

nickboldt:
Thanks, I've opened https://github.com/eclipse/che/issues/16371 for milestone 7.11 so you can track the work.

PR #200, opened 2020-03-17
Correctly propagate proxy settings in the operator and Keycloak  (CRW-709)

davidfestal:
Correctly propagate proxy settings in the operator and Keycloak

This fixes issue https://issues.redhat.com/browse/CRW-709

### DIFF
--- a/pkg/controller/che/k8s_helpers.go
+++ b/pkg/controller/che/k8s_helpers.go
@@ -16,21 +16,25 @@ import (
 	"context"
 	"crypto/tls"
 	"encoding/pem"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
 	orgv1 "github.com/eclipse/che-operator/pkg/apis/org/v1"
 	"github.com/eclipse/che-operator/pkg/deploy"
 	"github.com/eclipse/che-operator/pkg/util"
 	routev1 "github.com/openshift/api/route/v1"
 	"github.com/sirupsen/logrus"
-	"io"
+	"golang.org/x/net/http/httpproxy"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
-	"net/http"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
-	"time"
 )
 
 type k8s struct {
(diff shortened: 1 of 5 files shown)


tolusha:
[test]

pkg/controller/che/k8s_helpers.go
```diff
@@ -269,16 +272,61 @@ func (r *ReconcileChe) GetEndpointTlsCrt(instance *orgv1.CheCluster, url string)
 		// sometimes timing conditions apply, and host isn't available right away
 		if len(testRoute.Spec.Host) < 1 {
 			time.Sleep(time.Duration(1) * time.Second)
-			testRoute := r.GetEffectiveRoute(instance, "test")
```
sleshchenko: This line waited a long time to be fixed :+1:

pkg/controller/che/k8s_helpers.go
```diff
@@ -269,16 +272,61 @@ func (r *ReconcileChe) GetEndpointTlsCrt(instance *orgv1.CheCluster, url string)
...
+			NoProxy:    strings.Replace(instance.Spec.Server.NonProxyHosts, "|", ",", -1),
+		}
+		proxyFunc := config.ProxyFunc()
+		transport.Proxy = func(r *http.Request) (*url.URL, error) {
+			theProxyUrl, err := proxyFunc(r.URL)
+			if err != nil {
+				logrus.Warnf("Error when trying to get the proxy to access TLS endpoint URL: %s - %s", r.URL, err)
+			}
+			logrus.Infof("Using proxy: %s to access TLS endpoint URL: %s", theProxyUrl, r.URL)
+			return theProxyUrl, err
```
sleshchenko: AFAIU It would lead to a situation where test route is not removed and can't be created anymore since it exists 

Consider adding fetching route before try to create it to avoid such situation https://github.com/eclipse/che-operator/pull/66/commits/44212ab8136ae5af62c390ac396e9db728ca0356#diff-5899627ff2ee9c6f7afb34d27f2d3b3eR262

sleshchenko: BTW is the proxy URL returned as certificate by purpose?

davidfestal: Not sure I understand your questions here.This is the return of a *proxy function* that is passed as an argument to the transport object. We're not returning from the main function here.

sleshchenko: Ah, sorry. I need to be more attentive.

pkg/controller/che/k8s_helpers.go
```diff
@@ -269,16 +272,61 @@ func (r *ReconcileChe) GetEndpointTlsCrt(instance *orgv1.CheCluster, url string)
...
 
 	} else {
-		requestURL = url
+		requestURL = endpointUrl
+	}
+
+	//adding the proxy settings to the Transport object
+	transport := &http.Transport{}
+
+	if instance.Spec.Server.ProxyURL != "" {
```
sleshchenko: You're adding +40 lines in already huge func, what you think about extracting some parts in a dedicated func like https://github.com/eclipse/che-operator/commit/dec974f0429215b4731213a6a972000b1e8d1150

^ this is just proposal and does not block this PR

davidfestal: well, the thing is that I already tested this code in a disconnected environment. in both the upstream and CRW cases. And I'd have to restart disconnected env configuration + 2 tests in order to test the refactoring.
Would you mind opening a issue to track:
- the cherry-pick of this to the master branch
- the addition, when cherry-picking to master, of the function extraction you proposed ?

sleshchenko: I'm OK with it since it's an urgent fix which we need to deliver ASAP

PR #204, opened 2020-03-20
Duplicate Dockerfile in the build directory

tolusha:

### Reference issue
https://github.com/eclipse/che/issues/16206

### What does this PR do
For some tasks `operator-sdk` requires `Dockerfile` to be located into `build` directory otherwise it fails with the error:
`FATA[0000] must run command in project root dir: project structure requires build/Dockerfile `

I don't want to remove it from the root of the project since it is used by some build jobs.
Duplication it is enough for operator-sdk

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,0 +1,36 @@
+# Copyright (c) 2018-2020 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


AndrienkoAleksandr:
> Duplication it is enough for operator-sdk

For some operations(debug), for operator sdk is ok even an empty dockerfile...

tolusha:
Agree. It is better make it empty.

PR #208, opened 2020-03-25
Add error status if operator hasn't got permission to count users.

AndrienkoAleksandr:
## Referenced issue:
https://github.com/eclipse/che/issues/15345

### DIFF
--- a/pkg/controller/che/che_controller.go
+++ b/pkg/controller/che/che_controller.go
@@ -224,6 +224,7 @@ const (
 	failedNoOpenshiftUserReason  = "InstallOrUpdateFailed"
 	failedNoOpenshiftUserMessage = "No real user exists in the OpenShift cluster." +
 		" Either disable OpenShift OAuth integration or add at least one user (details in the Help link)"
+	failedUnableToGetOpenshiftUsers = "Unable to get users on the openshift."
 	howToCreateAUserLinkOS4 = "https://docs.openshift.com/container-platform/4.1/authentication/understanding-identity-provider.html#identity-provider-overview_understanding-identity-provider"
 	howToCreateAUserLinkOS3 = "https://docs.openshift.com/container-platform/3.11/install_config/configuring_authentication.html"
 )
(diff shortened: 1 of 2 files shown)


pkg/controller/che/che_controller.go
```diff
@@ -224,6 +224,7 @@ const (
 	failedNoOpenshiftUserReason  = "InstallOrUpdateFailed"
 	failedNoOpenshiftUserMessage = "No real user exists in the OpenShift cluster." +
 		" Either disable OpenShift OAuth integration or add at least one user (details in the Help link)"
+	failedUnableToGetOpenshiftUsers = "Unable to get users on the openshift."
```
tolusha: `openshift` - > `OpenShift cluster`

pkg/controller/che/che_controller.go
```diff
@@ -352,8 +353,13 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 			users := &userv1.UserList{}
 			listOptions := &client.ListOptions{}
 			if err := r.nonCachedClient.List(context.TODO(), listOptions, users); err != nil {
+				getUsersErrMsg := failedUnableToGetOpenshiftUsers + " Cause: " + err.Error()
+				if err := r.SetStatusDetails(instance, request, failedNoOpenshiftUserReason, getUsersErrMsg, ""); err != nil {
+					return reconcile.Result{}, err
+				}
```
tolusha: Besides updating CR with the message it is needed to log the err.

tolusha: otherwise we will have just messages with log level info
```
time="2020-03-25T10:18:03Z" level=info msg="Updating eclipse-che CR with status: Message: Unable to get users on the openshift. Cause: users.user.openshift.io is forbidden: User "system:serviceaccount:che4:che-operator" cannot list users.u
time="2020-03-25T10:18:03Z" level=info msg="Custom resource eclipse-che updated".
```

PR #209, opened 2020-03-25
Add method to determine if operation under k8s resource is permitted

AndrienkoAleksandr:
Can be useful for:
https://github.com/eclipse/che-operator/pull/166

Add method to determine if operation under k8s resource is permitted 

Example how we can used it:
```
func checkGetOpenshiftUsersPermission(cr *orgv1.CheCluster) error {
	getUsersResAttr := &authorizationv1.ResourceAttributes{
			Namespace: cr.Namespace,
			Verb: "get",
			Group: "user.openshift.io",
			Resource: "user",
	}
	ok, err := k8sclient.IsResourceOperationPermitted(getUsersResAttr);
	if  err != nil {
		return fmt.Errorf("fail to check permissions to get Openshift cluster users.")
	}

	if !ok {
		return fmt.Errorf("not enougth permissions to get Openshift cluster users.")
	}

	return nil
}
```

### DIFF
--- a/pkg/util/k8s_helpers.go
+++ b/pkg/util/k8s_helpers.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/remotecommand"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	authorizationv1 "k8s.io/api/authorization/v1"
 )
 
 type k8s struct {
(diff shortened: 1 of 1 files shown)


AndrienkoAleksandr:
@davidfestal we had some discussion https://github.com/eclipse/che-operator/pull/166#issuecomment-585691593 

> @davidfestal some ideas about providing several levels of permissions to the che operator (according to the installation context) ,and let the operator logic accommodate these differences to enable or disable some installation options.
 
What do you think method from current pr will be enough to determine permission and complete https://github.com/eclipse/che-operator/pull/166 ?

AndrienkoAleksandr:
This code was merged in the frame of the another pr https://github.com/eclipse/che-operator/pull/166, so I close it.

PR #213, opened 2020-03-31
CRW-579 don't include dupe entries; if not...

nickboldt:
CRW-579 don't include dupe entries; if not quiet, echo contents of digests-mapping.txt; add --tls-verify=false when doing skopeo inspect

Change-Id: I4599c0929e13db76b19750066cd6aedc3ea8eb3b


davidfestal:
Just a too-late followup: it seems to me it would be less error-prone to add an equality test of the digest, and skip the dupes only if the digests are the same.

We already know that in upstream devfile and plugin registry, there *are* such cases (same name + different tags / digests) 

Maybe a new issue should be created for that ?

nickboldt:
Follow up issue for CRW (and che-operator too):

https://issues.redhat.com/browse/CRW-831

PR #220, opened 2020-04-15
Automate TLS secrets generation for Kubernetes family infrastructures

mmorhun:

### What this PR does

This PR adds checks into Che Operator reconcile loop for required for Che deployment TLS secrets on Kubernetes infrastructure, and if they are absent, it starts job to generate them.

### Which issues this PR fixes or references

https://github.com/eclipse/che/issues/16546

### How to test

- Via `chectl`:
    - Build operator image, push it
    - Disable TLS secrets check and then run:
```sh
checlt server:start --platform=minikube --installer=operator --self-signed-cert --tls --che-operator-image=<operator-image>
```

- Via `local-debug.sh`:
Copy default CR, and change following fields:
```yaml
spec.server.tlsSupport: true
spec.server.selfSignedCert: true
spec.k8s.ingressDomain: <cluster-domain>
```


deploy/operator-local.yaml
```diff
@@ -51,6 +51,8 @@ spec:
               value: quay.io/eclipse/che-plugin-registry:7.11.0
             - name: IMAGE_default_devfile_registry
               value: quay.io/eclipse/che-devfile-registry:7.11.0
+            - name: IMAGE_default_che_tls_secrets_creation_job
+              value: quay.io/eclipse/che-tls-secret-creator:alpine-08c8aee
```
tolusha, line 55: It is needed automatically update this image on every release
Here is I find the latest ub8_minimal image and update all references
https://github.com/eclipse/che-operator/blob/master/make-release.sh#L27-L30

pkg/controller/che/che_controller.go
```diff
@@ -265,6 +265,14 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 		}
 	}
 
+	// Handle Che TLS certificates on Kubernetes like infrastructures
+	if instance.Spec.Server.TlsSupport && instance.Spec.Server.SelfSignedCert && !isOpenShift {
+		shouldReturn, reconsileResult, err := HandleCheTLSSecrets(instance, r)
```
tolusha, line 270: Can we return something like status structure? 
Something like this
https://github.com/eclipse/che-operator/blob/master/pkg/deploy/deployment.go#L42

pkg/controller/che/create.go
```diff
@@ -132,6 +133,29 @@ func (r *ReconcileChe) CreateNewRoleBinding(instance *orgv1.CheCluster, roleBind
 	return nil
 }
 
+// CreateNewJob deploys new instance of given job
+func (r *ReconcileChe) CreateNewJob(instance *orgv1.CheCluster, job *batchv1.Job) error {
```
tolusha, line 137: Create a `syncJobToCluster` function with the same logic as for pvc, ingress, route ...

pkg/controller/che/tls-secrets.go
```diff
@@ -0,0 +1,287 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// shouldReturn specifies whether caller of HandleCheTLSSecrets should return from reconcile loop.
+var shouldReturn = true
+
+// HandleCheTLSSecrets handles TLS secrets required for Che deployment
+func HandleCheTLSSecrets(checluster *orgv1.CheCluster, r *ReconcileChe) (bool, reconcile.Result, error) {
```
tolusha, line 34: Let's return status like this
https://github.com/eclipse/che-operator/blob/master/pkg/deploy/deployment.go#L42

pkg/controller/che/tls-secrets.go
```diff
@@ -0,0 +1,287 @@
...
+			// Che CA certificate doesn't exists (that's expected at this point), do nothing
+		} else {
+			// Remove Che CA secret because Che TLS secret is missing (they should be generated together).
+			if err = r.client.Delete(context.TODO(), cheCASelfSignedCertificateSecret); err != nil {
+				logrus.Errorf("Error deleting Che TLS secret: %v", err)
+				return reconcile.Result{}, err
+			}
+		}
+
+		// Prepare permissions for the certificate generation job
```
tolusha, line 75: Why do we need extra permissions?

deploy/operator-local.yaml
mmorhun: This image is used to generate TLS certificates and create corresponding secrets. It just should work. We don't have to update it on newer releases.

pkg/controller/che/che_controller.go
mmorhun: Theoretically we can, but I am against it. We need to return 3 values from the method and we can do it using embedded in the golang mechanism. Why do we need to introduce new abstraction then?

pkg/controller/che/tls-secrets.go
mmorhun: No, please see my answer above. Moreover, inside internal function we behave the same as in reconcile loop.

pkg/controller/che/tls-secrets.go
mmorhun: We are creating job there and the job should be able to creates secrets. By default it doesn't have permissions to do so and therefore will fail. To handle that we need to create service account, role and role binding to grant permissions to the job to create secrets. Also these permissions cannot be wider than permissions of the unit which gives them.

pkg/controller/che/tls-secrets.go
```diff
@@ -0,0 +1,287 @@
...
+
+		// Che TLS secret doesn't exist, generate a new one
+
+		// Remove Che CA certificate secret if any
+		cheCASelfSignedCertificateSecret := &corev1.Secret{}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: checluster.Namespace, Name: cheTLSSelfSignedCertificateSecretName}, cheCASelfSignedCertificateSecret)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				// Error reading secret info
+				logrus.Errorf("Error getting Che secert \"%s\": %v", cheTLSSecretName, err)
```
AndrienkoAleksandr, line 63: secert sounds not clear for me. Maybe `Self-Signed CA Certificate secret` ?

pkg/controller/che/create.go
mmorhun: I think it is better to keep ability to create an object (in our case job) configuration and then separately flush it to cluster. Because we want to have universal methods, but we never know for sure which aspects of it we may want to change in the future. Changing the sync* method signature is not good as it will affect other consumers. Updating object in runtime is worse or even impossible.
I may move this method into `job.go` and rename it to match `sync*` pattern, but I don't want to loose the flexibility here.

pkg/controller/che/tls-secrets.go
AndrienkoAleksandr: +1 Could we reuse operator ServiceAccount https://github.com/eclipse/che-operator/blob/master/deploy/service_account.yaml?

pkg/controller/che/tls-secrets.go
mmorhun: Ok, will change to `Che TLS secret`. Works for you?

pkg/controller/che/tls-secrets.go
mmorhun: I think it's possible, but the service account you suggest to use has much wider permissions than the job need. I tried to stick to minimal required permissions for the job.

pkg/controller/che/tls-secrets.go
```diff
@@ -0,0 +1,287 @@
...
+		cheTLSJobRole := &rbac.Role{}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: checluster.Namespace, Name: cheTLSJobRoleName}, cheTLSJobRole)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				logrus.Errorf("Error getting Che TLS job role \"%s\": %v", cheTLSJobRoleName, err)
+				return reconcile.Result{}, err
+			}
+
+			// Che TLS job role doesn't exist, creating one
+			cheTLSJobRole = deploy.NewRole(checluster, cheTLSJobRoleName, []string{"secrets"}, []string{"create"})
```
AndrienkoAleksandr, line 104: Could we apply this permission here https://github.com/eclipse/che-operator/blob/master/deploy/role.yaml ? Then we don't need to grammatically create it...

pkg/controller/che/tls-secrets.go
```diff
@@ -0,0 +1,287 @@
...
+				return reconcile.Result{}, err
+			}
+		}
+
+		// Check Che TLS job existence
+		const cheTLSJobName string = "che-tls-job"
+		cheTLSJob := &batchv1.Job{}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: checluster.Namespace, Name: cheTLSJobName}, cheTLSJob)
+		if err != nil && !errors.IsNotFound(err) {
+			logrus.Errorf("Error gitting Che TLS job \"%s\": %v", cheTLSJobRoleBindingName, err)
```
AndrienkoAleksandr, line 133: gitting -> getting

pkg/controller/che/tls-secrets.go
mmorhun: Already done [here](https://github.com/eclipse/che-operator/pull/220/files#diff-4374306bd7fd15cc09428b7ccf9bbec8R23). But I am not sure that child objects can reuse parent permissions, am I wrong?

pkg/controller/che/tls-secrets.go
mmorhun: Thanks, will fix.

pkg/controller/che/tls-secrets.go
AndrienkoAleksandr: Ok, minimal permission per job is a good point.

pkg/controller/che/tls-secrets.go
AndrienkoAleksandr: Ok minimal permissions per job make sense. About question: You could create role binding for this job with secrets only because parent already has this permissions. Parent can't provide more permission to resource object than it has itself.

pkg/controller/che/tls-secrets.go
```diff
@@ -0,0 +1,287 @@
...
+
+		// Give some time for the job to complete.
+		// It is not critical if the job will not be completed till next reconcile loop, such situation is handled above.
+		return reconcile.Result{Requeue: true}, nil
+	}
+
+	// Che TLS certificate exists, check for required data fields
+	if !isCheTLSSecretValid(cheTLSSecret) {
+		// The secret is invalid because required field(s) missing.
+		logrus.Infof("Che TLS secret \"%s\" is invalid. Recrating...", cheTLSSecretName)
```
AndrienkoAleksandr, line 185: Recrating -> Recreating

mmorhun: thanks

pkg/controller/che/tls-secrets.go
mmorhun: Thank you for answering, but my question was a bit different: whether child object has the same permissions as its parent. I don't think so, but not sure.

AndrienkoAleksandr:
Interesting, some day maybe we will have tls generation from the box https://github.com/operator-framework/operator-sdk/blob/master/doc/proposals/tls-utilities.md

mmorhun:
@AndrienkoAleksandr I've already seen that. Unfortunately current functionality they provide has only basic stuff which doesn't suite our needs.

pkg/controller/che/tls-secrets.go
```diff
@@ -0,0 +1,287 @@
...
+		if err = r.client.Delete(context.TODO(), cheTLSSecret); err != nil {
+			logrus.Errorf("Error deleting Che TLS secret \"%s\": %v", cheTLSSecretName, err)
+			return reconcile.Result{}, err
+		}
+		// Regenerate Che TLS certicates and recreate secrets
+		return reconcile.Result{Requeue: true}, nil
+	}
+
+	// Check owner reference
+	if cheTLSSelfSignedCertificateSecret.ObjectMeta.OwnerReferences == nil {
```
AndrienkoAleksandr, line 255: You set up owner references twice. Logic is very similar, maybe make sense to create separated method?

mmorhun: Yes, the two blocks have identical logic except: secret configuration object, secret name and error messages. Also we have to deal with returning values in case of error so in places where we would call such method we need to add new check for an error. Taking into account that this two pieces of code are small and we would still have to handle errors (it is not possible to turn calls into one line) I would leave them as is. But if you have strong opposite vision, I may rework this part, just let me know.

pkg/deploy/job.go
```diff
@@ -0,0 +1,96 @@
...
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// SyncJobToCluster deploys new instance of given job
+func SyncJobToCluster(instance *orgv1.CheCluster, job *batchv1.Job, clusterAPI ClusterAPI) error {
```
tolusha, line 29: pls. have a look at pvc.go for instance.
The implementation should be consistent

pkg/controller/che/tls-secrets.go
```diff
@@ -0,0 +1,308 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// shouldReturn specifies whether caller of HandleCheTLSSecrets should return from reconcile loop.
+var shouldReturn = true
+
+// HandleCheTLSSecrets handles TLS secrets required for Che deployment
+func HandleCheTLSSecrets(checluster *orgv1.CheCluster, r *ReconcileChe) (bool, reconcile.Result, error) {
+	shouldReturn = true
```
tolusha, line 35: Always true?

pkg/controller/che/che_controller.go
AndrienkoAleksandr: +1

pkg/controller/che/tls-secrets.go
mmorhun: No, it will be changed to false if everything is ok

pkg/deploy/job.go
mmorhun: What do you think is not consistent?

pkg/deploy/job.go
```diff
@@ -0,0 +1,96 @@
...
+		logrus.Errorf("An error occurred: %s", err)
+		return err
+	}
+	return nil
+}
+
+// NewJob creates new job configuration by giben parameters.
+func NewJob(checluster *orgv1.CheCluster, name string, namespace string, image string, serviceAccountName string, env map[string]string, backoffLimit int32) *batchv1.Job {
+	labels := GetLabels(checluster, util.GetValue(checluster.Spec.Server.CheFlavor, DefaultCheFlavor))
+	labels["component"] = "che-create-tls-secret-job"
```
tolusha, line 54: I would keep `che component

pkg/deploy/job.go
```diff
@@ -0,0 +1,96 @@
...
+		return nil
+	} else if err != nil {
+		logrus.Errorf("An error occurred: %s", err)
+		return err
+	}
+	return nil
+}
+
+// NewJob creates new job configuration by giben parameters.
+func NewJob(checluster *orgv1.CheCluster, name string, namespace string, image string, serviceAccountName string, env map[string]string, backoffLimit int32) *batchv1.Job {
```
tolusha, line 52: we can get namespace from checluster, not needed to pass it

pkg/deploy/job.go
mmorhun: It is not possible due to conflicts with Che master. All Che related pods have lebel `app=che` including this job.

pkg/deploy/job.go
tolusha: we don't configure ttlSecondsAfterFinished,
backoffLimit, why do we need to configure it? let's use some default value for instance 5.

mmorhun: Just for case when we need to start a job in different than Che namespace.

mmorhun: We indeed don't configure `ttlSecondsAfterFinished` so I may leave it with default value. As for `backoffLimit`, to me it may vary for different jobs, so I left it configurable. I would like to use some default values, but golang doesn't provide such ability.

pkg/deploy/job.go
tolusha: It must be done like this
https://github.com/eclipse/che-operator/blob/master/pkg/deploy/pvc.go

mmorhun: @tolusha which specific part do you think is bad? Return type?

tolusha: `job.go` should contain 3 methods:
`getClusterJob` - to return existed job by name
`getSpecJob` - to return a spec for a new job (+ set owner)
`syncJobToCluster` - to compare spec and cluster jobs, print differences, update the job if necessary and return the status.

It is how we did for others objects.

mmorhun: @tolusha jobs are running after creation. It has a little sense to update one. Are you sure we need to do diff and update?

mmorhun: I'll try to code it like ingress is done

tolusha:
```
time="2020-04-23T08:25:54+03:00" level=info msg="Creating a new object: Job, name: che-tls-job"
time="2020-04-23T08:25:54+03:00" level=info msg="Creating a new object: Job, name: che-tls-job"
time="2020-04-23T08:25:54+03:00" level=error msg="Failed to create che-tls-job Job: jobs.batch \"che-tls-job\" already exists"
time="2020-04-23T08:25:54+03:00" level=error msg="Error creating Che TLS job \"che-tls-job\": jobs.batch \"che-tls-job\" already exists"
time="2020-04-23T08:25:55+03:00" level=info msg="Creating a new object: Job, name: che-tls-job"
time="2020-04-23T08:25:55+03:00" level=error msg="Failed to create che-tls-job Job: jobs.batch \"che-tls-job\" already exists"
time="2020-04-23T08:25:55+03:00" level=error msg="Error creating Che TLS job \"che-tls-job\": jobs.batch \"che-tls-job\" already exists"
time="2020-04-23T08:25:56+03:00" level=info msg="Creating a new object: Job, name: che-tls-job"
time="2020-04-23T08:25:56+03:00" level=error msg="Failed to create che-tls-job Job: jobs.batch \"che-tls-job\" already exists"
time="2020-04-23T08:25:56+03:00" level=error msg="Error creating Che TLS job \"che-tls-job\": jobs.batch \"che-tls-job\" already exists"
time="2020-04-23T08:25:57+03:00" level=info msg="Creating a new object: Job, name: che-tls-job"
time="2020-04-23T08:25:57+03:00" level=error msg="Failed to create che-tls-job Job: jobs.batch \"che-tls-job\" already exists"
time="2020-04-23T08:25:57+03:00" level=error msg="Error creating Che TLS job \"che-tls-job\": jobs.batch \"che-tls-job\" already exists"
```

mmorhun:
@tolusha thanks. It was hard to reproduce the issue. So actually the only way I can see it is to:
1. Stop Che operator
2. Delete Che secrets and all deployments, but leave namespace with the job
3. Start operator again
This looks like artificial for me, but better to fix it. I can do it, when you let me know your thought about proposed approach.

PR #221, opened 2020-04-15
Add Che CR validator

mmorhun:

### What this PR does

This PR adds simple validation of Che CR at the beginning of reconcile loop. If an mistake found in the CR, corresponding message will be printed into Operator logs and it will wait till CR is changed.
For now it checks very basic things, like ingress domain for Kubernetes, but could be easily extended for other cases.

### Which issues this PR references

I ran into issue that Che operator may fall into infinite reconcile loop if some fields of CR are misconfigured. It wasn't fault of reconcile logic as it is just required to have some fields specified.
So this is additional changes in terms of https://github.com/eclipse/che-operator/pull/220

### DIFF
--- a/pkg/controller/che/che_controller.go
+++ b/pkg/controller/che/che_controller.go
@@ -255,6 +255,14 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 		logrus.Errorf("An error occurred when detecting current infra: %s", err)
 	}
 
+	// Check Che CR correctness
+	if isValid, errorMessage := ValidateCheCR(instance, isOpenShift); !isValid {
+		// Che cannot be deployed with current configuration.
+		// Print error message in logs  and wait until the configuration is changed.
+		logrus.Errorf(errorMessage)
+		return reconcile.Result{}, nil
+	}
+
 	if isOpenShift {
 		// delete oAuthClient before CR is deleted
 		doInstallOpenShiftoAuthProvider := instance.Spec.Auth.OpenShiftoAuth
--- a/pkg/controller/che/che_controller.go
+++ b/pkg/controller/che/che_controller.go
@@ -256,10 +256,10 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 	}
 
 	// Check Che CR correctness
-	if isValid, errorMessage := ValidateCheCR(instance, isOpenShift); !isValid {
+	if err := ValidateCheCR(instance, isOpenShift); err != nil {
 		// Che cannot be deployed with current configuration.
-		// Print error message in logs  and wait until the configuration is changed.
-		logrus.Errorf(errorMessage)
+		// Print error message in logs and wait until the configuration is changed.
+		logrus.Error(err)
 		return reconcile.Result{}, nil
 	}
 
--- a/pkg/controller/che/che_cr_validator.go
+++ b/pkg/controller/che/che_cr_validator.go
@@ -1,0 +1,56 @@
+//
+// Copyright (c) 2012-2019 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+package che
+
+import (
+	"fmt"
+
+	orgv1 "github.com/eclipse/che-operator/pkg/apis/org/v1"
+)
+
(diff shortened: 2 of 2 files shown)


pkg/controller/che/che_cr_validator.go
```diff
@@ -0,0 +1,56 @@
...
+// - configurations which miss required field(s) to deploy Che
+// - self-contradictory configurations
+// - configurations with which it is impossible to deploy Che
+func ValidateCheCR(checluster *orgv1.CheCluster, isOpenshift bool) (bool, string) {
+	var isValid bool
+	var errorMessage string
+
+	if !isOpenshift {
+		if checluster.Spec.K8s.IngressDomain == "" {
+			return false, fmt.Sprintf("Required parameter \"Spec.K8s.IngressDomain\" is not set.")
```
tolusha, line 31: paramete -> field
Spec -> spec

mmorhun: ok, will fix

pkg/controller/che/che_cr_validator.go
```diff
@@ -0,0 +1,56 @@
...
+
+	orgv1 "github.com/eclipse/che-operator/pkg/apis/org/v1"
+)
+
+// ValidateCheCR checks Che CR configuration.
+// It should detect:
+// - configurations which miss required field(s) to deploy Che
+// - self-contradictory configurations
+// - configurations with which it is impossible to deploy Che
+func ValidateCheCR(checluster *orgv1.CheCluster, isOpenshift bool) (bool, string) {
```
tolusha, line 25: Return just error and check if error != nil

pkg/controller/che/che_cr_validator.go
```diff
@@ -0,0 +1,56 @@
...
+	isValid, errorMessage = checkTLSConfiguration(checluster, isOpenshift)
+	if !isValid {
+		return isValid, errorMessage
+	}
+
+	return true, ""
+}
+
+func checkTLSConfiguration(checluster *orgv1.CheCluster, isOpenshift bool) (bool, string) {
+	if !checluster.Spec.Server.TlsSupport {
```
tolusha, line 44: Why do we need this check?

mmorhun: Because if TLS is turned off no need to check its configuration

pkg/controller/che/che_cr_validator.go
mmorhun: ok

pkg/controller/che/che_cr_validator.go
tolusha: Then it should be outside function, right?

pkg/controller/che/che_controller.go
```diff
@@ -255,6 +255,14 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 		logrus.Errorf("An error occurred when detecting current infra: %s", err)
 	}
 
+	// Check Che CR correctness
+	if err := ValidateCheCR(instance, isOpenShift); err != nil {
+		// Che cannot be deployed with current configuration.
+		// Print error message in logs  and wait until the configuration is changed.
+		logrus.Error(err)
+		return reconcile.Result{}, nil
```
tolusha, line 263: return reconcile.Result{}, err

tolusha: Have you tested if Eclipse Che is continued deploying when CR changed and validation passed?

mmorhun: No, because in that case it will mean requeue reconcile request and operator will fall into infinite loop.

pkg/controller/che/che_cr_validator.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func checkTLSConfiguration(checluster *orgv1.CheCluster, isOpenshift bool) error {
+	if !checluster.Spec.Server.TlsSupport {
+		return nil
+	}
+
+	if !isOpenshift {
+		// Check TLS secret name is set
+		if checluster.Spec.K8s.TlsSecretName == "" {
```
tolusha, line 46: If we auto-generate secrets  why not to set this field automatically?

mmorhun: It is possible, but should we do it?

mmorhun: Ok, let's automate it.
I will remove this check from this PR and add some handling in TLS automation PR: https://github.com/eclipse/che-operator/pull/220

pkg/controller/che/che_cr_validator.go
mmorhun: The idea was to have a list of checks:
```
checkTLS
checkABCD
...
```
But I may move that if to the top function.

PR #223, opened 2020-04-17
Release 7.9.3

tolusha:
Release 7.9.3


go.mod
```diff
@@ -0,0 +1,72 @@
+module github.com/eclipse/che-operator
```
davidfestal, line 1: Any idea why this change has been introduced ?

tolusha: Automatically generated is VS Code with go plugin and go 1.12 version

PR #229, opened 2020-04-28
Remove leftovers after refactoring that might prevent che deployed

tolusha:


nickboldt:
Can you explain what this does? what refactoring ? what related issue? 

Also if 7.12.0 is already live, should this be merged to 7.12.x and master?

PR #230, opened 2020-04-30
Remove leftovers after refactoring that might prevent che deployed

tolusha:

After refactoring done in this PR [1] I found some leftovers that might prevent che deployed.

[1] https://github.com/eclipse/che-operator/pull/212

Already fixed in master.


tolusha:
Will be fixed in the upcoming 7.13.0 release

PR #235, opened 2020-05-01
fix doc in org_v1_che_cr.yaml: version of...

nickboldt:
fix doc in org_v1_che_cr.yaml: version of registry.redhat.io/rhscl/postgresql-96-rhel7 used is NOT in pkg/deploy/defaults.go, but in the operator metadata CSV now

Change-Id: I8c0630b86fe78d5e7f3869b82a0633bb3b9d2174

### DIFF
--- a/deploy/crds/org_v1_che_cr.yaml
+++ b/deploy/crds/org_v1_che_cr.yaml
@@ -73,7 +73,7 @@ spec:
     chePostgresPassword: ''
     # Postgres database name that Che server uses to connect to. Defaults to dbche
     chePostgresDb: ''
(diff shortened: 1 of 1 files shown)


tolusha:
Changes should be done in che_types.go and then using `olm/update-cr-crd-files.sh` propagated to `crd` file.

nickboldt:
So... you want to document for an end user that we have to update 1 file and run a shell script just to see the default image tag? :facepalm: Seems a little over-engineered :D

tolusha:
oh.  It is about cr not crd file. So, then everything is fine.

PR #238, opened 2020-05-04
Include operator image in the release process

tolusha:

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -24,7 +24,7 @@ spec:
     spec:
       containers:
         - name: che-operator
-          image: quay.io/eclipse/che-operator:nightly
+          image: quay.io/eclipse/che-operator:7.12.1
           ports:
             - containerPort: 60000
               name: metrics
--- a/deploy/crds/org_v1_che_cr.yaml
+++ b/deploy/crds/org_v1_che_cr.yaml
@@ -19,11 +19,11 @@ spec:
     # server image used in Che deployment
     cheImage: ''
     # tag of an image used in Che deployment
-    cheImageTag: 'nightly'
+    cheImageTag: ''
     # image:tag used in Devfile registry deployment
-    devfileRegistryImage: 'quay.io/eclipse/che-devfile-registry:nightly'
+    devfileRegistryImage: ''
     # image:tag used in plugin registry deployment
-    pluginRegistryImage: 'quay.io/eclipse/che-plugin-registry:nightly'
+    pluginRegistryImage: ''
     # defaults to `che`. When set to `codeready`, CodeReady Workspaces is deployed
     # the difference is in images, labels, exec commands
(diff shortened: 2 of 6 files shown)


nickboldt:
I would expect that you would run a script akin to this 

* https://github.com/eclipse/che-plugin-registry/blob/master/make-release.sh#L162-L169 or 
* https://github.com/redhat-developer/codeready-workspaces/blob/master/product/tagRelease.sh#L59-L65

which would generate the content or copy it from another branch, then push it where it needs to be.

nickboldt:
Aside:

> still required to maintain a full list of previous verisons in the OLM catalog

Once you move to OCP 4.5 bundle format (not OCP 4.4 appregistry format) you no longer need to worry about this... old versions will be found from their source branches. Supposedly.

deploy/operator.yaml
```diff
@@ -24,7 +24,7 @@ spec:
     spec:
       containers:
         - name: che-operator
-          image: quay.io/eclipse/che-operator:nightly
+          image: quay.io/eclipse/che-operator:7.12.1
```
nickboldt, line 27: Why do you hardcode 7.12.1 here, but in org_v1_che_cr.yaml  you remove hard coding?

tolusha:
Let me clarify things. 
We have default image references in the `deploy/operator.yaml`.  To override default values user has to specify image refs in the CR file. That's why we have inconsistency in the release branches.
If use tries to use a sample cr (for instance)  from 7.12.1 branch and 7.12 che-operator image then he will have nightly images or che-server etc. That's why I want to remove all references from cr file.

Regarding master branch. Right now we have tags in `deploy/operator.yaml` in master.
https://github.com/eclipse/che-operator/blob/master/deploy/operator.yaml#L46
It means if we deploy che using nightly version of che-operator without cr file then use will have released images :( My second pr will be about removing image references in cr as well and setting nightly tags in operator yaml.

To sum up. 
- in master we will have nightly tags in operator.yaml
- release branch we will have released version in operator.yaml
- no image references in cr file

As usually once we  release a new version:
1. we will merge release branch into a master
2. set nightly tags in operator.yaml

davidfestal:
Thanks @tolusha for this clear explanation. You're right, we're inheriting the current way from the time when default docker images were specified inside the GO source code.
Now that we specify default docker images in the YAML / manifests, I agree it seems better and more consistent to switch to the new way you're proposing.

Afaict the only new thing that we'll have to be careful about now is to switch back to `nightly` tags in `operator.yaml` after merging a release branch back to `master|. Until now it was not necessary right ?

tolusha:
> Afaict the only new thing that we'll have to be careful about now is to switch back to nightly tags in operator.yaml after merging a release branch back to `master|. Until now it was not necessary right ?

right, I am going to automate this by creating corresponding PR after merging release branch to master

PR #240, opened 2020-05-06
Make default Keycloak password generated

mmorhun:

This PR removes default value for Keycloak admin user password which makes operator generate random password for Keycloak on each Che deploy and store it in secret.


flacatus:
@mmorhun The olm tests are failing because last nightly and 7.12.1 versions didn't include include in CSV files field `env`:IMAGE_default_che_tls_secrets_creation_job. Olm dedicated catalog source take all information from CSV files to deploy che

PR #248, opened 2020-05-11
fix: Check identity providers for oauth instead of check amount users.

AndrienkoAleksandr:
Check identity providers for oauth instead of check amount users.

### Referenced issue:
https://github.com/eclipse/che/issues/16836


pkg/controller/che/che_controller.go
```diff
@@ -377,29 +381,33 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
-					helpLink = howToCreateAUserLinkOS3
+					helpLink = howToAddIdentityProviderLinkOS3
 				}
-				logrus.Errorf(failedNoOpenshiftUserMessage)
-				if err := r.SetStatusDetails(instance, request, failedNoOpenshiftUserReason, failedNoOpenshiftUserMessage, helpLink); err != nil {
-					return reconcile.Result{}, err
+				logrus.Warn(warningNoIdentityProvidersMessage, " ", helpLink)
+				instance.Spec.Auth.OpenShiftoAuth = false
+				if err := r.UpdateCheCRSpec(instance, "OpenShiftoAuth", strconv.FormatBool(instance.Spec.Auth.OpenShiftoAuth)); err != nil {
+					if errors.IsConflict(err) {
```
tolusha, line 403: In which cases the conflict is possible?

pkg/controller/che/che_controller.go
```diff
@@ -377,29 +381,33 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+						logrus.Warn("Conflict while getting Openshift oauth info. Updating Che cluster info to resolve conflict...")
```
tolusha, line 404: `getting` is mentioned but we are updating the CR, is it correct?

pkg/controller/che/che_controller.go
```diff
@@ -377,29 +381,33 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+						instance, _ = r.GetCR(request)
+						return reconcile.Result{}, err
+					} else {
+						logrus.Errorf("Failed to get OpenShift oauth info")
```
tolusha, line 408: getting is mentioned but we are updating the CR, is it correct?

pkg/controller/che/che_controller.go
```diff
@@ -377,29 +383,35 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+					getUsersErrMsg := failedUnableToGetOpenshiftUsers + " Cause: " + err.Error()
+					logrus.Errorf(getUsersErrMsg)
+					if err := r.SetStatusDetails(instance, request, failedNoOpenshiftUserReason, getUsersErrMsg, ""); err != nil {
+						return reconcile.Result{}, err
+					}
 					return reconcile.Result{}, err
 				}
-				return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 10}, nil
+				if len(users.Items) < 1 {
+					logrus.Warn(warningNoIdentityProvidersMessage, " ", howToConfigureOAuthLinkOS3)
```
tolusha, line 412: It should be no users messages with corresponding link how to add user

tolusha:
To fix olm tests it is needed to rebase on master

AndrienkoAleksandr:
Tested with crc, openshift, minishift.

tolusha:
Tested on OpenShift with and without oAuth Providers

AndrienkoAleksandr:
Tested with minishift. Works well and stable. Also tested moving from enabled oAuth to disabled and otherwise.

PR #203, opened 2020-03-19
Added support for s390x

aditijadhav38:
### What does this PR do?
Add support for s390x

PR #255, opened 2020-05-14
Added support for ppc64le

junawaneshivani:
### What does this PR do?

I have added support for ppc64le architecture with reference to the [PR](https://github.com/eclipse/che-operator/pull/203) merged for s390x architecture.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,7 @@ ADD . /go/src/github.com/eclipse/che-operator
 RUN case $(uname -m) in \
        x86_64) ARCH="amd64" ;; \
        s390x) ARCH="s390x";; \
+       ppc64le) ARCH="ppc64le";; \
 esac
 
 # do no break RUN lines when building with UBI base images. https://projects.engineering.redhat.com/browse/OSBS-7398 & OSBS-7399


junawaneshivani:
@nickboldt PR now supports all architectures as suggested...

Dockerfile
```diff
@@ -23,6 +23,7 @@ ADD . /go/src/github.com/eclipse/che-operator
 RUN case $(uname -m) in \
        x86_64) ARCH="amd64" ;; \
        s390x) ARCH="s390x";; \
+       ppc64le) ARCH="ppc64le";; \
```
nickboldt, line 26: I might suggest making this less hard-coded and more generic... for s390x, ppc64le, and even future arches...

```
       x86_64) ARCH="amd64" ;; \
       *) ARCH="$1";; \
```

Dockerfile
```diff
@@ -23,6 +23,7 @@ ADD . /go/src/github.com/eclipse/che-operator
...
 esac
 
 # do no break RUN lines when building with UBI base images. https://projects.engineering.redhat.com/browse/OSBS-7398 & OSBS-7399
```
nickboldt, line 29: FYI this comment is no longer correct. Broken lines are allowed in Brew.

nickboldt:
Followup https://github.com/eclipse/che-operator/pull/266

Dockerfile
benoitf: why merging instead of having the suggested proper fix for other arches ?

PR #266, opened 2020-05-15
replace case with simple if-then

nickboldt:
Change-Id:...

replace case with simple if-then

Change-Id: I4663437944ec274b116a332e063b77b86b41ff95

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ WORKDIR /che-operator
 
 RUN export ARCH="$(uname -m)" && if [[ ${ARCH} == "x86_64" ]]; then export ARCH="amd64"; fi && \
     export MOCK_API=true && go test -mod=vendor -v ./... && \
-    OOS=linux GOARCH=${ARCH} CGO_ENABLED=0 go build -mod=vendor -o /tmp/che-operator/che-operator cmd/manager/main.go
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -19,17 +19,11 @@ ENV PATH=/opt/rh/go-toolset-1.12/root/usr/bin:$PATH \
...
-       ppc64le) ARCH="ppc64le";; \
-esac
-
 WORKDIR /che-operator
 
-# do no break RUN lines when building with UBI base images. https://projects.engineering.redhat.com/browse/OSBS-7398 & OSBS-7399
-RUN export MOCK_API=true && go test -mod=vendor -v ./... && OOS=linux GOARCH=$ARCH CGO_ENABLED=0 go build -mod=vendor -o /tmp/che-operator/che-operator cmd/manager/main.go
+RUN export ARCH="$(uname -m)" && if [[ ${ARCH} == "x86_64" ]]; then export ARCH="amd64"; fi && \
+    export MOCK_API=true && go test -mod=vendor -v ./... && \
+    OOS=linux GOARCH=${ARCH} CGO_ENABLED=0 go build -mod=vendor -o /tmp/che-operator/che-operator cmd/manager/main.go
```
amisevsk, line 26: Shouldn't it be `GOOS=linux`?

ghatwala: seems to me as well ..  
`$ go env | grep GOOS

GOOS="linux"`

nickboldt: well, it was `OOS` before my PR, but fair enough, fixed.

PR #268, opened 2020-05-18
Fix git installation. Right now git command execute before git is ins…

flacatus:
This PR fix installation of git. Right now git is installed after using git command in code.
Signed-off-by: flacatus <flacatus@redhat.com>


flacatus:
[test]

PR #272, opened 2020-05-21
Respect openshift cluster wide proxy

tolusha:

### Reference issues
https://issues.redhat.com/browse/CRW-732

### What does this PR do
- reuses openshift cluster wide proxy configuration https://docs.openshift.com/container-platform/4.3/networking/enable-cluster-wide-proxy.html

Tested on OpenShift with proxy configured. Workspace was successfully created and project was cloned.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -50,7 +50,8 @@
         "CONSOLE_LINK_IMAGE": "/dashboard/assets/branding/loader.svg",
         "CHE_IDENTITY_SECRET": "che-identity-secret",
         "CHE_IDENTITY_POSTGRES_SECRET": "che-identity-postgres-secret",
-        "CHE_POSTGRES_SECRET": "che-postgres-secret"
+        "CHE_POSTGRES_SECRET": "che-postgres-secret",
+        "CHE_SERVER_TRUST_STORE_CONFIGMAP_NAME": "ca-certs"
       },
       "cwd": "${workspaceFolder}",
       "args": [
--- a/deploy/cluster_role.yaml
+++ b/deploy/cluster_role.yaml
@@ -30,8 +30,11 @@ rules:
     resources:
       - infrastructures
       - oauths
+      - proxies
     verbs:
       - get
+      - list
+      - watch
   - apiGroups:
       - user.openshift.io
     resources:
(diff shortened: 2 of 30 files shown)


pkg/deploy/configmap_che.go
```diff
@@ -165,29 +123,18 @@ func GetConfigMapData(cr *orgv1.CheCluster) (cheEnv map[string]string) {
...
-	}
-	if len(cr.Spec.Server.ProxyURL) > 1 {
-		proxyJavaOpts, err = util.GenerateProxyJavaOpts(cr.Spec.Server.ProxyURL, cr.Spec.Server.ProxyPort, nonProxyHosts, proxyUser, proxyPassword, proxySecret, cr.Namespace)
-		if err != nil {
-			logrus.Errorf("Failed to generate java proxy options: %v", err)
+	proxyJavaOpts := ""
+	cheWorkspaceNoProxy := proxy.NoProxy
+	if proxy.HttpProxy != "" {
+		if proxy.NoProxy == "" {
+			cheWorkspaceNoProxy = os.Getenv("KUBERNETES_SERVICE_HOST")
```
davidfestal, line 131: I'm not 100% sure if the *external* K8S API URL is by default in the cluster-wide proxy settings, but I don't think so.
And in a number of cases it is required to add it in the non-proxy hosts, to allow the Keycloak server to contact the OpenShift ApiServer through this external URL that is configured in Keycloak for the `openshift-v4` provider in case OpenShift OAuth in enabled.

This external URL is already retrieved by the operator by [this function](https://github.com/eclipse/che-operator/blob/master/pkg/util/util.go#L195) to configure Keycloak with Openshift OAuth. So it could be added by default here since it is usually required in typical proxy-based installations.
Of course maybe a boolean field to disable this automatic addition of the external API host in the noproxy field could certainly be useful.

nickboldt:
If the intent is to include this fix in CRW 2.2, please let me know by the end of the day 06-24, and provide a PR against the 7.14.x branch.

nickboldt:
Slip to 2.3. 

Will this be merged in time for 7.16? 

I could merge it now but pr check failing :D

tolusha:
@nickboldt 
Pls pls pls never merge ;)

nickboldt:
> Pls pls pls never merge ;)

I assume you mean "never merge if tests are failing". If so, no problem!

tolusha:
PR to update CRD doc
https://github.com/eclipse/che-operator/pull/333

PR #273, opened 2020-05-22
Update keycloak db to turn sslRequired to none

tolusha:

### Reference issue
https://issues.redhat.com/browse/CRW-765

### What does this PR do:
set sslRequired=none by updating keycloak db

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -65,6 +65,7 @@ replace (
 )
 
 require (
+	github.com/PuerkitoBio/purell v1.1.1 // indirect
 	github.com/docker/spdystream v0.0.0-00010101000000-000000000000 // indirect
 	github.com/elazarl/goproxy v0.0.0-20200426045556-49ad98f6dac1 // indirect
 	github.com/go-logr/logr v0.0.0-00010101000000-000000000000 // indirect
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -78,6 +79,7 @@ require (
 	github.com/onsi/gomega v1.10.0 // indirect
 	github.com/openshift/api v0.0.0-00010101000000-000000000000
 	github.com/operator-framework/operator-sdk v0.0.0-00010101000000-000000000000
+	github.com/pborman/uuid v1.2.0 // indirect
```
AndrienkoAleksandr, line 82: Do we need CQ?

tolusha: I am not sure. I didn't add this dependency, go.mod was updated automatically somehow.

AndrienkoAleksandr: VSCode extension could update modify go.mod on the fly...

tolusha:
@davidfestal 
tested for both upstream and downstream
sslRequires is set to none

go.mod
nickboldt: if you depend on something new (and therefore so will CRW) then yes, please file a CQ.

tolusha: I've restored go.mod to prev state.

PR #274, opened 2020-05-24
feature: Use image digest in csv

AndrienkoAleksandr:
- [x] https://gist.github.com/AndrienkoAleksandr/69c7c8b2409748ddfb000c912266db79#file-gistfile1-txt-L609

- [x] Added related images to the che-operator env: https://gist.github.com/AndrienkoAleksandr/69c7c8b2409748ddfb000c912266db79#file-gistfile1-txt-L224

### Referenced issue/s:
https://github.com/eclipse/che/issues/14993
https://github.com/eclipse/che/issues/16815

### DIFF
--- a/olm/addDigests.sh
+++ b/olm/addDigests.sh
@@ -10,7 +10,6 @@
 # Contributors:
 #   Red Hat, Inc. - initial API and implementation
 
-SCRIPTS_DIR=$(cd "$(dirname "$0")"; pwd)
 BASE_DIR="$(pwd)"
 QUIET=""
 
@@ -26,7 +25,7 @@ command -v yq >/dev/null 2>&1 || { echo "yq is not installed. Aborting."; exit 1
 
 usage () {
 	echo "Usage:   $0 [-w WORKDIR] -s [SOURCE_PATH] -n [csv name] -v [VERSION] "
-	echo "Example: $0 -w $(pwd) -s eclipse-che-preview-openshift/deploy/olm-catalog/eclipse-che-preview-openshift -n eclipse-che-preview-openshift -v 7.9.0"
+	echo "Example: $0 -w $(pwd) -s eclipse-che-preview-openshift/deploy/olm-catalog/eclipse-che-preview-openshift -r "${CSV_NAME}.*${VERSION}.*.yaml" -n eclipse-che-preview-openshift -v 7.9.0"
 	echo "Example: $0 -w $(pwd) -s controller-manifests -n codeready-workspaces -v 2.1.0"
 }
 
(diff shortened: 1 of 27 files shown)


olm/addDigests.sh
```diff
@@ -50,32 +49,64 @@ rm -Rf ${BASE_DIR}/generated/${CSV_NAME}/
 mkdir -p ${BASE_DIR}/generated/${CSV_NAME}/
 cp -R ${BASE_DIR}/${SRC_DIR}/* ${BASE_DIR}/generated/${CSV_NAME}/
 
-CSV_FILE="$(find ${BASE_DIR}/generated/${CSV_NAME}/*${VERSION}/ -name "${CSV_NAME}.*${VERSION}.clusterserviceversion.yaml" | tail -1)"; # echo "[INFO] CSV = ${CSV_FILE}"
-${SCRIPTS_DIR}/buildDigestMap.sh -w ${BASE_DIR} -c ${CSV_FILE} -v ${VERSION} ${QUIET}
+CSV_FILE="$(find ${BASE_DIR}/generated/${CSV_NAME}/*${VERSION}/ -name "${CSV_NAME}.*${VERSION}.clusterserviceversion.yaml" | tail -1)";
```
nickboldt, line 52: As of OCP 4.4, you don't need to name this file with the version and "clusterserviceversion" so for CRW I've started calling these files simply `codeready-workspaces.csv.yaml`.

I point this out because downstream we need a more permissive `find` that can return more than one file name pattern.

olm/addDigests.sh
```diff
@@ -50,32 +49,64 @@ rm -Rf ${BASE_DIR}/generated/${CSV_NAME}/
...
-    fi
-    RELATED_IMAGES="${RELATED_IMAGES} { name: \"${name}\", image: \"${dest}\", tag: \"${source}\"}"
-    names="${names} ${name} "
-  fi
+  name=$(echo "${source}" | sed -e 's;.*/\([^\/][^\/]*\);\1;' | sed -r 's/[:]+/-/g')
+
+  nameForEnv=$(echo "${name}" | sed -r 's/[:\.\@-]+/_/g')
+  relatedImageEnvName="${RELATED_IMAGE_PREFIX}${nameForEnv}"
+
+CSV_FILE_IN_MEMORY=$( echo "${CSV_FILE_IN_MEMORY}" | \
```
nickboldt, line 82: this new section of code seems much more complicated than what we're doing for CRW.

https://github.com/redhat-developer/codeready-workspaces-operator/tree/master/build/scripts

Why does it need to be some complicated?

nickboldt:
Can you also rename variables so that we're doing the work requested in https://github.com/eclipse/che/issues/16815 ?

olm/addDigests.sh
AndrienkoAleksandr: Because we have an issue 
https://github.com/eclipse/che/issues/14993 with definitions of done:

> 1) Add a relatedImages section in the ClusterServiceVersionSpec. Some images should be flagged as required (che-server, keycloak, postres, devfile and plugin registry, che-theia, brokers etc...) others not (plugins sidecars and stacks)
> 2) Update the operator deployment yaml and the operator itself to leverage the olm.relatedImage annotations

To satisfy "definition of done" I followed spec: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/contributors/design-proposals/related-images.md

AndrienkoAleksandr:
> Can you also rename variables so that we're doing the work requested in eclipse/che#16815 

Most envs has prefix RELATED_IMAGES_ 
 https://gist.github.com/AndrienkoAleksandr/02dd792d63878ae5516ec9069950839e#file-gistfile1-txt-L492   ,
But, yes, for default images, I didn't change name https://gist.github.com/AndrienkoAleksandr/02dd792d63878ae5516ec9069950839e#file-gistfile1-txt-L436 ...

@tolusha What Do you think could we rename default images env variable, or we can get issue with back compatibility?

olm/addDigests.sh
nickboldt: that spec is listed as "Status: Pending" and "Version: Alpha". So... I would make sure it's final before implementing it.

Meanwhile we already have relatedImages in use in CRW operator.

https://github.com/redhat-developer/codeready-workspaces-operator/blob/master/controller-manifests/v2.1.1/codeready-workspaces.csv.yaml#L421

That only uses two fields: `name` and `image`. I don't know if the other stuff like `annotation` or `required` are implemented in OCP yet.

tolusha:
@AndrienkoAleksandr 
We shouldn't fail into any issue with that.

tolusha:
@AndrienkoAleksandr 
When you rebase on master update this line pls
https://github.com/eclipse/che-operator/blob/master/olm/update-nightly-olm-files.sh#L74

olm/addDigests.sh
AndrienkoAleksandr: I changed script to provide CSV file regexp.

olm/addDigests.sh
AndrienkoAleksandr: Yes, you are right. I reverted some changes related to the unimplemented spec... Related comment https://github.com/eclipse/che/issues/14993#issuecomment-637540933

AndrienkoAleksandr:
I update pull request description.

tolusha:
```
  relatedImages:
    - name: che-operator-digest
      image: docker.io/aandrienko/che-operator@sha256:7f4a80ab817caa97ca808d8934e542fceedeeed9468f13babb6a3a1fb066ac30
      # tag: docker.io/aandrienko/che-operator:digest
    - name: che-server-7.14.1
```

Maybe it isn't relevant but tag is `digest`
`  # tag: docker.io/aandrienko/che-operator:digest`

AndrienkoAleksandr:
> Maybe it isn't relevant but tag is digest

Yes, I really used such name for my test tag.

AndrienkoAleksandr:
I will generate nightly scripts later, otherwise this pr is always in merge conflict.

tolusha:
@nickboldt 
Could you clarify why we have to wait for the last 7.15.x release?

nickboldt:
I was saying that now that it's updated to the last 7.15.x release, it's safe to merge. the new CSVs are no longer causing a conflict. Merge when ready, I already +1'd above.

AndrienkoAleksandr:
CI is green

nickboldt:
This change seems to have caused https://github.com/eclipse/che/issues/17480 but I'm not sure if the formatting change is a problem or just results in illegible yaml.

PR #282, opened 2020-05-28
Increase time to wait for OLM and Eclipse installation

flacatus:
Sometimes in centos ci we have some network issue and is necessary to increase the time to wait for eclipse che images to download or OLM as well.


flacatus:
[test]

PR #283, opened 2020-05-28
Set sslRequired=NONE by updating DB

tolusha:
### Reference issue
https://issues.redhat.com/browse/CRW-765

### What does this PR do
Set `ssl_required` to `NONE`  for master realm using db query instead of admin scripts.
Restarts keycloak pod to reload keycloak


pkg/deploy/exec_commands.go
```diff
@@ -67,7 +67,7 @@ func GetKeycloakProvisionCommand(cr *orgv1.CheCluster, cheHost string) (command
 		"$realmDisplayName", realmDisplayName,
 		"$keycloakClientId", keycloakClientId,
 		"$keycloakTheme", keycloakTheme,
-		"$cheHost", cheHost,
```
davidfestal, line 70: Could it be that it was passed as an argument because at the point in time, the status was still not updated with the last changes (http to https for example) ?
Maybe a dumb question, but still asking ;-)

tolusha: I don't think so.
Previously it was
```
	cheHost := instance.Spec.Server.CheHost
	keycloakProvisionCommand := deploy.GetKeycloakProvisionCommand(instance, cheHost)
```

I think it is just some leftovers.

PR #285, opened 2020-05-29
Always propogate router certificate into trust store

mmorhun:

### What this PR does
Deprecates `selfSignedCert` option. CR value is present but actually has no effect.
In case of Openshift router certificate will always be propagated to Che trust store.
In case of Kubernetes certificate is generated and propagated into Che trust store unless use provides single TLS secret (no self signed one): then we consider certificate commonly trusted.

### Related issues:
https://github.com/eclipse/che/issues/16764

### Tests
Tested on the following infrastructures with self-signed certificates:
 - Minikube
 - CRC
 - Openshift (with oAuth)


pkg/controller/che/tls-secrets.go
```diff
@@ -161,55 +162,41 @@ func HandleCheTLSSecrets(checluster *orgv1.CheCluster, clusterAPI deploy.Cluster
...
-		// Also delete Che TLS as the certificates should be created together
-		// Here it is not mandatory to check Che TLS secret existence as it is handled above
-		if err = clusterAPI.Client.Delete(context.TODO(), cheTLSSecret); err != nil {
-			logrus.Errorf("Error deleting Che TLS secret \"%s\": %v", cheTLSSecretName, err)
-			return reconcile.Result{RequeueAfter: time.Second}, err
+		// This means that commonly trusted certificate is used.
+	} else {
+		// Che CA self-signed certificate secret exists, check for required data fields
+		if !isCheCASecretValid(cheTLSSelfSignedCertificateSecret) {
+			logrus.Infof("Che self-signed certificate secret \"%s\" is invalid. Recrating...", CheTLSSelfSignedCertificateSecretName)
```
davidfestal, line 169:
```suggestion
			logrus.Infof("Che self-signed certificate secret \"%s\" is invalid. Recreating...", CheTLSSelfSignedCertificateSecretName)
```

go.mod
```diff
@@ -65,6 +65,7 @@ replace (
 )
 
 require (
+	github.com/PuerkitoBio/purell v1.1.1 // indirect
```
AndrienkoAleksandr, line 68: I think these unused dependencies spawn older go. We have similar discussion https://github.com/eclipse/che-machine-exec/pull/83#issuecomment-624731577 I think you should use a bit newer go to avoid this.

AndrienkoAleksandr: With newer go: `go mod tidy` and you will see that newer go will remove these deps from go go.mod

mmorhun:
@davidfestal do you remember what kind of problems causes an attempt to add commonly trusted certificate into trust store?

go.mod
mmorhun: This changes is come when I build operator image in docker.

mmorhun:
Rebased

README.md
```diff
@@ -47,14 +47,11 @@ Make sure your current user has cluster-admin privileges.
...
-When using self-signed certificates make sure you set `server.selfSignedCert` to true
-or create a secret called `self-signed-certificate` in a target namespace with ca.crt holding your OpenShift router crt body.
-When `server.selfSignedCert` the operator will create a test TLS route, GET it, extract certificate chain, convert to a secret `self-signed-certificate`,
-and Che/CRW server will automatically add it to Java trust store.
+TLS enabled and handled automatically by adding router certificate into trust store.
 
 #### K8S
 
-When enabling TLS, make sure you create a secret with crt and key, and let the Operator know about it in `k8s.tlsSecretName`
+TLS enabled by default. If secrets is missing, self-signed certificate will be generated. If one need to provide own certificate, just create a secret with `tls.crt` and `tls.key` keys in `data`(and `self-signed-certificate` secret in case of using one), and let the Operator know about the certificate for Che server in `k8s.tlsSecretName` field of Che custom resource (`che-tls` by default).
```
tolusha, line 54: `If one need` -> `If it is needed`

mmorhun:
Rebased

mmorhun:
Closing this in favour of https://github.com/eclipse/che-operator/pull/301 because propagation of router certificate all the time leads to malfunction of some components in case of commonly trusted certificate.

PR #301, opened 2020-06-04
Implement autodetetion for self-signed certificate flag

mmorhun:

### What this PR does
Deprecates `selfSignedCert` option. CR value is present, but actually has no effect.

In case of Kubernetes, `self-signed-certificate` secret is checked for presence. If it exists, then it will be propagated to Che components. Also, if no secrets are precreated, they will be generated and treated as self-signed.

In case of Openshift, router certificate is examined for self-signed certificate presence. If so, then `self-signed-certificate` secret will be created and propagated to Che trust store. Otherwise, it is considered that commonly trusted certificate is used.

### Related issues:
https://github.com/eclipse/che/issues/16764

### Tests
Tested on:
 - Openshift wih self-signed-certificate
 - Openshift wiht commonly trusted certificate
 - Minikube
 - CRC
 - Minishift (requires manual secret creation as Che docs says).


pkg/apis/org/v1/che_types.go
```diff
@@ -130,9 +127,9 @@ type CheClusterSpecServer struct {
 	// configuration for Git.
 	// +optional
 	GitSelfSignedCert bool `json:"gitSelfSignedCert"`
-	// Instructs the operator to deploy Che in TLS mode, ie with TLS routes or ingresses.
-	// This is disabled by default.
-	// WARNING: Enabling TLS might require enabling the `selfSignedCert` field also in some cases.
+	// Instructs the operator to deploy Che in TLS mode.
+	// This is enabled by default.
+	// Deprecated. Disabling TLS may cause malfunction of some Che components (Che Theia web-views).
```
tolusha, line 132: Don't put any reference to che since it is used for crw as well

README.md
```diff
@@ -83,16 +83,18 @@ Make sure your current user has cluster-admin privileges.
 
 ### TLS
 
+TLS is enabled by default.
+Turning it off is not recommended as it will cause malfunction of some components (for example Che Theia web-views).
```
tolusha, line 87: Remove reference to any specific components names

pkg/deploy/tls.go
```diff
@@ -0,0 +1,448 @@
...
+				return reconcile.Result{RequeueAfter: time.Second}, err
+			}
+			// Regenerate Che TLS certicates and recreate secrets
+			return reconcile.Result{RequeueAfter: time.Second}, nil
+		}
+
+		// Check owner reference
+		if cheTLSSelfSignedCertificateSecret.ObjectMeta.OwnerReferences == nil {
+			// Set owner Che cluster as Che TLS secret owner
+			if err := controllerutil.SetControllerReference(checluster, cheTLSSelfSignedCertificateSecret, clusterAPI.Scheme); err != nil {
```
tolusha, line 387: Is it better to move this code to `getSpecSecret`
We set controllerRefence for all components at getSpec func

pkg/deploy/tls.go
```diff
@@ -0,0 +1,448 @@
...
+			return reconcile.Result{RequeueAfter: time.Second}, err
+		}
+		// Recreate the secret
+		return reconcile.Result{RequeueAfter: time.Second}, err
+	}
+
+	// Check owner reference
+	if cheTLSSecret.ObjectMeta.OwnerReferences == nil {
+		// Set owner Che cluster as Che TLS secret owner
+		if err := controllerutil.SetControllerReference(checluster, cheTLSSecret, clusterAPI.Scheme); err != nil {
```
tolusha, line 343: Is it better to move this code to `getSpecSecret`
We set controllerRefence for all components at getSpec func

pkg/deploy/secret.go
```diff
@@ -32,3 +101,24 @@ func NewSecret(cr *orgv1.CheCluster, name string, data map[string][]byte) *corev
 		Data: data,
 	}
 }
+
+// CreateTLSSecretFromRoute creates TLS secret with given name which contains certificates obtained from give url.
+// If the url is empty string, then router certificate will be obtained.
+// Works only on Openshift family infrastructures.
+func CreateTLSSecretFromRoute(checluster *orgv1.CheCluster, url string, name string, clusterAPI ClusterAPI) (err error) {
+	secret := &corev1.Secret{}
+	if err := clusterAPI.Client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: checluster.Namespace}, secret); err != nil && errors.IsNotFound(err) {
```
tolusha, line 110: It seems we will create a test route on every reconcile loop.

pkg/deploy/secret.go
```diff
@@ -32,3 +101,24 @@ func NewSecret(cr *orgv1.CheCluster, name string, data map[string][]byte) *corev
...
+		crtBytes, err := GetEndpointTLSCrtBytes(checluster, url, clusterAPI)
```
tolusha, line 111: It seems we will create a test route on every reconcile loop.

pkg/deploy/tls.go
mmorhun: The reason why we add controller reference here is that the secret could be created by self-signed secrets generator job which doesn't set any reference.

pkg/deploy/tls.go
mmorhun: The same as above

deploy/crds/org_v1_che_crd.yaml
```diff
@@ -422,13 +422,9 @@ spec:
...
                   type: string
                 selfSignedCert:
-                  description: Enables the support of OpenShift clusters whose router
-                    uses self-signed certificates. When enabled, the operator retrieves
-                    the default self-signed certificate of OpenShift routes and adds
-                    it to the Java trust store of the Che server. This is usually
-                    required when activating the `tlsSupport` field on demo OpenShift
-                    clusters that have not been setup with a valid certificate for
-                    the routes. This is disabled by default.
+                  description: Obsolete. The value of this flag is ignored. Che operator
```
tolusha, line 425: Obsolete - Deprecated

deploy/crds/org_v1_che_crd.yaml
```diff
@@ -446,10 +442,9 @@ spec:
                     its CA cert to be able to request it. This is disabled by default.
                   type: string
                 tlsSupport:
-                  description: 'Instructs the operator to deploy Che in TLS mode,
-                    ie with TLS routes or ingresses. This is disabled by default.
-                    WARNING: Enabling TLS might require enabling the `selfSignedCert`
-                    field also in some cases.'
+                  description: Instructs the operator to deploy Che in TLS mode. This
```
tolusha, line 445: Move `Deprecated` to the beginning of the description

deploy/crds/org_v1_che_crd.yaml
tolusha: Make changes in che_types and update olm filesd

pkg/deploy/tls.go
```diff
@@ -0,0 +1,448 @@
...
+
+	if util.IsOpenShift {
+		// Get router TLS certificates chain
+		peerCertificates, err := GetEndpointTLSCrtChain(checluster, "", clusterAPI)
+		if err != nil {
+			return false, err
+		}
+
+		// Check the chain if ti contains self-signed CA certificate
+		for i := range peerCertificates {
```
AndrienkoAleksandr, line 73: You can avoid using index of element and simple get element from range: `for _, cert := range arr {`

pkg/deploy/tls.go
```diff
@@ -0,0 +1,448 @@
...
+	peerCertificates, err := GetEndpointTLSCrtChain(instance, endpointURL, clusterAPI)
+	if err != nil {
+		if util.IsTestMode() {
+			fakeCrt := make([]byte, 5)
+			return fakeCrt, nil
+		}
+		return nil, err
+	}
+
+	for i := range peerCertificates {
```
AndrienkoAleksandr, line 181: You can avoid using index of element and simple get element from range: `for _, peerCertificate := range peerCertificates {`

deploy/crds/org_v1_che_crd.yaml
mmorhun: ok

pkg/deploy/tls.go
mmorhun: done

nickboldt:
will this fix be backported to the 7.14.x branch? Seems like a great UX improvement for CRW 2.2 too

mmorhun:
Rebased and updated OLM files

PR #307, opened 2020-06-11
Added buildx support in cico scripts

aditijadhav38:
Signed-off-by: Aditi Jadhav <aditi.jadhav@ibm.com>

### DIFF
--- a/cico_build_nightly.sh
+++ b/cico_build_nightly.sh
@@ -22,5 +22,6 @@ export SCRIPT_DIR
 
 load_jenkins_vars
 install_deps
+check_buildx_support
 set_nightly_tag
 build_and_push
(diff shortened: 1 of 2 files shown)


davidfestal:
cc @benoitf

nickboldt:
As the cico scripts are no longer being maintained ([centos CI is EOL](https://github.com/eclipse/che/issues/17857)) this PR will never be merged.

Instead, please submit a new PR with code that calls a PLATFORMS file and runs buildx like this:

https://github.com/eclipse/che-plugin-registry/blob/master/.ci/sidecar-build-publish.sh#L31-L35
or like this:

https://github.com/che-incubator/che-dashboard-next/blob/master/.github/workflows/dockerimage.yml#L35

See also: 

* [Move Che release workflow from Centos CI #17857](https://github.com/eclipse/che/issues/17857)
* https://github.com/eclipse/che-devfile-registry/pull/309
* https://github.com/eclipse/che-plugin-registry/pull/723

PR #310, opened 2020-06-17
CRW-979 use strings.TrimPrefix instead of...

nickboldt:
CRW-979 use strings.TrimPrefix instead of strings.TrimLeft

Change-Id: I7969e88c84ab211738fe071313ccc0fae4bac25c


nickboldt:
See #315

PR #311, opened 2020-06-17
CRW-979 use strings.TrimPrefix instead of...

nickboldt:
CRW-979 use strings.TrimPrefix instead of strings.TrimLeft

Change-Id: I7969e88c84ab211738fe071313ccc0fae4bac25c

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -253,7 +253,7 @@ func GenerateProxyJavaOpts(proxyURL string, proxyPort string, nonProxyHosts stri
 		}
 	}
 
-	proxyHost := strings.TrimLeft(proxyURL, "https://")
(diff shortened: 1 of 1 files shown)


nickboldt:
I don't see what needs to be updated. The test passed. :shrug:

gorkem:
@AndrienkoAleksandr Curious, How did you approve this PR? Did you test it for instance against the url that is originally reported broken "http://proxy.customer.com" ?

davidfestal:
> I don't see what needs to be updated. The test passed. 🤷

I somehow agree that we should add a test with the proxy URL that initially produced the failure.
So that if someone changes the code back to `strings.TrimLeft`, the error will this time be caught by unit tests.

mmorhun:
I've created a separate PR which must fix the issue: https://github.com/eclipse/che-operator/pull/315

PR #316, opened 2020-06-24
Fix typo in eclipse-che-preview-openshift.crd.yaml

tonyjames:


tolusha:
There is no such an error in the current version of CRD.

PR #320, opened 2020-06-25
Remove tls conditions from source code

mmorhun:

### What does this PR do?
Removes tls support flag from source code which makes it impossible to deploy Eclipse Che without TLS any more.
`tlsSupport` fields is still present in Che custom resource, however it is only for backward compatibility and the flag has no effect any more.

https://github.com/eclipse/che/issues/17090

### Tests
Testing was done on CRC with the folowing scenario:
1. Deploy Eclipse Che without TLS support
2. Create and start a workspace
3. Stop the workspace
4. Update operator image
5. Wait until new operator reconcile
6. Import self-signed-certificate into browser (`chectl cacert:export` helpful here)
7. Open dashboard and start the workspace
8. Open Che-Theia (welcome webview works now), build and run an embedded project via a command

### DIFF
--- a/deploy/crds/org_v1_che_cr.yaml
+++ b/deploy/crds/org_v1_che_cr.yaml
@@ -35,8 +35,6 @@ spec:
     # If enabled then the certificate from `che-git-self-signed-cert` config map
     # will be propagated to the Che components and provide particular configuration for Git.
     gitSelfSignedCert: false
-    # TLS mode for Che. It is not recommended to turn this off.
-    tlsSupport: true
     # protocol+hostname of a proxy server. Automatically added as JAVA_OPTS and https(s)_proxy
     # to Che server and workspaces containers
     proxyURL: ''
--- a/pkg/apis/org/v1/che_types.go
+++ b/pkg/apis/org/v1/che_types.go
@@ -127,10 +127,8 @@ type CheClusterSpecServer struct {
 	// configuration for Git.
 	// +optional
 	GitSelfSignedCert bool `json:"gitSelfSignedCert"`
-	// Deprecated.
-	// Instructs the operator to deploy Che in TLS mode.
-	// This is enabled by default.
-	// Disabling TLS may cause malfunction of some Che components.
+	// Deprecated. The value of this flag is ignored.
+	// Operator always deploys Che in TLS mode.
 	// +optional
 	TlsSupport bool `json:"tlsSupport"`
 	// Public URL of the Devfile registry, that serves sample, ready-to-use devfiles.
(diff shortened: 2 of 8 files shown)


nickboldt:
How long ago did we announce this was deprecated? @l0rd isn't the policy something like 3 sprints after deprecation, THEN we can remove API?

I'm all for removing cruft but do we need to at least WARN people since this DOES affect users and customers?

l0rd:
> How long ago did we announce this was deprecated? @l0rd isn't the policy something like 3 sprints after deprecation, THEN we can remove API?
> 
> I'm all for removing cruft but do we need to at least WARN people since this DOES affect users and customers?

I agree with @nickboldt. And I have to disagree myself of just a couple of weeks ago.

I mean removing unsecure HTTP support is NOT removing a fully working API but an API that [got (partially) broken 6 months ago](https://github.com/eclipse/che/issues/15283), so that's not a big concern but...we agreed that we would remove unsecure HTTP support only after we [had improved self-signed certificates scenarios UX](https://github.com/eclipse/che/issues/16764) and even if #16764 got fixed:
- we still didn't get any feedback on automatic configuration from real world customers (we need to wait CRW 2.2 release for that)
- more important I figured out that some customers will never accept to install an untrusted certificate on their local workstations. Hence self-signed is not an option for them until we get rid of the local certificate installation requirement (i.e. [this](https://github.com/eclipse/che/issues/12914) get fixed).

In other words a broken, unsecure Che is still better than no-Che at all. Let's wait 3 sprints to get some feedback on automatic TLS configuration and https://github.com/eclipse/che/issues/12914 resolution.

In the meantime we can start warning the community on che-dev mailing list. @mmorhun @tolusha are you ok with that? That means putting this work on hold and can be frustrating but it looks the better thing to do right now.

tolusha:
@l0rd 
So, We are going to remove unsecure HTTP once [https://github.com/eclipse/che/issues/12914](https://github.com/eclipse/che/issues/12914) is done and after warning the community,

But what I don't understand what is the difference between  `fully working API` and `API that got (partially) broken 6 months ago`. Could you explain pls?

l0rd:
> But what I don't understand what is the difference between `fully working API` and `API that got (partially) broken 6 months ago`. Could you explain pls?

@tolusha that was because @nickboldt was asking: "isn't the policy something like 3 sprints after deprecation, THEN we can remove API?" (the "API" that we want to remove is the no-TLS option). And my reply was that is true, we should first deprecate, wait and then remove, but in this case the situation is slightly different because the no-TLS "API" is broken since 6 months (i.e. Che Theia webview doesn't work).

mmorhun:
PR is heavily outdated.

PR #330, opened 2020-07-08
Add property to allow use authenticator for proxies;

mshaposhnik:
To be able to work with authenticated proxies, special Java authenticator should be used. Besides that, using custom authenticator requires special properties  to be set.
```
jdk.http.auth.proxying.disabledSchemes=
jdk.http.auth.tunneling.disabledSchemes=
```

More info: 
https://www.oracle.com/java/technologies/javase/8u111-relnotes.html
https://stackoverflow.com/questions/41806422/java-web-start-unable-to-tunnel-through-proxy-since-java-8-update-111
(yes java 11 requires it too)


tolusha:
Once I merge https://github.com/eclipse/che-operator/pull/272 you have to update yours.
The changes will go in https://github.com/eclipse/che-operator/blob/certs/pkg/deploy/proxy.go#L124

PR #331, opened 2020-07-08
Add tls support to che operator updates

flacatus:
This PR add the ability to use tls support in CI.

Reference issue: https://github.com/eclipse/che/issues/17337
Signed-off-by: flacatus <flacatus@redhat.com>

### DIFF
--- a/.ci/cico_updates_minikube.sh
+++ b/.ci/cico_updates_minikube.sh
@@ -86,7 +86,21 @@ waitCheUpdateInstall() {
   fi
 }
 
+self_signed_minikube() {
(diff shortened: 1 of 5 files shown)


.ci/cico_updates_minikube.sh
```diff
@@ -86,7 +86,21 @@ waitCheUpdateInstall() {
   fi
 }
 
+self_signed_minikube() {
```
tolusha, line 89: We don't have to generate certificate for minikube.
It covers by che-operator itself.

flacatus: yes, you are right. I'll remove that

PR #370, opened 2020-07-28
CRW-445 remove tls and self-signed certs params from csv

nickboldt:
remove tls and self-signed certs params from cvs template

Change-Id: I7c90faaab8387fc89a459ed8673abe5be2365365

Fix for:

https://issues.redhat.com/browse/CRW-445


flacatus:
@nickboldt There is a bug in CI with Minishift job. Please skip.

nickboldt:
@flacatus does that mean you approve of this change?

tolusha:
@nickboldt 
This is correct place.
Pls update kubernetes files as well.

nickboldt:
Thanks! I've also cherrypicked this to 7.16.x branch so we can have it in CRW 2.3.

PR #378, opened 2020-07-31
Manage the Traefik gateway for implementing single host on OpenShift

metlos:
This add support for managing the singlehost gateway using the operator.

The code is a little bit rough around the edges but the idea is there and it works.

issue: https://github.com/eclipse/che/issues/17065


pkg/deploy/defaults.go
```diff
@@ -76,6 +76,11 @@ const (
 	DefaultSecurityContextFsGroup   = "1724"
 	DefaultSecurityContextRunAsUser = "1724"
 
+	DefaultServerExposureStrategy              = "multi-host"
+	DefaultSingleHostWorkspaceExposureType     = "native"
+	DefaultSingleHostGatewayImage              = "traefik:v2.2.8"
```
sparkoo, line 81:
```suggestion
	DefaultSingleHostGatewayImage              = "docker.io/traefik:v2.2.8"
```

sparkoo (commented):
Beside what @metlos mentioned in the description, I see few major things missing:
 - [x] configure single ingress/route to che-gateway service
 - [ ] configure gateway to handle che-server traffic

deploy/crds/org_v1_che_cr.yaml
```diff
@@ -56,6 +56,27 @@ spec:
     workspaceNamespaceDefault: ''
     # defines if user is able to specify namespace different from the default
     allowUserDefinedWorkspaceNamespaces: false
+    # Sets the server and workspaces exposure type. Possible values are "multi-host", "single-host", "default-host".
+    # Defaults to "multi-host" which creates a separate ingress (or route on OpenShift) for every requied
+    # endpoint.
+    # "single-host" creates a single ingress/route and exposes workspaces on subpaths. Please read the docs
```
sparkoo, line 62: this is true only if `singleHostWorkspaceExposureType=gateway` or defaults on OpenShift, right?

deploy/crds/org_v1_che_cr.yaml
```diff
@@ -56,6 +56,27 @@ spec:
...
+    # to learn about the limitations of this approach.
+    # "default-host" exposes che server on the host of the cluster. Please read the docs to learn about
+    # the limitations of this approach.
+    serverExposureStrategy: ''
+    # When the serverExposureStrategy is set to "single-host", the way server and and workspaces
```
sparkoo, line 67:
```suggestion
    # When the serverExposureStrategy is set to "single-host", the way server and workspaces
```

deploy/crds/org_v1_che_cr.yaml
```diff
@@ -56,6 +56,27 @@ spec:
...
+    # are exposed is further configured by this property. The possible values are "native" (which means
+    # that the server and workspaces are exposed using ingresses on K8s) or "gateway" where the server
+    # and workspaces are exposed using a custom gateway based on Traefik.
+    # "native" is only supported on Kubernetes, "gateway" is supported on both openshift and kubernetes.
+    # On OpenShift, this property defaults to "gateway". On Kubernetes, it defaults to "native".
+    singleHostWorkspaceExposureType: ''
+    # The image used for the gateway in the single host mode. Defaults to docker.io/traefik:v2.2.8.
```
sparkoo, line 74: I don't think we should write concrete image tags into the docs. It's already inconsistent before it's even merged to master :)

pkg/deploy/gateway.go
```diff
@@ -0,0 +1,446 @@
...
+// below functions declare the desired states of the various objects required for the gateway
+
+func getGatewayServerConfigSpec(instance *orgv1.CheCluster) corev1.ConfigMap {
+	data := `
+http:
+  routers:
+    che-server:
+      rule: "PathPrefix(` + "`/`" + `)"
+      service: che-server
+      priority: 100
```
sparkoo, line 230: IMHO che-server should have the lowest priority. We're using `priority: 100` for workspaces so this would not work https://github.com/eclipse/che/blob/master/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/external/TraefikGatewayRouteConfigGenerator.java#L185

sparkoo:
what about devfile/plugin registries and keycloak? Do we keep them separate from Traefik routing?

pkg/deploy/gateway.go
sparkoo, line 230:
```suggestion
      priority: 1
```

pkg/deploy/defaults.go
```diff
@@ -76,6 +79,12 @@ const (
 	DefaultSecurityContextFsGroup   = "1724"
 	DefaultSecurityContextRunAsUser = "1724"
 
+	DefaultServerExposureStrategy              = "multi-host"
+	DefaultKubernetesSingleHostExposureType    = "native"
+	DefaultOpenShiftSingleHostExposureType     = "gateway"
+	DefaultSingleHostGatewayImage              = "docker.io/traefik:v2.2.8"
+	DefaultSingleHostGatewayConfigSidecarImage = "quay.io/che-incubator/configbump:0.1.2"
```
sparkoo, line 86:
```suggestion
	DefaultSingleHostGatewayConfigSidecarImage = "quay.io/che-incubator/configbump:0.1.4"
```

pkg/deploy/plugin_registry.go
```diff
@@ -27,49 +27,93 @@ type PluginRegistryConfigMap struct {
...
+
 		if !util.IsOpenShift {
-			ingress, err := SyncIngressToCluster(checluster, PluginRegistry, "", PluginRegistry, 8080, clusterAPI)
-			if !util.IsTestMode() {
-				if ingress == nil {
-					logrus.Infof("Waiting on ingress '%s' to be ready", PluginRegistry)
-					if err != nil {
-						logrus.Error(err)
+			if useGateway {
+				cfg := GetGatewayRouteConfig(checluster, pluginRegistryGatewayConfig, "/"+PluginRegistry, 1, "http://"+PluginRegistry+":8080")
```
sparkoo, line 47: IMO registries and keycloak must have higher priority than che-server. I don't know what's the Treafik behavior with same priorities, so it may actually work, but I would keep it clear and set it to e.g. 10, to have prio 1 for che-server, 10 for registries and keycloak, 100 for workspaces

pkg/deploy/devfile_registry.go
```diff
@@ -27,49 +27,90 @@ type DevFileRegistryConfigMap struct {
...
 			}
 		} else {
-			route, err := SyncRouteToCluster(checluster, DevfileRegistry, "", DevfileRegistry, 8080, clusterAPI)
-			if !util.IsTestMode() {
-				if route == nil {
-					logrus.Infof("Waiting on route '%s' to be ready", DevfileRegistry)
-					if err != nil {
-						logrus.Error(err)
+			if useGateway {
+				cfg := GetGatewayRouteConfig(checluster, devfileRegistryGatewayConfig, "/"+DevfileRegistry, 1, "http://"+DevfileRegistry+":8080")
```
sparkoo, line 82: https://github.com/eclipse/che-operator/pull/378/files#r484326087

pkg/deploy/devfile_registry.go
```diff
@@ -27,49 +27,90 @@ type DevFileRegistryConfigMap struct {
...
+				host = DevfileRegistry + "-" + checluster.Namespace + "." + checluster.Spec.K8s.IngressDomain
+			} else {
+				host = cheHost + "/" + DevfileRegistry
 			}
 
-			ingressStrategy := util.GetValue(checluster.Spec.K8s.IngressStrategy, DefaultIngressStrategy)
-			if ingressStrategy == "multi-host" {
-				host = DevfileRegistry + "-" + checluster.Namespace + "." + checluster.Spec.K8s.IngressDomain
+			if useGateway {
+				cfg := GetGatewayRouteConfig(checluster, devfileRegistryGatewayConfig, "/"+DevfileRegistry, 1, "http://"+DevfileRegistry+":8080")
```
sparkoo, line 52: https://github.com/eclipse/che-operator/pull/378/files#r484326087

pkg/controller/che/che_controller.go
```diff
@@ -803,19 +831,15 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 					}
 				}
 			} else {
-				// create Keycloak route
-				route, err := deploy.SyncRouteToCluster(instance, "keycloak", "", "keycloak", 8080, clusterAPI)
-				if !tests {
-					if route == nil {
-						logrus.Info("Waiting on route 'keycloak' to be ready")
+				if useGateway {
+					cfg := deploy.GetGatewayRouteConfig(instance, "keycloak", "/auth", 1, "http://keycloak:8080")
```
sparkoo, line 835: https://github.com/eclipse/che-operator/pull/378/files#r484326087

pkg/controller/che/che_controller.go
```diff
@@ -777,17 +782,40 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+
 			// create Keycloak ingresses when on k8s
 			if !isOpenShift {
-				ingress, err := deploy.SyncIngressToCluster(instance, "keycloak", "", "keycloak", 8080, clusterAPI)
-				if !tests {
-					if ingress == nil {
-						logrus.Info("Waiting on ingress 'keycloak' to be ready")
+				if useGateway {
+					// try to guess where in the ingress-creating code the /auth endpoint is defined...
+					cfg := deploy.GetGatewayRouteConfig(instance, "keycloak", "/auth", 1, "http://keycloak:8080")
```
sparkoo, line 793: https://github.com/eclipse/che-operator/pull/378/files#r484326087

deploy/crds/org_v1_che_cr.yaml
```diff
@@ -56,6 +56,27 @@ spec:
...
+    # "single-host" creates a single ingress/route and exposes workspaces on subpaths. Please read the docs
+    # to learn about the limitations of this approach.
+    # "default-host" exposes che server on the host of the cluster. Please read the docs to learn about
+    # the limitations of this approach.
+    serverExposureStrategy: ''
+    # When the serverExposureStrategy is set to "single-host", the way server and and workspaces
+    # are exposed is further configured by this property. The possible values are "native" (which means
+    # that the server and workspaces are exposed using ingresses on K8s) or "gateway" where the server
+    # and workspaces are exposed using a custom gateway based on Traefik.
+    # "native" is only supported on Kubernetes, "gateway" is supported on both openshift and kubernetes.
```
davidfestal, line 71: Since this property only makes sense for the raw kubernetes case (it has only 1 possible value on OpenShift), why not adding it under the `K8S` section ?

deploy/crds/org_v1_che_cr.yaml
```diff
@@ -56,6 +56,27 @@ spec:
     workspaceNamespaceDefault: ''
     # defines if user is able to specify namespace different from the default
     allowUserDefinedWorkspaceNamespaces: false
+    # Sets the server and workspaces exposure type. Possible values are "multi-host", "single-host", "default-host".
+    # Defaults to "multi-host" which creates a separate ingress (or route on OpenShift) for every requied
```
davidfestal, line 60:
```suggestion
    # Defaults to "multi-host" which creates a separate ingress (or route on OpenShift) for every required
```

deploy/crds/org_v1_che_crd.yaml
```diff
@@ -460,6 +475,32 @@ spec:
                     signed with self-signed cert. So, Che server must be aware of
                     its CA cert to be able to request it. This is disabled by default.
                   type: string
+                singleHostExposureType:
```
davidfestal, line 478: Since this property only makes sense for the raw kubernetes case (it has only 1 possible value on OpenShift), why not adding it under the K8S section ?
IMO it would make it clearer and would avoid having 2 different default values according to the underlying K8S flavor.

pkg/apis/org/v1/che_types.go
```diff
@@ -236,6 +237,38 @@ type CheClusterSpecServer struct {
...
+
+	// When the serverExposureStrategy is set to "single-host", the way the server, registries and workspaces
+	// are exposed is further configured by this property. The possible values are "native" (which means
+	// that the server and workspaces are exposed using ingresses on K8s) or "gateway" where the server
+	// and workspaces are exposed using a custom gateway based on Traefik. All the endpoints whether backed by the ingress
+	// or gateway "route" always point to the subpaths on the same domain.
+	// "native" is only supported on Kubernetes, "gateway" is supported on both openshift and kubernetes.
+	// On OpenShift, this property defaults to "gateway". On Kubernetes, it defaults to "native".
+	// +optional
+	SingleHostExposureType string `json:"singleHostExposureType,omitempty"`
```
davidfestal, line 260: Since this property only makes sense for the raw kubernetes case (it has only 1 possible value on OpenShift), why not adding it under the `CheClusterSpecK8SOnly` section ?
IMO it would make it clearer and would avoid having 2 different default values according to the underlying K8S flavor.

pkg/controller/che/che_cr_validator.go
```diff
@@ -23,7 +23,15 @@ import (
 // - self-contradictory configurations
 // - configurations with which it is impossible to deploy Che
 func ValidateCheCR(checluster *orgv1.CheCluster, isOpenshift bool) error {
-	if !isOpenshift {
+	if isOpenshift {
+		exposureStrategy := checluster.Spec.Server.ServerExposureStrategy
+		exposureType := checluster.Spec.Server.SingleHostExposureType
+		if exposureStrategy != "" && exposureStrategy != "multi-host" {
+			if exposureType != "" && exposureType != "gateway" {
```
davidfestal, line 30: I assume you would not need this test and possible failure at runtime, if the `singleHostExposureType` was a K8S only field, just ignored for installations on OpenShift.

pkg/deploy/defaults.go
```diff
@@ -76,6 +79,12 @@ const (
 	DefaultSecurityContextFsGroup   = "1724"
 	DefaultSecurityContextRunAsUser = "1724"
 
+	DefaultServerExposureStrategy              = "multi-host"
+	DefaultKubernetesSingleHostExposureType    = "native"
+	DefaultOpenShiftSingleHostExposureType     = "gateway"
+	DefaultSingleHostGatewayImage              = "docker.io/traefik:v2.2.8"
```
davidfestal, line 85: I assume that the docker image references here should also be configurable through `RELATED_IMAGE_xxx` properties as done for the other installation components, to enable the support of downstreaming and digest management.
cc @tolusha

pkg/controller/che/che_controller.go
```diff
@@ -269,6 +270,9 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 		// Che cannot be deployed with current configuration.
 		// Print error message in logs and wait until the configuration is changed.
 		logrus.Error(err)
+		if err := r.SetStatusDetails(instance, request, failedValidationReason, err.Error(), ""); err != nil {
```
sparkoo, line 273: what is this good for and how is it related to single-host ?

pkg/controller/che/che_cr_validator.go
```diff
@@ -23,7 +23,15 @@ import (
 // - self-contradictory configurations
 // - configurations with which it is impossible to deploy Che
 func ValidateCheCR(checluster *orgv1.CheCluster, isOpenshift bool) error {
-	if !isOpenshift {
+	if isOpenshift {
+		exposureStrategy := checluster.Spec.Server.ServerExposureStrategy
+		exposureType := checluster.Spec.Server.SingleHostExposureType
+		if exposureStrategy != "" && exposureStrategy != "multi-host" {
```
sparkoo, line 29: isn't `if exposureStrategy == "single=host" && exposureType != "gateway"` the same but simpler?

pkg/deploy/defaults.go
sparkoo: I think it already is
```
# The image used for the gateway in the single host mode. Defaults to docker.io/traefik:v2.2.8.
singleHostGatewayImage: ''
# The image used for the gateway sidecar that provides configuration to the gateway.
# Default to quay.io/che-incubator/configbump:latest
singleHostGatewayConfigSidecarImage: ''

File: pkg/deploy/defaults.go [hunk 20]
```diff
@@ -279,6 +288,16 @@ func GetCheMultiUser(cr *orgv1.CheCluster) string {
 	return DefaultCheMultiUser
 }
 
+func GetSingleHostExposureType(cr *orgv1.CheCluster) string {
+	dflt := DefaultKubernetesSingleHostExposureType
```
sparkoo, line 292: `dflt` please no :D

pkg/deploy/devfile_registry.go
```diff
@@ -27,49 +27,90 @@ type DevFileRegistryConfigMap struct {
 }
 
 const (
-	DevfileRegistry = "devfile-registry"
+	DevfileRegistry              = "devfile-registry"
+	devfileRegistryGatewayConfig = "devfile-registry-gtw"
```
sparkoo, line 31: I've named these like `che-gateway-route-devfile-registry` in helm. I don't like `gtw` in your naming. Otherwise I'm not suggesting to change. Might be good to sync operator and helm. wdyt ?

pkg/deploy/devfile_registry.go
```diff
@@ -27,49 +27,90 @@ type DevFileRegistryConfigMap struct {
...
  */
-func SyncDevfileRegistryToCluster(deployContext *DeployContext) (bool, error) {
+func SyncDevfileRegistryToCluster(deployContext *DeployContext, cheHost string) (bool, error) {
 	devfileRegistryURL := deployContext.CheCluster.Spec.Server.DevfileRegistryUrl
 	if !deployContext.CheCluster.Spec.Server.ExternalDevfileRegistry {
 		var host string
+		exposureStrategy := util.GetServerExposureStrategy(deployContext.CheCluster, DefaultServerExposureStrategy)
+		singleHostExposureType := GetSingleHostExposureType(deployContext.CheCluster)
+		useGateway := exposureStrategy == "single-host" && (util.IsOpenShift || singleHostExposureType == "gateway")
 		if !util.IsOpenShift {
```
sparkoo, line 44: I feel like this could be simplified and extracted to functions with something like
```
if isOpenshift {
  if exposureStrategy == "multi-host" {
    exposeOpenshiftMultihost(...)
  } else if exposureStrategy == "single-host" {
    exposeOpenshiftSinglehost(...) 
  }
} else {
  if multi-host {
    exposeKubeMulti()
  } else if single-host {
 ......
```
wdyt?

pkg/deploy/devfile_registry.go
```diff
@@ -27,49 +27,90 @@ type DevFileRegistryConfigMap struct {
...
  * Create devfile registry resources unless an external registry is used.
  */
-func SyncDevfileRegistryToCluster(deployContext *DeployContext) (bool, error) {
+func SyncDevfileRegistryToCluster(deployContext *DeployContext, cheHost string) (bool, error) {
 	devfileRegistryURL := deployContext.CheCluster.Spec.Server.DevfileRegistryUrl
 	if !deployContext.CheCluster.Spec.Server.ExternalDevfileRegistry {
 		var host string
+		exposureStrategy := util.GetServerExposureStrategy(deployContext.CheCluster, DefaultServerExposureStrategy)
+		singleHostExposureType := GetSingleHostExposureType(deployContext.CheCluster)
+		useGateway := exposureStrategy == "single-host" && (util.IsOpenShift || singleHostExposureType == "gateway")
```
sparkoo, line 43: do we really need this? on the places we want to ask whether to use gateway, we should already know that we're on single-host right? Also it should not be possible to make it here on openshift with `exposureStrategy` set to `single-host` and `singleHostExposureType` set to anything different than `gateway`. Such configuration should fail earlier.

pkg/deploy/gateway.go
```diff
@@ -0,0 +1,494 @@
...
+		},
+	}
+
+	return delete(deployContext.ClusterAPI, obj)
+}
+
+// below functions declare the desired states of the various objects required for the gateway
+
+func getGatewayServerConfigSpec(instance *orgv1.CheCluster) corev1.ConfigMap {
+	return GetGatewayRouteConfig(instance, gatewayServerConfigName, "/", 1, "http://che-host:8080", false)
```
sparkoo, line 287: is `che-host:8080` hardcoded when the service is created? Would it make sense to put it into variable so the actual service is synced with route in the code?

deploy/crds/org_v1_che_crd.yaml
metlos: I have put it here because of the supposed Ingress v2 coming at some point in the future. But otherwise you're of course right, it would make more sense under k8s.

pkg/deploy/identity_provider.go
```diff
@@ -0,0 +1,265 @@
...
+	exposureStrategy := util.GetServerExposureStrategy(instance, DefaultServerExposureStrategy)
+	singleHostExposureType := GetSingleHostExposureType(instance)
+	useGateway := exposureStrategy == "single-host" && (util.IsOpenShift || singleHostExposureType == "gateway")
+
+	// create Keycloak ingresses when on k8s
+	var keycloakURL string
+	if !isOpenShift {
+		if useGateway {
+			// try to guess where in the ingress-creating code the /auth endpoint is defined...
+			cfg := GetGatewayRouteConfig(instance, "keycloak", "/auth", 10, "http://keycloak:8080", false)
```
sparkoo, line 59: you've named all gateway routes configmaps with `-gtw` suffix, but keycloak one is named only `keycloak`. Sync it please so it's more clear.

pkg/deploy/devfile_registry.go
sparkoo: the same applies for plugin-registry and keycloak.

deploy/crds/org_v1_che_cr.yaml
sparkoo: I think it's still valid. `"single-host" creates a single ingress/route` is imho not true

deploy/crds/org_v1_che_cr.yaml
sparkoo: why this is resolved? isn't it a typo?

deploy/crds/org_v1_che_cr.yaml
sparkoo: why resolved? `quay.io/che-incubator/configbump:latest` is still out of sync, so this comment that we should not write concrete versions into code docs still applies

sparkoo (commented):
It's very hard to review the code with so much if/else blocks and without any tests. From manual testing it is working for me, but I'm wondering how many cases are not covered here.

Unfortunately, this PR does not decrease code-quality of che-operator :crying_cat_face:, but does not make things better either. So most of my comments are only suggestions.

deploy/crds/org_v1_che_crd.yaml
davidfestal: OK. I understand.  Thanks for the explanation.

pkg/deploy/defaults.go
davidfestal: I'm not saying _configurable from the `CheCluster` custom resource_. This is clearly implemented.
I'm speaking of the ability to configure the default value itself (when nothing is defined in the CR) with RELATED_IMAGES_XXX environment variables passed to the operator deployment, by the OLM ClusterServiceVersion manifest. This is a requirement for downstreaming and for being compatible with Openshift digest management and disconnected OLM installation mode.

To be more precise:
- I don't see a GO variable for each of the new docker images [here](https://github.com/eclipse/che-operator/blob/c4484c0c56dd5a002acf48b0a2cf5d5f20bc13b5/pkg/deploy/defaults.go#L30), but default values of the new docker images are provided as constants.
- I don't see the default values of the new docker images initialized based on environment variables in the [`InitDefaultsFromEnv`](https://github.com/eclipse/che-operator/blob/c4484c0c56dd5a002acf48b0a2cf5d5f20bc13b5/pkg/deploy/defaults.go#L108) method for example.

pkg/controller/che/che_controller.go
metlos: As the comment says:
```
		// Che cannot be deployed with current configuration.
		// Print error message in logs and wait until the configuration is changed.
```

pkg/controller/che/che_cr_validator.go
metlos: Yep, I moved `singleHostExposureType` under K8s and removed this validation logic.

pkg/deploy/defaults.go
metlos: Thanks for the suggestion. I believe I added the necessary code but I am a little bit ignorant about how OLM and chectl interact. With this change, the operator is not functional when deployed using chectl, because the operator deployment lacks the above env vars.

pkg/deploy/defaults.go
metlos: I simplified the logic such that the var is no longer ncssry. I stnd by dflt though :)

pkg/deploy/identity_provider.go
metlos: Good point. Renamed them all to `che-gateway-route-*`.

metlos:
https://ci.centos.org/job/devtools-che-operator-code-pr-check/698/ fails for environmental reasons I believe. I will try to re-run it until it passes or the failure reason changes.

metlos:
> It's very hard to review the code with so much if/else blocks and without any tests. From manual testing it is working for me, but I'm wondering how many cases are not covered here.
>

Yeah, but the operator testsuite is frankly beyond my comprehension. It is definitely something we need to improve but I am not sure this PR should be the place where we make sense of the testsuite.
 
> Unfortunately, this PR does not decrease code-quality of che-operator :crying_cat_face:, but does not make things better either. So most of my comments are only suggestions.

I sort tried to balance the ugliness, the amount of change and the actual implementation of the feature. I tried to keep the changes to the minimum where I was only modifying the existing logic and thus the code is fairly ugly there. If you look in https://github.com/eclipse/che-operator/blob/62facaa4dfe0b3f31a8de36f2aa79c65c32d7469/pkg/deploy/gateway.go for example where most of the new code resides, I think I stayed as succint and non-ugly as I could in Go.

metlos:
[ci-test]

metlos:
I learned that the `ci.centos.org PR check` job should be ignored, because it is obsolete.

flacatus:
@metlos sorry for delay. Yes you're right. Right now deploy team is migrating jobs from ci.centos.org to github actions and openshift ci... For some reason this job was not deleted completely.

deploy/crds/org_v1_che_cr.yaml
```diff
@@ -124,6 +142,13 @@ spec:
...
     securityContextRunAsUser: ''
-  metrics:
+    # When the serverExposureStrategy is set to "single-host", the way the server, registries and workspaces
+    # are exposed is further configured by this property. The possible values are "native" (which means
+    # that the server and workspaces are exposed using ingresses on K8s) or "gateway" where the server
+    # and workspaces are exposed using a custom gateway based on Traefik. All the endpoints whether backed by the ingress
+    # or gateway "route" always point to the subpaths on the same domain.
+    # Defaults to "native".
+    singleHostExposureType: ''
+    metrics:
```
tolusha, line 152: Indentation is not correct.
`metrics` should be on the level like `server`.

pkg/controller/che/create.go
```diff
@@ -249,6 +168,14 @@ func (r *ReconcileChe) GenerateAndSaveFields(deployContext *deploy.DeployContext
 		}
 	}
 
+	serverExposureStrategy := util.GetValue(deployContext.CheCluster.Spec.Server.ServerExposureStrategy, deploy.DefaultServerExposureStrategy)
+	if len(deployContext.CheCluster.Spec.Server.ServerExposureStrategy) < 1 {
+		deployContext.CheCluster.Spec.Server.ServerExposureStrategy = serverExposureStrategy
```
tolusha, line 173: Do we really need to update `spec.server.serverExposureStrategy` with a default value ?
Could we get rid of it ?

pkg/controller/che/che_controller.go
```diff
@@ -761,128 +764,17 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
-				doInstallOpenShiftoAuthProvider := instance.Spec.Auth.OpenShiftoAuth
-				if doInstallOpenShiftoAuthProvider {
-					openShiftIdentityProviderStatus := instance.Status.OpenShiftoAuthProvisioned
-					if !openShiftIdentityProviderStatus {
-						if err := r.CreateIdentityProviderItems(instance, request, cheFlavor, deploy.KeycloakDeploymentName, isOpenShift4); err != nil {
-							return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 5}, err
-						}
-					}
-				}
+	provisioned, err := deploy.SyncIdentityProviderToCluster(deployContext, cheHost, protocol, cheFlavor)
```
tolusha, line 767: I like this refactoring.

pkg/deploy/deployment.go
```diff
@@ -31,7 +31,7 @@ var deploymentDiffOpts = cmp.Options{
 	cmpopts.IgnoreFields(appsv1.DeploymentSpec{}, "Replicas", "RevisionHistoryLimit", "ProgressDeadlineSeconds"),
 	cmpopts.IgnoreFields(appsv1.DeploymentStrategy{}, "RollingUpdate"),
 	cmpopts.IgnoreFields(corev1.Container{}, "TerminationMessagePath", "TerminationMessagePolicy"),
-	cmpopts.IgnoreFields(corev1.PodSpec{}, "DNSPolicy", "SchedulerName", "SecurityContext"),
+	cmpopts.IgnoreFields(corev1.PodSpec{}, "DNSPolicy", "SchedulerName", "SecurityContext", "DeprecatedServiceAccount", "TerminationGracePeriodSeconds"),
```
tolusha, line 34: Agree with "DeprecatedServiceAccount" but why not to add "TerminationGracePeriodSeconds" field to the deployment instead of adding to IgnoreFields?

pkg/controller/che/che_controller.go
```diff
@@ -761,128 +764,17 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+		if !provisioned {
+			if err != nil {
+				logrus.Errorf("Error provisioning the identity provider to cluster: %v", err)
 			}
+			return reconcile.Result{RequeueAfter: time.Second * 1}, err
 		}
 	}
 
-	provisioned, err := deploy.SyncDevfileRegistryToCluster(deployContext)
+	provisioned, err = deploy.SyncDevfileRegistryToCluster(deployContext, cheHost)
```
tolusha, line 777: `cheHost` variable is defined below. So, you always pass `""` here.
Am I right?

pkg/deploy/gateway.go
```diff
@@ -0,0 +1,494 @@
...
+}
+
+func deleteAll(instance *orgv1.CheCluster, clusterAPI ClusterAPI) error {
+	deployment := appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      GatewayServiceName,
+			Namespace: instance.Namespace,
+		},
+	}
+	if err := delete(clusterAPI, &deployment); err != nil {
```
tolusha, line 100: if it is NotFound error, should be ignore it instead returning?

pkg/controller/che/create.go
metlos: The same argument could be made for the majority of other fields for which we update the default value. I was just trying to follow suite. Do you see a strong argument not to?

pkg/controller/che/che_controller.go
metlos: Well, I more or less just moved a giant if-block into a separate file :smile: It could be improved still.

pkg/controller/che/che_controller.go
metlos: I think `cheHost` is declared on https://github.com/eclipse/che-operator/pull/378/files/62facaa4dfe0b3f31a8de36f2aa79c65c32d7469#diff-d9445e614f0d728fd140771b0fd29c50R721 and first assigned on https://github.com/eclipse/che-operator/pull/378/files/62facaa4dfe0b3f31a8de36f2aa79c65c32d7469#diff-d9445e614f0d728fd140771b0fd29c50R733 for Kubernetes and on https://github.com/eclipse/che-operator/pull/378/files/62facaa4dfe0b3f31a8de36f2aa79c65c32d7469#diff-d9445e614f0d728fd140771b0fd29c50R752 for OpenShift.

I think you're referring to https://github.com/eclipse/che-operator/pull/378/files/62facaa4dfe0b3f31a8de36f2aa79c65c32d7469#diff-d9445e614f0d728fd140771b0fd29c50R864. I think that assignment is superfluous because the `instance.Spec.Server.CheHost` has been set on the lines mentioned above to be the same as `cheHost`. The control flow is a little bit confusing in that super massive method though so I may be wrong...

pkg/deploy/deployment.go
metlos: :+1:

pkg/deploy/gateway.go
metlos: `delete()` does a "fetch&delete" (which is an approach used on other places in the codebase, too), so a chance of that happening are IMHO low, but it is a good point nonetheless.

sparkoo:
this fails for me on minikube
1. installed as default -> multihost
1. changed to `single-host` in CheCluster
1. as a result of this, che server is on `che-che.192.168.39.127.nip.io`, others are on `192.168.39.127.nip.io`. Che then fails to reach registries because they are set to `che-che` subdomain 
```
  devfileRegistryURL: https://che-che.192.168.39.127.nip.io/devfile-registry
  pluginRegistryURL: https://che-che.192.168.39.127.nip.io/plugin-registry/v3
```

pkg/controller/che/create.go
tolusha: Well, we do have several places where sync values back to CR (our legacy).
But later @davidfestal and me realized that it is a bad practice and we try to avoid it as much as possible.

pkg/controller/che/che_controller.go
tolusha: Agree. Thank you for explanation.

pkg/deploy/gateway.go
```diff
@@ -0,0 +1,498 @@
...
+			return getErr
+		}
+		actual = nil
+	}
+
+	kind := blueprintObject.GetObjectKind().GroupVersionKind().Kind
+
+	if actual == nil {
+		logrus.Infof("Creating a new object: %s, name %s", kind, blueprint.GetName())
+		err := clusterAPI.Client.Create(context.TODO(), blueprintObject)
```
tolusha, line 179: I would recommend to add `errors.IsAlreadyExists(err)` check. like here https://github.com/eclipse/che-operator/blob/master/pkg/deploy/route.go#L57-L64
I came across on this recently.

tolusha: I guess it is needed to invoke `controllerutil.SetControllerReference` as well.
So, instead of duplicating it, it is better to move it to `get***Spec` functions.

pkg/deploy/gateway.go
```diff
@@ -0,0 +1,498 @@
...
+		}
+	} else {
+		actualMeta := actual.(metav1.Object)
+
+		diff := cmp.Diff(actual, blueprint, diffOpts)
+		if len(diff) > 0 {
+			logrus.Infof("Updating existing object: %s, name: %s", kind, actualMeta.GetName())
+			fmt.Printf("Difference:\n%s", diff)
+
+			err := clusterAPI.Client.Delete(context.TODO(), actual)
```
tolusha, line 191: I use Delete/Create approach to update the object for Ingress/Routes/Services only. I don't remember the exact reason. 
For another objects I use `Update`. 
I don't think that Delete/Update is good for deployment. Deleting requires stopping pods, so we will have AlreadyExists error on Create below. Besides that I check if pods are running
https://github.com/eclipse/che-operator/blob/master/pkg/deploy/deployment.go#L96-L103
In another word I am afraid if deployment is changed we have a lot errors in operator logs.

pkg/controller/che/create.go
metlos: I removed this. Thanks for the explanation.

metlos:
> this fails for me on minikube

Should be fixed now.

tolusha (approved):
Thank you for resolving all remarks

tolusha:
One question. Have your tried to change single-host / multi-host on the same deployment to check if it works?
It is interesting to check also operator logs during this if any errors are printed.

sparkoo (approved):

sparkoo:
what is behavior when you update the operator as you've deprecated the `ingressStrategy` property?

pkg/deploy/defaults.go
davidfestal: Thanks !

davidfestal (approved):

metlos:
> One question. Have your tried to change single-host / multi-host on the same deployment to check if it works?
> It is interesting to check also operator logs during this if any errors are printed.

Yes, this works nicely without openshift oauth. With openshift oauth there are some issues with authorization which I think are caused by the client requiring different paths etc.. Haven't gotten to the bottom of it - sometimes the problem disappears with time (I guess some session timeouts kicking in), sometimes it takes a long time... I would not consider it a blocker though. It's something we can investigate once this feature is in...

metlos:
> what is behavior when you update the operator as you've deprecated the ingressStrategy property?
An explicitly set `ingressStrategy` takes precedence over the implicit default of `serverExposureStrategy`. https://github.com/eclipse/che-operator/pull/378/files#diff-6c6a7fefe37261e2fa8a010cf9e7ae08R171

tolusha:
@flacatus 
Could you help with failing tests?

flacatus:
@tolusha @metlos The errors comes in ci/prow/v5-images job. We migrate recently to new olm bundles and I forgot to remove this job. My mistake.
PR to fix it: https://github.com/openshift/release/pull/11884

metlos:
The failing CI jobs seem to be caused by CI itself. There's a pending PR that should fix that: https://github.com/openshift/release/pull/11884. Thanks to @flacatus for a super fast fix.

PR #391, opened 2020-08-06
Set the default value for openShiftoAuth to true

tolusha:

### Reference issue 
https://github.com/eclipse/che/issues/17488

### What does this PR do
Set the default value for openShiftoAuth to true


tolusha:
@davidfestal 
Yes, I did the following tests:
- Configuring custom hostname on minikube, minishift, openshift
- Deploying without custom host name on minikube

tolusha:
It shouldn't since everything is protected by `if isOpenShift` clause in the operator code.
But I will do some tests.

tolusha:
@davidfestal 
everything works fine

flacatus:
I did as well some tests in cluster bot OCP 4.5 and works...

PR #420, opened 2020-08-27
Create a general deployContext to pass to all functions instead of pass numerous arguments

flacatus:
Reference issue: https://github.com/eclipse/che/issues/17589

Signed-off-by: Flavius Lacatusu <flacatus@redhat.com>

PR #421, opened 2020-08-28
Add server TLS key and cert into Che configmap if a separate namespace for Che workspaces is configured

mmorhun:

### What does this PR do
If Eclipse Che is configured to use a separate namespace for Che workspaces via `CHE_INFRA_KUBERNETES_NAMESPACE_DEFAULT` Che property, then add into Che config map Che server TLS secret (key and certificate) in order for Che server to be able to configure workspace components with right TLS key and certificate (as it is not possible to access Che server TLS secret from a separate namespace) .

### Which issues this PR resolves
https://github.com/eclipse/che/issues/17631

### DIFF
--- a/pkg/deploy/che_configmap.go
+++ b/pkg/deploy/che_configmap.go
@@ -77,7 +77,10 @@ type CheConfigMap struct {
 }
 
 func SyncCheConfigMapToCluster(deployContext *DeployContext) (*corev1.ConfigMap, error) {
-	data := GetCheConfigMapData(deployContext)
+	data, err := GetCheConfigMapData(deployContext)
+	if err != nil {
+		return nil, err
+	}
 	specConfigMap, err := GetSpecConfigMap(deployContext, CheConfigMapName, data)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


flacatus:
Please rebase with master changes; there some changes passing arguments to functions after merging: https://github.com/eclipse/che-operator/pull/420

pkg/deploy/che_configmap.go
```diff
@@ -228,19 +231,30 @@ func GetCheConfigMapData(cr *orgv1.CheCluster, proxy *Proxy) (cheEnv map[string]
...
+			"CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_RUN__AS__USER": securityContextRunAsUser,
+			"CHE_INFRA_KUBERNETES_INGRESS_DOMAIN":                      ingressDomain,
+			"CHE_INFRA_KUBERNETES_SERVER__STRATEGY":                    ingressStrategy,
+			"CHE_INFRA_KUBERNETES_TLS__SECRET":                         tlsSecretName,
+			"CHE_INFRA_KUBERNETES_INGRESS_ANNOTATIONS__JSON":           "{\"kubernetes.io/ingress.class\": " + ingressClass + ", \"nginx.ingress.kubernetes.io/rewrite-target\": \"/$1\",\"nginx.ingress.kubernetes.io/ssl-redirect\": " + tls + ",\"nginx.ingress.kubernetes.io/proxy-connect-timeout\": \"3600\",\"nginx.ingress.kubernetes.io/proxy-read-timeout\": \"3600\"}",
+			"CHE_INFRA_KUBERNETES_INGRESS_PATH__TRANSFORM":             "%s(.*)",
+		}
+		// Add TLS key and server certificate to properties when user workspaces should be created in another
+		// than Che server namespace, from where the Che TLS secret is not accesible.
+		if _, keyExists := cr.Spec.Server.CustomCheProperties["CHE_INFRA_KUBERNETES_NAMESPACE_DEFAULT"]; keyExists {
```
tolusha, line 246: We should also check if `che.infra.kubernetes.namespace != ""`
https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties#L262-L287

tolusha (approved):

pkg/deploy/che_configmap.go
mmorhun: I would like to avoid that check for deprecated property which causes problems (if believe the property description).

AndrienkoAleksandr (approved):

flacatus (approved):

PR #449, opened 2020-09-17
Define labels for route/ingress

tolusha:

### What does this PR do
Possibility to define label(s) for route to work with router sharding 

### Reference issue
https://issues.redhat.com/browse/CRW-1034


davidfestal (commented):
I'm not very convinced about the use of new `CheClusterSpecDevfileRegistry` and `CheClusterSpecPluginRegistry` sections.

These new sections would in fact contain only the label-related field, while most other registry-related settings would still be in the "Server" section.
Of course adding 2 new sections would completely make sense if we were ready to refactor the schema completely, and bring all the registry-related settings in the dedicated registry section. But I assume we would not start such a refactoring now since it which would break backward compatibility of the CRD schema.

So I'd rather follow the existing pattern, and use names like `devfileRegistryLabels` in the `server` section since we already had `devfileRegistryImage`, `devfileRegistryPullPolicy`, ...

davidfestal (approved):

pkg/deploy/labels.go
```diff
@@ -12,10 +12,21 @@
...
+
 	orgv1 "github.com/eclipse/che-operator/pkg/apis/org/v1"
 )
 
 func GetLabels(cr *orgv1.CheCluster, component string) (labels map[string]string) {
 	labels = map[string]string{"app": DefaultCheFlavor(cr), "component": component}
 	return labels
 }
+
+func MergeLabels(labels map[string]string, additionalLabels string) {
```
AndrienkoAleksandr, line 25: Sorry, that's enougth to have one additionalLabels ? If it is only one than it should be called additionalLabel. If we need list, than we should use array instead of string.

AndrienkoAleksandr: Oh, I see, comma separated/

AndrienkoAleksandr (approved):

PR #454, opened 2020-09-21
CRW-1229 use ArchitectureDependentEnv values...

nickboldt:
CRW-1229 use ArchitectureDependentEnv values instead of amd64-specific ones to match values set in defaults.go

Change-Id: I88022bd2b4faa09c1cbf489d75b198ddc98f19c1

### DIFF
--- a/pkg/deploy/defaults_test.go
+++ b/pkg/deploy/defaults_test.go
@@ -18,6 +18,7 @@ import (
 	"testing"
 
 	orgv1 "github.com/eclipse/che-operator/pkg/apis/org/v1"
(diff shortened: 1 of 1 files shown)


nickboldt:
If you like this PR, we should also cherry pick to 7.19.x so we don't break CRW 2.5 when we start those builds.

tolusha (approved):

nickboldt:
Cherrypicked to 7.19.x too:
https://github.com/eclipse/che-operator/commit/26cae0276dcefd99d9401294cc0b2fbe8063a34e

PR #455, opened 2020-09-21
Fix tag argument for update-nightly-bundle.

AndrienkoAleksandr:
### Referenced issue:

https://github.com/eclipse/che/issues/17922

### DIFF
--- a/olm/update-nightly-bundle.sh
+++ b/olm/update-nightly-bundle.sh
@@ -131,7 +131,7 @@ do
 
   if [[ -n "$TAG" ]]; then
     echo "[INFO] Set tags in nightly OLM files"
-    sed -i 's/'$RELEASE'/'$TAG'/g' ${NEW_CSV}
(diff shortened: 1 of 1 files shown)


olm/update-nightly-bundle.sh
```diff
@@ -131,7 +131,7 @@ do
 
   if [[ -n "$TAG" ]]; then
     echo "[INFO] Set tags in nightly OLM files"
-    sed -i 's/'$RELEASE'/'$TAG'/g' ${NEW_CSV}
+    sed -ri "s/(.*:\s?)7.19.0([^-])?$/\1nightly\2/" "${NEW_CSV}"
```
tolusha, line 134: Can we have a more generic regexp ?

AndrienkoAleksandr: Yes, sorry, initially it was.

tolusha (approved):

PR #468, opened 2020-09-23
Provide internal network urls for plugin registry, devfile registry and internal che api endpoint url.

AndrienkoAleksandr:
### Referenced issue:
https://github.com/eclipse/che/issues/17644


deploy/crds/org_v1_che_crd.yaml
```diff
@@ -507,6 +507,11 @@ spec:
                     TLS mode. This is enabled by default. Disabling TLS may cause
                     malfunction of some Che components.
                   type: boolean
+                useServiceHostNames:
```
sleshchenko, line 510: If you google hostname - it will be one word
```suggestion
                useServiceHostnames:
```

tolusha: or just `serviceHostnames` ?

AndrienkoAleksandr: As for me both variants are Ok:  serviceHostnames and useServiceHostnames

tolusha:
`.cluster.local`  is a cluster domain which configurable. 
I am wondering if we just can use `<service>.<namespace>.svc:<port>`

AndrienkoAleksandr:
> https://github.com/eclipse/che-operator/pull/468#issuecomment-698227384

Ok, I see what you meant. There are some tricky ways how to retrieve this info, we could right some "autodetector". But also I think we could provide field in case if this auto-detector will be not working. What do you think?

pkg/deploy/devfile_registry.go
```diff
@@ -50,10 +52,12 @@ func SyncDevfileRegistryToCluster(deployContext *DeployContext, cheHost string)
 			domain = cheHost
 			endpoint = domain + "/" + DevfileRegistry
 		}
+
+		routePathPrefix := fmt.Sprintf("%s://%s:%d", DevfileRegistryServiceProtocol, DevfileRegistry, DevfileRegistryServicePort)
```
metlos, line 56: This is the internal URL.

pkg/deploy/devfile_registry.go
```diff
@@ -151,6 +155,8 @@ func SyncDevfileRegistryToCluster(deployContext *DeployContext, cheHost string)
 			}
 		}
 
+		deployContext.InternalService.DevfileRegistryHost = fmt.Sprintf("%s://%s.%s.svc.cluster.local:%d", DevfileRegistryServiceProtocol, DevfileRegistry, deployContext.CheCluster.Namespace, DevfileRegistryServicePort)
```
metlos, line 158: The internal URL of a service is `serviceName.namespace.svc.cluster.local` where everything apart of `serviceName` is optional. The `namespace` portion resolves to the current namespace of the "caller" if I am not mistaken, so this URL is identical to the above defined `routePathPrefix` (which is not a prefix but a URL :) ).

Therefore we can use just 1 of the two for these both purposes IMHO.

pkg/deploy/plugin_registry.go
```diff
@@ -127,6 +131,8 @@ func SyncPluginRegistryToCluster(deployContext *DeployContext, cheHost string) (
 			}
 		}
 
+		deployContext.InternalService.PluginRegistryHost = fmt.Sprintf("%s://%s.%s.svc.cluster.local:%d/v3", PluginRegistryServiceProtocol, PluginRegistry, deployContext.CheCluster.Namespace, PluginRegistryServicePort)
```
metlos, line 134: Same comment as for the devfile registry. This and the above defined  `routePathPrefix` (which is a URL and not a prefix) resolve to the same thing and therefore there's no need to have them both.

pkg/deploy/devfile_registry.go
AndrienkoAleksandr: Code was very changed after rebase with master.

pkg/deploy/plugin_registry.go
AndrienkoAleksandr: Code was very changed after rebase with master.

pkg/deploy/devfile_registry.go
AndrienkoAleksandr: Code was very changed after rebase with master.

AndrienkoAleksandr:
@davidfestal Could you review, please?

deploy/crds/org_v1_che_cr.yaml
```diff
@@ -16,6 +16,7 @@ metadata:
   name: eclipse-che
 spec:
   server:
+    useServiceHostNames: false
```
davidfestal, line 19: Why is it `false` here while it is `true` in PR https://github.com/eclipse/che-operator/pull/507/files#diff-c9772ddc1ed08d9f54fa301ae58047756e33f73e5309795df6dca62d80411e51R19 ?

pkg/apis/org/v1/che_types.go
```diff
@@ -138,6 +138,15 @@ type CheClusterSpecServer struct {
 	// Disabling TLS may cause malfunction of some Che components.
 	// +optional
 	TlsSupport bool `json:"tlsSupport"`
+
+	// Use internal host names provided by cluster dns.
+	// It should be usefull to speed internal services traffic.
+	// Also it could avoid some proxy issues.
+	ServiceHostnames bool `json:"serviceHostnames"`
```
davidfestal, line 145: What is the expected default value for this field ?
If it's `true`, I assume that the Go type should be `*bool` instead of `bool` ?

davidfestal (commented):
I added 2 comments / questions

One additional remark would be that I don't see any additional test case in order to cover this new option / scenario. But maybe it is planned in a distinct PR ?

deploy/crds/org_v1_che_cr.yaml
AndrienkoAleksandr: This pr was created earlier. And initially we thought about default `false`. But looks like we can use true value to detect any `possible` regression faster. @tolusha @davidfestal Do you agree?

pkg/deploy/devfile_registry.go
AndrienkoAleksandr: > The internal URL of a service is serviceName.namespace.svc.cluster.local where everything apart of serviceName is optional.

@metlos As far I remember url with `serviceName` will be working inside one namespace. `serviceName.namespace.svc.cluster.local` should work between namespaces inside cluster.

deploy/crds/org_v1_che_cr.yaml
AndrienkoAleksandr: Turned "on" this option. Related mail list here: https://www.eclipse.org/lists/che-dev/msg03934.html . If we will get bad feed back for this default I'll move it back to false.

AndrienkoAleksandr:
> One additional remark would be that I don't see any additional test case in order to cover this new option / scenario. 

I will apply tests.

pkg/apis/org/v1/che_types.go
```diff
@@ -138,6 +138,16 @@ type CheClusterSpecServer struct {
...
 	TlsSupport bool `json:"tlsSupport"`
+
+	// Use internal host names provided by cluster dns.
+	// It should be usefull to speed internal services traffic.
+	// Also it could avoid some proxy issues.
+	// Default value is true.
+	ServiceHostnames *bool `json:"serviceHostnames"`
+
+	// Internal cluster DNS service hostname suffix. It's a root dns domain. "cluster.local" by default for Kube-DNS.
+	ServiceHostnameSuffix string `json:"serviceHostnameSuffix"`
```
tolusha, line 149: We need a way to detect cluster domain.

AndrienkoAleksandr: I applied draft method getClusterDomain

AndrienkoAleksandr: `getClusterDomain` gets cname for default service "kubernetes.default.svc" and we parse dns domain from request value. I tested it works with minikube, minishift, Openshift 4. 

A bit more context for Openshift 4: dns in the openshift 4 was implemented with operator: https://github.com/openshift/cluster-dns-operator . In the README.md of this operator I see `cluster.local` is default and in the code looks like this dns domain even hardcoded: https://github.com/openshift/cluster-dns-operator/blob/924a03d47a42b42c942722eb41c673bab4bd8030/pkg/operator/controller/controller.go#L329
And you can't configure dns domain using custom operator's resource - there isn't  any field to do that. You have only status field: ClusterDomain which has value `cluster.local` . Also you can't change dns domain `cluster.local` using dns config map - operator will erase this changes: https://github.com/openshift/cluster-dns-operator/blob/0d46f0303675473ce5d022904ddaff23132bac97/pkg/operator/controller/controller_dns_configmap.go#L90 , because .cluster.local is desired value.
Also I see uses  default service resolution https://github.com/openshift/cluster-dns-operator/blob/924a03d47a42b42c942722eb41c673bab4bd8030/pkg/operator/controller/controller.go#L158 and https://github.com/openshift/cluster-dns-operator/blob/924a03d47a42b42c942722eb41c673bab4bd8030/pkg/operator/controller/controller.go#L172 That's why I think our detector should work fine on the Openshift 4.

AndrienkoAleksandr: We decided to use services in format: service-name.namespace.svc

deploy/crds/org_v1_che_crd.yaml
```diff
@@ -555,6 +555,11 @@ spec:
                     signed with self-signed cert. So, Che server must be aware of
                     its CA cert to be able to request it. This is disabled by default.
                   type: string
+                serviceHostnames:
```
tolusha, line 558: maybe
`internalServiceHostnames`
but I am not usre

tolusha (approved):

deploy/crds/org_v1_che_crd.yaml
```diff
@@ -555,6 +555,11 @@ spec:
...
+                  description: Use internal host names provided by cluster dns. It
+                    should be usefull to speed internal services traffic. Also it
+                    could avoid some proxy issues. Default value is true.
```
tolusha, line 561: `speed` -> `speed up`

tolusha: `Use internal hostnames provided by cluster DNS to speed up the internal services traffic and avoid proxy issues. The default value is `true`.`

deploy/crds/org_v1_che_crd.yaml
tolusha: `useInternalClusterSVCNames`

deploy/crds/org_v1_che_crd.yaml
```diff
@@ -555,6 +555,11 @@ spec:
...
+                  type: boolean
```
tolusha, line 562: `useInternalClusterSVCNames`
```
Use internal cluster svc names to communicate between components to speed up the traffic and avoid proxy issues. The default value is `true`.
```

AndrienkoAleksandr:
Pr update and rebased.

tolusha (approved):

AndrienkoAleksandr:
Looks like quay.io has troubles: https://status.quay.io/

AndrienkoAleksandr:
"Minishift / Testing stable versions updates (pull_request)" - checked locally - passed.

AndrienkoAleksandr:
"Minishift / Testing latest changes" - checked locally - passed.

PR #484, opened 2020-10-02
CRW-1254 can we use the standard,...

nickboldt:
CRW-1254 can we use the standard, non-deprecated registry registry.redhat.io instead of registry.access.redhat.com for ubi minimal images? turns out NO, because no one wants to set up authentication from GH actions.

But we should use the safer path /ubi8/ubi-minimal vs. /ubi8-minimal so that OCP 4.x airgap mirroring works

Change-Id: I04112ce8cf9a3311872e7fb295fd742d75055d06

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,8 +25,8 @@ RUN export ARCH="$(uname -m)" && if [[ ${ARCH} == "x86_64" ]]; then export ARCH=
     export MOCK_API=true && go test -mod=vendor -v ./... && \
     GOOS=linux GOARCH=${ARCH} CGO_ENABLED=0 go build -mod=vendor -o /tmp/che-operator/che-operator cmd/manager/main.go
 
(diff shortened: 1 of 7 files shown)


nickboldt:
Hey so ...

* I ran the requested script - `./olm/update-nightly-bundle.sh` using `operator-sdk 0.10.1`
* and added the updated files in https://github.com/eclipse/che-operator/pull/484/commits/a3a4003c6ce024aacf8c63a1b6579740870a294e

but I'm still getting nightly opm bundle PR check failure:
```
Error:  Nightly bundle file deploy/olm-catalog/eclipse-che-preview-kubernetes/manifests/che-operator.clusterserviceversion.yaml should be updated in your pr, please. Use script 'che-operator/olm/update-nightly-bundle.sh' for this purpose.
```

Help?

tolusha (approved):

PR #491, opened 2020-10-12
Limit keycloak redirect urls

tolusha:

### Reference issue
https://github.com/eclipse/che/issues/17902

### DIFF
--- a/templates/keycloak_provision
+++ b/templates/keycloak_provision
@@ -20,7 +20,7 @@ if [ $? -eq 0 ]; then echo "Realm exists"; exit 0; fi \
                                         -s clientId=$keycloakClientId \
                                         -s id=$keycloakClientId \
                                         -s 'webOrigins=["http://$cheHost", "https://$cheHost"]' \
-                                        -s 'redirectUris=["http://$cheHost/dashboard/", "https://$cheHost/dashboard/", "http://$cheHost/workspace-loader/", "https://$cheHost/workspace-loader/", "http://$cheHost/_app/*", "https://$cheHost/_app/*"]' \
+                                        -s 'redirectUris=["http://$cheHost/dashboard/", "https://$cheHost/dashboard/", "http://$cheHost/workspace-loader/*", "https://$cheHost/workspace-loader/*", "http://$cheHost/_app/*", "https://$cheHost/_app/*"]' \
                                         -s 'directAccessGrantsEnabled'=true \
                                         -s publicClient=true \
 && $script create users   -s username=admin \
--- a/templates/keycloak_provision
+++ b/templates/keycloak_provision
@@ -20,7 +20,7 @@ if [ $? -eq 0 ]; then echo "Realm exists"; exit 0; fi \
                                         -s clientId=$keycloakClientId \
                                         -s id=$keycloakClientId \
                                         -s 'webOrigins=["http://$cheHost", "https://$cheHost"]' \
-                                        -s 'redirectUris=["http://$cheHost/dashboard/", "https://$cheHost/dashboard/", "http://$cheHost/workspace-loader/", "https://$cheHost/workspace-loader/", "http://$cheHost/_app/loader.html", "https://$cheHost/_app/loader.html"]' \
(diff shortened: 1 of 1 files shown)


templates/keycloak_provision
```diff
@@ -20,7 +20,7 @@ if [ $? -eq 0 ]; then echo "Realm exists"; exit 0; fi \
                                         -s clientId=$keycloakClientId \
                                         -s id=$keycloakClientId \
                                         -s 'webOrigins=["http://$cheHost", "https://$cheHost"]' \
-                                        -s 'redirectUris=["http://$cheHost/*", "https://$cheHost/*"]' \
+                                        -s 'redirectUris=["http://$cheHost/dashboard/", "https://$cheHost/dashboard/", "http://$cheHost/workspace-loader/", "https://$cheHost/workspace-loader/", "http://$cheHost/_app/loader.html", "https://$cheHost/_app/loader.html"]' \
```
sleshchenko, line 23: What about an ability to get token from localhost? I kept in mind running local dashboard/workspace-loader instance without modifying redirects uris manually on each new cluster, but I assume it could also help `chectl server:login` do login through web interface, as https://github.com/int128/kubelogin is able to do.

I assume it should be safe to do since it's allowed on OpenShift.io oauth.

tolusha: I am not sure that it is secure, who can confirm that?

sleshchenko: @ibuziuk Could you help to discover the answer here?

sleshchenko (approved):

templates/keycloak_provision
sleshchenko: Seems `_app/oauth.html` needs to be added here as well https://github.com/eclipse/che/blob/master/assembly/assembly-root-war/src/main/webapp/_app/oauth.html.

What if we even add `_app/*` since we are sure that `_app` is an application hosted by Che Server and we don't really need to list them one by one. @skabashnyuk WDYT?

templates/keycloak_provision
```diff
@@ -20,7 +20,7 @@ if [ $? -eq 0 ]; then echo "Realm exists"; exit 0; fi \
                                         -s clientId=$keycloakClientId \
                                         -s id=$keycloakClientId \
                                         -s 'webOrigins=["http://$cheHost", "https://$cheHost"]' \
-                                        -s 'redirectUris=["http://$cheHost/*", "https://$cheHost/*"]' \
+                                        -s 'redirectUris=["http://$cheHost/dashboard/", "https://$cheHost/dashboard/", "http://$cheHost/workspace-loader/", "https://$cheHost/workspace-loader/", "http://$cheHost/_app/*", "https://$cheHost/_app*"]' \
```
sleshchenko, line 23: why `http://_app/` but `https://_app*` (without slash in the end)?

tolusha: Fixed, my mistake

templates/keycloak_provision
```diff
@@ -20,7 +20,7 @@ if [ $? -eq 0 ]; then echo "Realm exists"; exit 0; fi \
                                         -s clientId=$keycloakClientId \
                                         -s id=$keycloakClientId \
                                         -s 'webOrigins=["http://$cheHost", "https://$cheHost"]' \
-                                        -s 'redirectUris=["http://$cheHost/*", "https://$cheHost/*"]' \
+                                        -s 'redirectUris=["http://$cheHost/dashboard/*", "https://$cheHost/dashboard/*", "http://$cheHost/workspace-loader/*", "https://$cheHost/workspace-loader/*", "http://$cheHost/_app/*", "https://$cheHost/_app/*"]' \
```
sleshchenko, line 23: Not really related to the current PR but can we instead of duplication http and https redirects just provide the right protocol via parameter as we provide $cheHost? It would help us be more secure and make sure token if not send through http when https is configured

sleshchenko (approved):
Approving since http/https issue is already in the code but it can be solved separately

templates/keycloak_provision
tolusha: Agree

PR #498, opened 2020-10-17
Add commit new nightly bundle github action.

AndrienkoAleksandr:
## Referenced issue:
https://github.com/eclipse/che/issues/17338

## What does this pull request do:
Add commit new nightly bundle github action: if developer changed deployment yaml files, then github action at the night should commit new bundle and bundle images will be pushed. If there is no changes in the deployment yamls - then nothing to commit.
Fix not needed push bundle images, when we have had already pushed images.

### DIFF
--- a/.github/workflows/pushCatalogImages.yaml
+++ b/.github/workflows/pushCatalogImages.yaml
@@ -19,10 +19,28 @@ jobs:
 
   build:
     name: Build
-    runs-on: ubuntu-latest
-    
+    runs-on: ubuntu-20.04
+
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
+      with:
+        path: go/src/github.com/eclipse/${{ github.repository }}
+        fetch-depth: 0
+
+    - name: Prepare new nightly Olm bundle
+      run: >-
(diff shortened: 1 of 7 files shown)


tolusha (approved):

AndrienkoAleksandr:
[retest]

.github/workflows/pushCatalogImages.yaml
```diff
@@ -19,10 +19,28 @@ jobs:
...
+        fetch-depth: 0
+
+    - name: Prepare new nightly Olm bundle
+      run: >-
+        export GOPATH="${GITHUB_WORKSPACE}/go" &&
+        echo "GOPATH: ${GOPATH}" &&
+        export GITHUB_WORKSPACE="${GOPATH}/src/github.com/eclipse/${{ github.repository }}" &&
+        cd ${GITHUB_WORKSPACE} && "${GITHUB_WORKSPACE}/.github/action_scripts/provideNewNightlyBundle.sh"
+
+    - name: Commit new nightly Olm bundle
```
flacatus, line 37: WDYT about adding a custom message in the commit?

AndrienkoAleksandr: It's a good idea, but I think, we can improve it in the next pr's.

PR #507, opened 2020-10-28
feat: Add ability to use internal network for Keycloak

AndrienkoAleksandr:
### What does this PR do?
Add ability to use internal network for keycloak.

https://github.com/eclipse/che/issues/17644

### DIFF
--- a/pkg/deploy/identity-provider/deployment_keycloak.go
+++ b/pkg/deploy/identity-provider/deployment_keycloak.go
@@ -13,6 +13,7 @@ package identity_provider
 
 import (
 	"context"
+	"net/url"
 	"regexp"
 	"strconv"
 	"strings"
@@ -91,6 +92,10 @@ func getSpecKeycloakDeployment(
 	if cheFlavor == "che" {
 		// writable dir in the upstream Keycloak image
 		jbossDir = "/scripts"
+	}
+	jbossCli := "/opt/jboss/keycloak/bin/jboss-cli.sh"
+	if cheFlavor == "codeready" {
+		jbossCli = "/opt/eap/bin/jboss-cli.sh"
 	}
 
 	if clusterDeployment != nil {
@@ -205,10 +210,8 @@ func getSpecKeycloakDeployment(
 			quotedNoProxy += "\"" + noProxyEntry + ";NO_PROXY\""
 		}
 
-		jbossCli := "/opt/jboss/keycloak/bin/jboss-cli.sh"
 		serverConfig := "standalone.xml"
 		if cheFlavor == "codeready" {
-			jbossCli = "/opt/eap/bin/jboss-cli.sh"
 			serverConfig = "standalone-openshift.xml"
 		}
 		addProxyCliCommand = " && echo Configuring Proxy && " +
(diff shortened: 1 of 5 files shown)


pkg/deploy/identity-provider/deployment_keycloak.go
```diff
@@ -485,8 +485,25 @@ func getSpecKeycloakDeployment(
 		keycloakEnv = append(keycloakEnv, envvar)
 	}
 
-	command := addCertToTrustStoreCommand + addProxyCliCommand + applyProxyCliCommand + " && " + changeConfigCommand +
-		" && /opt/jboss/docker-entrypoint.sh -b 0.0.0.0 -c standalone.xml"
+	var enableFixedHostNameProvider string
+	if deployContext.CheCluster.Spec.Server.UseServiceHostNames {
+		hostname := "keycloak-che.apps-crc.testing"
```
tolusha, line 490: some testing values?

AndrienkoAleksandr: Oh, yes, sorry, I will fix it.

pkg/deploy/identity-provider/identity_provider.go
```diff
@@ -75,6 +76,7 @@ func SyncIdentityProviderToCluster(deployContext *deploy.DeployContext) (bool, e
 		return false, err
 	}
 	keycloakURL := protocol + "://" + endpoint
+	deployContext.InternalService.KeycloakHost = fmt.Sprintf("%s://%s.%s.svc.cluster.local:%d", "http", "keycloak", deployContext.CheCluster.Namespace, 8080)
```
tolusha, line 79: I guess instead of `cluster.local` we have to use `spec.server.serviceHostnameSuffix`

AndrienkoAleksandr: Yes, I will do it after merge https://github.com/eclipse/che-operator/pull/468

AndrienkoAleksandr:
@davidfestal Could you review, please?

pkg/deploy/identity-provider/deployment_keycloak.go
```diff
@@ -485,8 +489,37 @@ func getSpecKeycloakDeployment(
...
+			} else {
+				enableFixedHostNameProvider += "/subsystem=keycloak-server/spi=hostname/provider=fixed:write-attribute(name=properties.httpPort,value=\"80\") \n"
+			}
+			enableFixedHostNameProvider += "stop-embedded-server\" > " + jbossDir  + "/use_fixed_hostname_provider.cli && " +
+			jbossCli + " --file=" + jbossDir + "/use_fixed_hostname_provider.cli "
+		}
+		if cheFlavor == "codeready" {
+			keycloakEnv = append(keycloakEnv, corev1.EnvVar{
+				Name: "KEYCLOAK_FRONTEND_URL",
+				Value: deployContext.CheCluster.Status.KeycloakURL + "/auth",
```
davidfestal, line 516: Isn't it the Keycloak *internal* URL that we want here ?

davidfestal (commented):
I have added 3 comments / questions

Additional remark would be that I don't see any additional test case in order to cover this new option / scenario.

pkg/deploy/identity-provider/deployment_keycloak.go
AndrienkoAleksandr: Keycloak includes `issuer` url inside each token like one of the fields. For our case user dashboard, using keycloak adapters, retrieves tokens from keycloak. And user dashboard uses public host url for this purpose, because user dashboard is frontend and can't use internal network. After that we have decoded 'issuer' field with `public hostname` url inside token. Then che-server needs to reuse this token from user dashboard to make some server side requests(retrieve user profile, link user account, get openshift 4 token, delete user from keycloak and so on...). che-server should use internal url to make requests, but requests with token generated for `public hostname url` fails due keycloak validation. Keycloak compare request url(with internal network value) to token `issuer` and fail che-server request. That's why we set up to keycloak `KEYCLOAK_FRONTEND_URL` env. Keycloak in this mode compare token issuer to the `KEYCLOAK_FRONTEND_URL` and che-server request works fine.

pkg/deploy/identity-provider/identity_provider.go
AndrienkoAleksandr: I used `serviceName.namespace.svc`

deploy/crds/org_v1_che_cr.yaml
```diff
@@ -16,6 +16,7 @@ metadata:
   name: eclipse-che
 spec:
   server:
+    useServiceHostNames: true
```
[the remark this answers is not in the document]
AndrienkoAleksandr: We decided to enable internal network using CR configuration with bool `true`

pkg/apis/org/v1/che_types.go
```diff
@@ -138,6 +138,12 @@ type CheClusterSpecServer struct {
 	// Disabling TLS may cause malfunction of some Che components.
 	// +optional
 	TlsSupport bool `json:"tlsSupport"`
+
+	// Use internal host names provided by cluster dns.
+	// It should be usefull to speed internal services traffic.
+	// Also it could avoid some proxy traffic issues.
+	UseServiceHostNames bool `json:"useServiceHostNames"`
```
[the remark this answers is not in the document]
AndrienkoAleksandr: We decided to enable internal network using CR configuration with bool `true`

AndrienkoAleksandr:
Pr is updated(I applied tests) and re-based.

tolusha (approved):

AndrienkoAleksandr:
"Testing stable versions updates (pull_request)" - launched locally - passed.

PR #509, opened 2020-10-29
Set preserveUnknownFields: false for nighlty bundle

tolusha:

### DIFF
--- a/olm/update-nightly-bundle.sh
+++ b/olm/update-nightly-bundle.sh
@@ -127,7 +127,16 @@ do
     incrementNightlyVersion "${platform}"
   fi
 
-  cp -rf "${ROOT_PROJECT_DIR}/deploy/crds/org_v1_che_crd.yaml" "${bundleFolder}/manifests"
+  templateCRD="${ROOT_PROJECT_DIR}/deploy/crds/org_v1_che_crd.yaml"
+  platformCRD="${bundleFolder}/manifests/org_v1_che_crd.yaml"
+
+  cp -rf $templateCRD $platformCRD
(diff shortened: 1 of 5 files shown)


olm/update-nightly-bundle.sh
```diff
@@ -127,7 +127,16 @@ do
     incrementNightlyVersion "${platform}"
   fi
 
-  cp -rf "${ROOT_PROJECT_DIR}/deploy/crds/org_v1_che_crd.yaml" "${bundleFolder}/manifests"
+  templateCRD="${ROOT_PROJECT_DIR}/deploy/crds/org_v1_che_crd.yaml"
+  platformCRD="${bundleFolder}/manifests/org_v1_che_crd.yaml"
+
+  cp -rf $templateCRD $platformCRD
+  if [[ $platform == "openshift" ]]; then
```
davidfestal, line 134: Not even sure it is *only* for OpenShift. I assume that it would also work on not-too-old versions of Kubernetes with OLM installed.

tolusha: Maybe, but I'm afraid to break something.

davidfestal: > Not even sure it is _only_ for OpenShift. I assume that it would also work on not-too-old versions of Kubernetes with OLM installed.

By the way, adding this also for the Kubernetes case can also be done in a future PR (if we create a new issue to track it). So let me approve this PR to go faster.

davidfestal (approved):

AndrienkoAleksandr (approved):

PR #516, opened 2020-11-03
Fix: add ability to set up namespace using -n argument and CR file using -cr.

AndrienkoAleksandr:
Fix: add ability to set up namespace using -n argument and CR file using -cr.


AndrienkoAleksandr:
In the pr https://github.com/eclipse/che-operator/pull/514 was wrongly used first argument from command line to set up namespace. Wrongly, because this argument was busy by CR file path option.

tolusha (approved):

PR #523, opened 2020-11-05
Revert add bundle commit github action. We can't use it for protected master branch.

AndrienkoAleksandr:


AndrienkoAleksandr:
I talk with @benoitf . `master` branch is protected and github action can't directly push to the master branch without pr's and reviewer's approve. 
**Alternative variant for future**

For automation we could create separated `nightly` branch. Github action could copy deploy folder from `master` branch, generate new bundle using operator-sdk, commit this stuff and push it to 'nightly' branch. Commit to this branch will trigger nightly image push. Also to use this approach we should fix all e2e tests and rework release scripts...

tolusha (approved):

.github/workflows/pushCatalogImages.yaml
```diff
@@ -19,28 +19,10 @@ jobs:
 
   build:
     name: Build
-    runs-on: ubuntu-20.04
-
+    runs-on: ubuntu-latest
```
tolusha, line 22: Should we stick to 20.04 ?

AndrienkoAleksandr: done

PR #526, opened 2020-11-12
Add release workflow

mkuznyetsov:
Use GH Actions to perform release of che-operator

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,6 +14,16 @@ jobs:
     runs-on: ubuntu-20.04
 
     steps:
+      - name: Check release version
+      # we will not allow next major release
+      # until https://github.com/eclipse/che/issues/18349 is fixed,
+      # then this step will be removed
+        run: |
+          CHE_VERSION=${{ github.event.inputs.version }}
+          if [[ ${CHE_VERSION%.*} == 7.22 ]];then
+            echo "Do not release next major version"
+            exit 1
+          fi
       - uses: actions/checkout@v2
 
       - name: Set up Python 3.6
@@ -57,6 +67,7 @@ jobs:
           if [[ ${CHE_VERSION} == *".0" ]]; then
             ./make-release.sh ${CHE_VERSION} --release --release-olm-files --update-nightly-olm-files 
           else
+            git checkout ${BRANCH}
             ./make-release.sh ${CHE_VERSION} --release --release-olm-files
           fi
           
@@ -67,8 +78,8 @@ jobs:
           export QUAY_USERNAME_OS=${{ secrets.QUAY_ECLIPSE_CHE_OPERATOR_OPENSHIFT_USERNAME }}
           export QUAY_USERNAME_OS=${{ secrets.QUAY_ECLIPSE_CHE_OPERATOR_OPENSHIFT_PASSWORD }}
 
-          # git checkout ${CHE_VERSION}
-          # ./make-release.sh ${CHE_VERSION} --push-olm-files
+          git checkout ${CHE_VERSION}-release
+          ./make-release.sh ${CHE_VERSION} --push-olm-files
           
           # perform extra checkouts to ensure branches exist locally
           git checkout ${BRANCH}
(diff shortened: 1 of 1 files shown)


tolusha:
The correct version of the operator-sdk is 0.15.2 right now.
We are working on https://github.com/eclipse/che/issues/18349 to fix some problems with that

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+
+          set -e
+
+          OP_SDK_DIR=/opt/operator-sdk
+          mkdir -p $OP_SDK_DIR
+          wget https://github.com/operator-framework/operator-sdk/releases/download/v0.10.0/operator-sdk-v0.10.0-x86_64-linux-gnu -O $OP_SDK_DIR/operator-sdk 
+          chmod +x $OP_SDK_DIR/operator-sdk
+          export PATH="$PATH:$OP_SDK_DIR"
+
+          # get base32 util from che-release project
```
nickboldt, line 45: do we still need this if we can just install the real base32?

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+          fi
+          
+          # default robot account on quay does not have permissions for application repos
+          # so we provide credentials for accounts that do
+          export QUAY_USERNAME_K8S=${{ secrets.QUAY_ECLIPSE_CHE_OPERATOR_KUBERNETES_PASSWORD }}
+          export QUAY_PASSWORD_K8S${{ secrets.QUAY_ECLIPSE_CHE_OPERATOR_KUBERNETES_PASSWORD }}
+          export QUAY_USERNAME_OS=${{ secrets.QUAY_ECLIPSE_CHE_OPERATOR_OPENSHIFT_USERNAME }}
+          export QUAY_USERNAME_OS=${{ secrets.QUAY_ECLIPSE_CHE_OPERATOR_OPENSHIFT_PASSWORD }}
+
+          # git checkout ${CHE_VERSION}
```
nickboldt, line 72: why are these lines commented out?

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+          # ./make-release.sh ${CHE_VERSION} --push-olm-files
+          
+          # perform extra checkouts to ensure branches exist locally
+          git checkout ${BRANCH}
```
nickboldt, line 76: if branch already exists will this fail? should we have a `|| true` on it?

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+      - uses: actions/checkout@v2
+
+      - name: Set up Python 3.6
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+      - name: Install operator courier
+        run: |
+          python -m pip install --upgrade pip
+          # pip install operator-courier
```
nickboldt, line 26: you're not installing operator-courier here. why commented out?

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Release operator
+        run: |
+          CHE_VERSION=${{ github.event.inputs.version }}
+          BRANCH=${CHE_VERSION%.*}.x
+
+          #preinstall operator-courier and operator-sdk
+          pip install operator-courier==2.1.7
```
nickboldt, line 34: should all the python related installations be done at the same time in the previous step?

tolusha:
We switched to a new operator-sdk version (0.15.2) in master and might switch to even newer in context of https://github.com/eclipse/che/issues/18349

.github/workflows/release.yml
mkuznyetsov: it shouldn't

.github/workflows/release.yml
mkuznyetsov: fixed

.github/workflows/release.yml
mkuznyetsov: I'll use the same for consitency for now

tolusha (approved):

PR #543, opened 2020-11-19
Gh18399 che SA cluster roles

sparkoo:
adds `server.cheClusterRoles` to CheCluster CR and assign these ClusterRoles to `che` ServiceAccount

issue:https://github.com/eclipse/che/issues/18399


pkg/deploy/clusterrolebinding.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if clusterRB == nil {
+		logrus.Infof("Creating a new object: %s, name %s", specCRB.Kind, specCRB.Name)
+		err := deployContext.ClusterAPI.Client.Create(context.TODO(), specCRB)
+		return nil, err
+	}
+
```
tolusha, line 45: Let's check if `clusterRB` and `specCRB` are equal, like here.
https://github.com/eclipse/che-operator/blob/master/pkg/deploy/pvc.go#L71-L80

sparkoo: ok. Should I change return value to ProvisioningStatus?

tolusha: No, that's ok

sparkoo:
@tolusha CI fails to find `namespaces_cluster_role.yaml`
```
[14:28:22] → ENOENT: no such file or directory, open '/Users/runner/Library/Caches/chectl/templates/che-operator/namespaces_cluster_role.yaml'
```
Is it some cache issue? because the file should be there...

tolusha:
Pls ignore `Minishift / Testing stable versions updates (pull_request)`

sparkoo:
> Pls ignore `Minishift / Testing stable versions updates (pull_request)`

ok, I still don't know what to do about `Minishift / Testing latest changes (pull_request)`... :/

tolusha:
Postgres pod never comes up ^( probably due to docker.io pull rate limit

pkg/deploy/clusterrolebinding.go
```diff
@@ -43,6 +45,16 @@ func SyncClusterRoleBindingToCluster(
 		return nil, err
 	}
 
+	diff := cmp.Diff(clusterRB, specCRB, pvcDiffOpts)
+	if len(diff) > 0 {
```
AndrienkoAleksandr, line 49: Sorry, did you tested update flow? I meant migration from one version che to next version che with more permissions 
 verb set. I asked, because I see that for role.go we have ignore fields https://github.com/eclipse/che-operator/blob/2e0bfecd456e20e2c289533fe40f05e5c5d78f3e/pkg/deploy/role.go#L29 and I guess such fields need to be to in the clusterole too. But I'm not sure in 100%. What do you think @tolusha ?

sparkoo: I did not, will try.

tolusha:
@sparkoo 
Could you rebase on master branch pls.
Then I will update csv files.

pkg/deploy/clusterrolebinding.go
tolusha: `pvcDiffOpts` it is just a copy/paste
For `clusterrolebinding` we need a dedicates diffOpts.

sparkoo:
> @sparkoo
> Could you rebase on master branch pls.
> Then I will update csv files.

yes, done

mmorhun (approved):
LGTM, but please check diff/update

tolusha (approved):

pkg/controller/che/che_controller.go
```diff
@@ -580,6 +581,24 @@ func (r *ReconcileChe) Reconcile(request reconcile.Request) (reconcile.Result, e
 		}
 	}
 
+	if len(instance.Spec.Server.CheClusterRoles) > 0 {
+		cheClusterRoles := strings.Split(instance.Spec.Server.CheClusterRoles, ",")
+		for _, cheClusterRole := range cheClusterRoles {
+			cheClusterRole := strings.TrimSpace(cheClusterRole)
+			cheClusterRoleBindingName := "che-" + cheClusterRole
```
tolusha, line 588: Pls use CRB name like this `che-<clusterrole>-<namespace>` then we can distinguish CRB created for different namesaces and would be able to remove on chectl with `server:delete` command.

tolusha (changes requested):
See my comments please

sparkoo:
@tolusha @mmorhun @AndrienkoAleksandr I've fixed the comments. However, I'm not sure how to properly remove the ClusterRoleBindings because their name is based on CheCluster CR. That means once we remove the CheCluster instance, we no longer know the name. The only solution I can think of is to to use labels to find objects to delete when `server:delete` from chectl.  WDYT? Or is there other way?

sparkoo:
@tolusha @mmorhun @AndrienkoAleksandr I've added the namespace label to clusterrolebinding created by operator, so we can now find it with chectl to remove. Are you ok with this? Or any other ideas?

AndrienkoAleksandr:
As far I remember we usually applies 
https://github.com/eclipse/che-operator/blob/master/pkg/deploy/rolebinding.go#L100 and https://github.com/eclipse/che-operator/blob/master/pkg/deploy/role.go#L160 for k8s resource object garbage collection.But to prevent permissions error, you also need apply such permissions to the cluster_role.yaml:

```yaml  
  - apiGroups:
      - org.eclipse.che
    resources:
      - checlusters/finalizers
```

For deploy/role.yaml we have it, but it is hidden, like '*':

```yaml
- apiGroups:
  - org.eclipse.che
  resources:
  - '*'
  verbs:
  - '*'
```

sparkoo:
@AndrienkoAleksandr yes, I wanted to do the same for clusterrolebinding, but it is not possible. I think this works only for namespaced objects.

AndrienkoAleksandr:
> @AndrienkoAleksandr yes, I wanted to do the same for clusterrolebinding, but it is not possible. I think this works only for namespaced objects.

In my own opinion: Object created from operator code should be garbage collected using `SetControllerReference`. Objects created from chectl should be removed from chectl code. SetControllerReference should fail for cluster objects, if you don't apply permission described above.

sparkoo:
I agree that operator (and every component) should be responsible for objects it creates. I'll try to update the permissions, but I don't think it will help.

https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents
> Cross-namespace owner references are disallowed by design. This means:
> 1. Namespace-scoped dependents can only specify owners in the same namespace, and owners that are cluster-scoped.
> 1. Cluster-scoped dependents can only specify cluster-scoped owners, but not namespace-scoped owners.

sparkoo:
@AndrienkoAleksandr this is an error I'm getting
```time="2020-12-01T08:57:29Z" level=error msg="cluster-scoped resource must not have a namespace-scoped owner, owner's namespace che"
time="2020-12-01T09:08:25Z" level=info msg="Waiting on cluster role binding 'che-che-che-che-namespace-editor' to be created"
```

sparkoo:
@AndrienkoAleksandr I've looked at some other examples in che-operator. I've found `consolelink` and `oauthclient` objects that are cluster-scoped and neither of them is deleted on `server:delete`! So I guess we need to have some generic solution for cluster-scoped objects garbage collection. Or I can hack in some solution just for this clusterrolebinding, but that would not fix the issue as whole. I'm more inclining to leave it as is and create an issue for garbage collection. WDYT?

tolusha (approved):

pkg/controller/che/che_controller.go
sparkoo: done

pkg/deploy/clusterrolebinding.go
sparkoo: done

sparkoo:
@tolusha @AndrienkoAleksandr I've addressed all the comments. I've created separate issue for garbage collection as it is more generic problem https://github.com/eclipse/che/issues/18501
When CI pass, can I merge? Or do we need to do some more regenerate magic here? :)

sparkoo:
@tolusha do you know what is failing in CI ? is it dockerhub limits again?

tolusha:
I am checking...

sparkoo:
if CI tests are using chectl deploy, it won't work without https://github.com/che-incubator/chectl/pull/1003, because chectl is renaming the che-namespace-editor clusterrole

tolusha:
@sparkoo @flacatus 
Yes, we are checking manually
Until https://github.com/eclipse/che-operator/pull/557 is merged

flacatus:
@sparkoo I was able to tests your changes in my local machine and works fine in minikube. There is a problem with actions somewhere is creaeting 2 separate catalog sources

flacatus:
@sparkoo Tested in my local machine using minishift!

PR #550, opened 2020-11-23
Refresh oauth client

metlos:
When changing the `serverExposureStrategy` between `multi-host` and `single-host`, the URL of the keycloak server changes. This must be reflected in the OAuthClient object on OpenShift when using OpenShift OAuth otherwise OpenShift won't allow the users to login.

This fixes eclipse/che#17886.

### DIFF
--- a/pkg/deploy/identity-provider/identity_provider.go
+++ b/pkg/deploy/identity-provider/identity_provider.go
@@ -165,7 +165,7 @@ func CreateIdentityProviderItems(deployContext *deploy.DeployContext, cheFlavor 
 	keycloakURL := instance.Spec.Auth.IdentityProviderURL
 	keycloakRealm := util.GetValue(instance.Spec.Auth.IdentityProviderRealm, cheFlavor)
 	oAuthClient := deploy.NewOAuthClient(oAuthClientName, oauthSecret, keycloakURL, keycloakRealm, isOpenShift4)
-	if err := deploy.Sync(deployContext, oAuthClient, oAuthClientDiffOpts); err != nil {
+	if _, err := deploy.Sync(deployContext, oAuthClient, oAuthClientDiffOpts); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/deploy/identity-provider/identity_provider.go
```diff
@@ -155,16 +159,17 @@ func CreateIdentityProviderItems(deployContext *deploy.DeployContext, cheFlavor
...
 	keycloakURL := instance.Spec.Auth.IdentityProviderURL
 	keycloakRealm := util.GetValue(instance.Spec.Auth.IdentityProviderRealm, cheFlavor)
 	oAuthClient := deploy.NewOAuthClient(oAuthClientName, oauthSecret, keycloakURL, keycloakRealm, isOpenShift4)
-	if err := createNewOauthClient(deployContext, oAuthClient); err != nil {
+	if err := deploy.Sync(deployContext, oAuthClient, oAuthClientDiffOpts); err != nil {
 		return err
 	}
 
-	if !tests {
+	if !tests && initializeKeycloak {
```
tolusha, line 172: Is it better to reuse `instance.Status.OpenShiftoAuthProvisioned` field here ?

pkg/deploy/identity-provider/identity_provider.go
```diff
@@ -123,11 +128,8 @@ func SyncIdentityProviderToCluster(deployContext *deploy.DeployContext) (bool, e
 	if isOpenShift {
 		doInstallOpenShiftoAuthProvider := instance.Spec.Auth.OpenShiftoAuth
 		if doInstallOpenShiftoAuthProvider {
-			openShiftIdentityProviderStatus := instance.Status.OpenShiftoAuthProvisioned
-			if !openShiftIdentityProviderStatus {
-				if err := CreateIdentityProviderItems(deployContext, cheFlavor); err != nil {
-					return false, err
-				}
+			if err := CreateIdentityProviderItems(deployContext, cheFlavor); err != nil {
```
tolusha, line 131: pls, rename it to `Sync..` otherwise it looks like we create items on every reconcile loop

tolusha (approved):

tolusha (changes requested):
pls. see my comments

tolusha (approved):

PR #560, opened 2020-12-02
Import all certificates from propagated bundles into Keycloak's truststore

mmorhun:

### What does this PR do?
Makes it possible to import several CA certificates from a single file into Keycloak's java trust store.

https://github.com/eclipse/che/issues/18339

### How to test this PR?
1. Create a *.pem file that contains several root CA certificates
2. Create a configmap with single data field which contains the *.pem file
3. Add the configmap as trusted for Che
4. Check that all CA certs are added into Keycloak's trust store

### DIFF
--- a/pkg/deploy/identity-provider/deployment_keycloak.go
+++ b/pkg/deploy/identity-provider/deployment_keycloak.go
@@ -111,6 +111,9 @@ func getSpecKeycloakDeployment(
 	cheCertSecretVersion := getSecretResourceVersion("self-signed-certificate", deployContext.CheCluster.Namespace, deployContext.ClusterAPI)
 	openshiftApiCertSecretVersion := getSecretResourceVersion("openshift-api-crt", deployContext.CheCluster.Namespace, deployContext.ClusterAPI)
 
+	// holds bash functions that should be available when run init commands in shell
+	bashFunctions := ""
+
 	// add various certificates to Java trust store so that Keycloak can connect to OpenShift API
 	// certificate that OpenShift router uses (for 4.0 only)
 	addRouterCrt := "if [ ! -z \"${CHE_SELF__SIGNED__CERT}\" ]; then echo \"${CHE_SELF__SIGNED__CERT}\" > " + jbossDir + "/che.crt && " +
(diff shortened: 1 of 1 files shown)


pkg/deploy/identity-provider/deployment_keycloak.go
```diff
@@ -714,3 +718,51 @@ func ProvisionKeycloakResources(deployContext *deploy.DeployContext) error {
 	_, err = util.K8sclient.ExecIntoPod(podToExec, keycloakProvisionCommand, "create realm, client and user", deployContext.CheCluster.Namespace)
 	return err
 }
+
+// getImportCABundleScript returns string which cntains bash function that imports ca-bundle into jks
+// The function has three arguments:
+// - absolute path to ca-bundle file
+// - absolute path to java keystore
+// - java keystore password
+func getImportCABundleScript() string {
```
davidfestal, line 727: How do you manage the case where there are duplicate certificates (I assume we should be robust and ignore this error) ?
What about other errors when importing a certificate, Do we expect them to fail the start of the server ?

davidfestal (commented):
What about the Che server ? Don't we add certificates in the java truststore in the Che Java deployment as well ? Does it need the same fix ?

pkg/deploy/identity-provider/deployment_keycloak.go
mmorhun: > How do you manage the case where there are duplicate certificates (I assume we should be robust and ignore this error) ?

When adding a certificate into trust store we pass unique alias for it. So if user provides identical certificates - both will be added to trust store without any error.

> What about other errors when importing a certificate, Do we expect them to fail the start of the server ?

Invalid certificates will not be imported, Keycloak and Che server will be started without invalid certificates

mmorhun:
> What about the Che server ? Don't we add certificates in the java truststore in the Che Java deployment as well ? Does it need the same fix ?

@davidfestal already there: https://github.com/eclipse/che/pull/18504

pkg/deploy/identity-provider/deployment_keycloak.go
davidfestal: Thanks for the explanation.

davidfestal:
> > What about the Che server ? Don't we add certificates in the java truststore in the Che Java deployment as well ? Does it need the same fix ?
> 
> @davidfestal already there: [eclipse/che#18504](https://github.com/eclipse/che/pull/18504)

Great ! Thanks !

mmorhun:
@davidfestal do you have any other remarks?

pkg/deploy/identity-provider/deployment_keycloak.go
```diff
@@ -714,3 +718,51 @@ func ProvisionKeycloakResources(deployContext *deploy.DeployContext) error {
 	_, err = util.K8sclient.ExecIntoPod(podToExec, keycloakProvisionCommand, "create realm, client and user", deployContext.CheCluster.Namespace)
 	return err
 }
+
+// getImportCABundleScript returns string which cntains bash function that imports ca-bundle into jks
```
AndrienkoAleksandr, line 722: cntains > contains

AndrienkoAleksandr (approved):

PR #569, opened 2020-12-07
Replace postgres image

tolusha:

### What does this PR do
Used images that moved from docker hub

### Reference issue
https://github.com/eclipse/che/issues/18292

### DIFF
--- a/deploy/operator-local.yaml
+++ b/deploy/operator-local.yaml
@@ -56,7 +56,7 @@ spec:
             - name: RELATED_IMAGE_pvc_jobs
               value: registry.access.redhat.com/ubi8-minimal:8.3-201
             - name: RELATED_IMAGE_postgres
-              value: registry.centos.org/centos/postgresql-96-centos7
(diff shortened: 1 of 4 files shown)


skabashnyuk (approved):

deploy/operator-local.yaml
```diff
@@ -56,7 +56,7 @@ spec:
             - name: RELATED_IMAGE_pvc_jobs
               value: registry.access.redhat.com/ubi8-minimal:8.3-201
             - name: RELATED_IMAGE_postgres
-              value: centos/postgresql-96-centos7:9.6
+              value: registry.centos.org/centos/postgresql-96-centos7
```
nickboldt, line 59: call me crazy but don't we want to use a tag here instead of :latest ?I know the project is fast-and-loose and bleeding-edge, but do you want unexpected changes here?

tolusha: Agree

nickboldt (approved):

PR #588, opened 2020-12-21
fix: Improve openshift auto detection and related tests

AndrienkoAleksandr:
### What does this pr do:
Improve openshift version auto detection and related tests

### Tests
Tested installation on crc, minishfit, minikube using chectl.


tolusha:
Good refactoring. Can we do more:
1. Invoke `util.DetectOpenShift(discoveryClient)` only once in `main` ?
2. Not to pass `util.IsOpenShift()` and `util.IsOpenShift4()` in any function ?

PR #596, opened 2020-12-23
Add ability to select platforms for community operator update PR

mkuznyetsov:
Signed-off-by: Mykhailo Kuznietsov <mkuznets@redhat.com>

When generating the community operator PR update, you can now specify platforms for which to run it, which makes it possible to run just for either kubernetes, or openshift (by default it does for both platforms). Should be useful, if it needs to rerun them separately

### DIFF
--- a/.github/workflows/release-community-operator-PRs.yml
+++ b/.github/workflows/release-community-operator-PRs.yml
@@ -7,6 +7,9 @@ on:
       forceflag:
         description: 'To update an existing PR, use -f or --force flag here'
         default: ''
+      platforms:
+        description: 'List platforms for which to create PRs, which could be \'openshift\', \'kubernetes\' or both
+        default: '\'kubernetes\' \'openshift\''  
   # trigger on commit to master branch of new CSVs, eg., https://github.com/eclipse/che-operator/pull/571/files
   push:
     branches: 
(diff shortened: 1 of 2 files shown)


tolusha (approved):

olm/prepare-community-operators-update.sh
```diff
@@ -14,6 +14,7 @@ set -e
 
 CURRENT_DIR=$(pwd)
 BASE_DIR=$(cd "$(dirname "$0")"; pwd)
+PLATFORMS="'kubernetes' 'openshift'"
```
nickboldt, line 17: why do we need single-quotes inside double quotes?

.github/workflows/release-community-operator-PRs.yml
```diff
@@ -7,6 +7,9 @@ on:
       forceflag:
         description: 'To update an existing PR, use -f or --force flag here'
         default: ''
+      platforms:
+        description: 'List platforms for which to create PRs, which could be \'openshift\', \'kubernetes\' or both
+        default: '\'kubernetes\' \'openshift\''
```
nickboldt, line 12: if a comma-separated list, it would be easier to pass this in as:
 * kubernetes,openshift
 * kubernetes
 * openshift

olm/prepare-community-operators-update.sh
```diff
@@ -49,7 +51,7 @@ Options:
 "
 }
 
-for platform in 'kubernetes' 'openshift'
+for platform in $PLATFORMS
```
nickboldt, line 54: if a comma-separated list, would need to simply transform with `tr`:

```
$➔ for platform in abc,def; do echo $platform; done
abc,def
```

```
$➔ for platform in $(echo "abc,def"|tr "," " "); do echo $platform; done
abc
def
```
Thus:

```
for platform in $(echo $PLATFORMS | tr "," " ")
```

nickboldt (changes requested):
please switch to comma-separated instead of space-with-single-quotes separated

PR #598, opened 2020-12-29
Add liveness and readiness probes

tolusha:

### Reference issue
https://issues.redhat.com/browse/CRW-683
https://issues.redhat.com/browse/CRW-916

### What does this PR do
- add liveness and readiness probes for operator
- add liveness probe for keycloak and postgres containers

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -35,6 +35,7 @@ import (
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 )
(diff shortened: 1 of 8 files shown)


cmd/manager/main.go
```diff
@@ -155,6 +160,16 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Setup health checks
+	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+		log.Error(err, "Unable to set up health check")
+		os.Exit(1)
+	}
+	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
```
sleshchenko, line 168: I understand why Keycloak and Postgres, registries have readiness probe

> Readiness
> Readiness probes are designed to let Kubernetes know when your app is ready to serve traffic. Kubernetes makes sure the readiness probe passes before allowing a service to send traffic to the pod. If a readiness probe starts to fail, Kubernetes stops sending traffic to the pod until it passes.

But as far as I understand Che Operator does not provide any REST API.
So, I'm missing some point or the readiness probe is redundant.

The readiness probe won't affect metrics, right https://github.com/eclipse/che-operator/blob/probes/deploy/olm-catalog/eclipse-che-preview-kubernetes/manifests/che-operator.clusterserviceversion.yaml#L372 ?

tolusha: > But as far as I understand Che Operator does not provide any REST API.

`mgr.AddReadyzCheck("readyz", healthz.Ping)` does the trick.

> The readiness probe won't affect metrics, right

It shouldn't. Do you have any concerns ?

tolusha:
OpenShift tests fail due to fact that it tries to use the `quay.io/eclipse/che-operator:nightly` image where readiness and liveness probes aren't added.

tolusha:
Tested locally on OpenShift, it works fine.

mmorhun (approved):

sleshchenko (approved):

PR #607, opened 2021-01-05
Set Recreate strategy type for operator deployment

tolusha:

### What does this PR do
Set `Recreate` strategy type for operator deployment.
Operator update fails after adding liveness and readiness probe.


sleshchenko (approved):
It would be cool to understand if it's possible to do a rolling update to avoid downtime while the cluster is pulling the image
But it makes sense to merge this PR to fix the described issues

nickboldt (approved):
+0, no idea what this change does or why it's helpful.

Also "Operator update fails after adding liveness and readiness probe" sounds like a problem ... is this attempting to solve that? You had the same comment on description of https://github.com/eclipse/che-operator/pull/608 too

PR #609, opened 2021-01-06
Revert recreate strategy and probes for operator

tolusha:

### What does this pr do
Revert recreate strategy and probes for operator

### DIFF
--- a/deploy/olm-catalog/eclipse-che-preview-kubernetes/manifests/che-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/eclipse-che-preview-kubernetes/manifests/che-operator.clusterserviceversion.yaml
@@ -84,13 +84,13 @@ metadata:
     categories: Developer Tools
     certified: "false"
     containerImage: quay.io/eclipse/che-operator:nightly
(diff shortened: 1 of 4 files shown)


deploy/olm-catalog/eclipse-che-preview-kubernetes/manifests/che-operator.clusterserviceversion.yaml
```diff
@@ -291,8 +291,7 @@ spec:
             selector:
               matchLabels:
                 app: che-operator
-            strategy:
-              type: Recreate
+            strategy: {}
```
sleshchenko, line 294: Maybe instead of default one better to explicitly set which we expect, like Rolling?

sleshchenko (approved):

deploy/olm-catalog/eclipse-che-preview-kubernetes/manifests/che-operator.clusterserviceversion.yaml
tolusha: I've decided just to rollback. 
Once I configure readiness and liveness probe I will set the strategy explicitly.

PR #843, opened 2021-05-24
chore: Define CheCluster v2alpha1

metlos:
### What does this PR do?
This PR adds types for `CheCluster` `v2alpha1` that should be used by devworkspace che operator (DWCO) in the near future. The new `CheCluster` type is completely convertible from `CheCluster` `v1` and there are conversion functions defined between the two versions.

The `v2alpha1` version IS NOT in any way exposed to the cluster. For the time being, this type is only intended to be used in-memory at runtime and cannot be persisted to the cluster (and is not part of the CRD). 

eclipse/che#19846

### How to test this PR?
This actually is not connected to any existing code (the only visible manifestation of this code is a new optional `devworkspace` section in the `CheCluster` status. To test this, just review and run the unit tests (`conversion_test.go`).

### PR Checklist

### Reviewers


pkg/apis/org/conversion.go
```diff
@@ -0,0 +1,193 @@
...
+	if util.IsOpenShift {
+		v2.Spec.WorkspaceBaseDomain = v1.Spec.Server.CustomCheProperties[routeDomainSuffixPropertyKey]
+	} else {
+		v2.Spec.WorkspaceBaseDomain = v1.Spec.K8s.IngressDomain
+	}
+}
+
+func v1ToV2alpha1_GatewayEnabled(v1 *v1.CheCluster, v2 *v2alpha1.CheCluster) {
+	val := util.GetServerExposureStrategy(v1) == "single-host"
+	v2.Spec.Gateway.Enabled = &val
```
tolusha, line 97: What if  `spec.k8s.singleHostExposureType: native` ?

pkg/apis/org/conversion.go
```diff
@@ -0,0 +1,193 @@
...
+
+func v1ToV2alpha1_GatewayImage(v1 *v1.CheCluster, v2 *v2alpha1.CheCluster) {
+	v2.Spec.Gateway.Image = v1.Spec.Server.SingleHostGatewayImage
+}
+
+func v1ToV2alpha1_GatewayConfigurerImage(v1 *v1.CheCluster, v2 *v2alpha1.CheCluster) {
+	v2.Spec.Gateway.ConfigurerImage = v1.Spec.Server.SingleHostGatewayConfigSidecarImage
+}
+
+func v1ToV2alpha1_TlsSecretName(v1 *v1.CheCluster, v2 *v2alpha1.CheCluster) {
```
tolusha, line 108: How it works know, for instance: 
* OpenShift domain is `my.openshift.com`
* User is user is able to specify `spec.server.cheHost: my-ide.com` (something different from OpenShift domain)
* `spec.server.cheHostTLSSecret` is used to secure that domain.

So, how it will work after changes? Will user able to set and secure any Che host we wants?

pkg/apis/org/conversion.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func v1ToV2alpha1_K8sIngressAnnotations(v1 *v1.CheCluster, v2 *v2alpha1.CheCluster) {
+	annotations := map[string]string{
+		"kubernetes.io/ingress.class":                       v1.Spec.K8s.IngressClass,
+		"nginx.ingress.kubernetes.io/proxy-read-timeout":    "3600",
+		"nginx.ingress.kubernetes.io/proxy-connect-timeout": "3600",
+		"nginx.ingress.kubernetes.io/ssl-redirect":          "true",
+	}
+	// This is what is applied in the deploy/ingress.go but I don't think it is applicable in our situation
```
tolusha, line 127: To be honest I don't know either. 
If registries will be accessible without that annotation then it is fine for me.

pkg/apis/org/conversion.go
```diff
@@ -0,0 +1,193 @@
...
+		if v1.Spec.Server.CustomCheProperties == nil {
+			v1.Spec.Server.CustomCheProperties = map[string]string{}
+		}
+		v1.Spec.Server.CustomCheProperties[routeDomainSuffixPropertyKey] = v2.Spec.WorkspaceBaseDomain
+	} else {
+		v1.Spec.K8s.IngressDomain = v2.Spec.WorkspaceBaseDomain
+	}
+}
+
+func v2alpha1ToV1_GatewayEnabled(v1 *v1.CheCluster, v2 *v2alpha1.CheCluster) {
```
tolusha, line 150: We don't take `spec.k8s.singleHostExposureType` into account here.

pkg/apis/org/conversion.go
```diff
@@ -0,0 +1,193 @@
+package org
+
+import (
+	v1 "github.com/eclipse-che/che-operator/pkg/apis/org/v1"
+	"github.com/eclipse-che/che-operator/pkg/apis/org/v2alpha1"
+	"github.com/eclipse-che/che-operator/pkg/util"
+	"gopkg.in/yaml.v2"
```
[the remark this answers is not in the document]
metlos: Yep, seems like VS code was trying to be helpful somehow.. I remember the same thing - I used `sigs.k8s.io/yaml` but it got lost somehow. Good catch.

pkg/apis/org/conversion.go
metlos: To be fair, the situation is not that bad in DWCO. DWCO only deals with workspace endpoints, so for example the dashboard URL is not its concern. But still we need to support `domain/subpath` endpoints as well as `subdomain.domain`. To be able to use a different certificate for each, I added the capability to the model.

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	k8s.io/apiextensions-apiserver v0.18.2
 	k8s.io/apimachinery v0.18.2
 	k8s.io/client-go v12.0.0+incompatible
+	k8s.io/utils v0.0.0-20191010214722-8d271d903fe4
```
tolusha, line 20: Just for information
It is ok to use: 
https://clearlydefined.io/definitions/git/github/kubernetes/utils/8d271d903fe4c290aa361acfb242cff7bcee96f1

pkg/apis/org/v2alpha1/checluster.go
```diff
@@ -0,0 +1,203 @@
...
+	GatewayPhase GatewayPhase `json:"gatewayPhase,omitempty"`
+
+	// GatewayHost is the resolved host of the ingress/route. This is equal to the Host in the spec
+	// on Kubernetes but contains the actual host name of the route if Host is unspecified on OpenShift.
+	GatewayHost string `json:"gatewayHost,omitempty"`
+
+	// Phase is the phase in which the Che cluster as a whole finds itself in.
+	Phase ClusterPhase `json:"phase,omitempty"`
+
+	// A brief CamelCase message indicating details about why the pod is in this state.
```
tolusha, line 157: Is it really about the pod?

tolusha (approved):

deploy/crds/org_v1_che_crd-v1beta1.yaml
```diff
@@ -868,6 +868,38 @@ spec:
             devfileRegistryURL:
               description: Public URL to the devfile registry.
               type: string
+            devworkspaceStatus:
```
sparkoo, line 871: Can you please explain to me why it's named `devworkspaceStatus` and not just `devworkspace`? I don't understand the `status` part.

sparkoo:
what is purpose of v1beta1 crd?

sparkoo (commented):
I have no comments to the code. However, this whole conversion and pseudo-crd feels wrong. Why we cannot just add necessary configuration to CheCluster v1 and use it in dwco?

tolusha:
@sparkoo 
v1beta1 crd is used only for OCP 3.11

sparkoo:
> @sparkoo
> v1beta1 crd is used only for OCP 3.11

So do we need devworkspace changes there? Do we support devworkspaces in 3.11?

tolusha:
As far as I know DevWorkspace doesn't support OCP 3.11

metlos:
> I have no comments to the code. However, this whole conversion and pseudo-crd feels wrong. Why we cannot just add necessary configuration to CheCluster v1 and use it in dwco?

Yes, that is a valid concern. There are a couple of reasons why I chose this approach:
* duplicity of information - most of the configuration of v2alpha1 can be deduced from the configuration of v1. It would feel wrong to configure the "same" thing twice on 2 different places in the codebase.
* future-proofing - IMHO we will want to have a new version of CheCluster CR usable in the cluster at some point. If for nothing else then for dropping the support of v1 at some point in time. The conversion code is there to plug into the versioning machinery of Kubernetes - we will need to support coexistence of the 2 versions of the CR for a limited period of time. And this is what we're actually doing with this code already, only in memory. This should make us ready for the actual use of the v2 in the cluster - we will have ironed out the bugs by the time we start exposing the new CR version to the cluster users, hopefully.

deploy/crds/org_v1_che_crd-v1beta1.yaml
metlos: It's the status of the devworkspace. I guess `devworkspace` makes sense, too.

sparkoo:
>     * duplicity of information - most of the configuration of v2alpha1 can be deduced from the configuration of v1. It would feel wrong to configure the "same" thing twice on 2 different places in the configuration.

Why duplicate? Cannot we just use the value that's already in place ?

> 
>     * future-proofing - IMHO we will want to have a new version of CheCluster CR usable in the cluster at some point. If for nothing else then for dropping the support of v1 at some point in time. The conversion code is there to plug into the versioning machinery of Kubernetes - we will need to support coexistence of the 2 versions of the CR for a limited period of time. And this is what we're actually doing with this code already, only in memory. This should make us ready for the actual use of the v2 in the cluster - we will have ironed out the bugs by the time we start exposing the new CR version to the cluster users, hopefully.

I understand this and I hope you're right. However, I still think that safer and faster way would be to do what we need to do now and keep the v1->v2 when we really need it (deprecating v1). With this, we're predicting the future, which will be different and we will need to do some refactoring. In other words, this approach might be better. But given the odds my bet would be we won't save any dev-time in the future, and maybe we will be even slower because we guessed something wrong now.

I won't block the PR, just sharing my concerns.

metlos:
> I understand this and I hope you're right. However, I still think that safer and faster way would be to do what we need to do now and keep the v1->v2 when we really need it (deprecating v1). With this, we're predicting the future, which will be different and we will need to do some refactoring. In other words, this approach might be better. But given the odds my bet would be we won't save any dev-time in the future, and maybe we will be even slower because we guessed something wrong now.
>
> I won't block the PR, I'm just sharing my concerns.

To a large degree, I share your concerns. I didn't want to abandon the structure of `CheManager` from devworkspace-che (from which `CheCluster` v2 is very much derived), because I think it expresses the current thinking about configuring devworkspaces/che much better than the old structure. And that's mostly because our thinking has evolved quite a lot, not because `CheCluster` was bad.

Given that premise, I think keeping the conversion code concentrated in one place and letting the che-operator and devworkspace-che work in their "native" way of thinking is going to prove itself worthy going forward. But as you say, that is debatable to a degree and I am not 100% sure we won't hit a wall myself. Currently, I don't see it.

PR #844, opened 2021-05-24
feat: Implement backup / restore of Che server

mmorhun:
#### Short description
This PR implements ability to backup Eclipse Che installation, send data to a backup server and then restore Che from it. It doesn't handle content of user's projects, but backups configuration of user workspaces.
As of now, backup is bound to specific cluster and cannot be restored on another one.

#### Full description

This PR introduces two new CRDs:
- `CheClusterBackup`
- `CheClusterRestore`

There is a separate operator controller for backup as well as for restore CR, so they don't affect main operator functionality and can run in parallel.

Backup controller, when requested, gathers the following information about Eclipse Che installation:
- Che CR
- Dumps Che and Keycloaks databases
- custom CA certificates
- credentials secrets (ChePostgresSecret, IdentityProviderSecret, IdentityProviderPostgresSecret)
- Some cluster and Che info

and then encrypts and sends it to the configured backup server.

Also, it is possible to request (by setting `useInternalBackupServer` in backup CR) from Che operator (backup controller) to deploy internal backup server. It will be deployed in the same namespace as Che server and might be used only to rollback Che installation to the previous state (even if Che gets deleted, however backup server must survive). The deployment is a single container with restic REST backup server inside. Note, this server configuration doesn't have any protection settings (beside data encryption) and designed to be used only inside the cluster.
For convenience, restore CR has property `copyBackupServerConfiguration` that gets backup server config from backup CR in the same namespace.

Restore controller, on user request, restores Che. It might restore Che independently from the fact of existing of previous installation. Restore controller works in the following way:
- Cleans previous installation, if any: deletes Che CR, secrets, config maps, custom CAs, etc.
- Restores Che resources: secrets, custom CAs and finally Che CR. This will trigger deploy process in main operator controller which will pick up all the restored from the backup resources. Restore controller waits until deployment of clean Che (but with configuration from the backup) is finished.
- Restores Che and Keycloak databases from the backuped dumps and patches some values that are bind to specific cluster.
- Restarts Keycloak to refresh its caches after database restore.

<del> On Kubernetes, it is possible to create Che backup on one cluster and restore it on another. However, it requires to add new ingress domain in restore configuration as it cannot be autodetected on Kubernetes.
In case of Openshift it is not possible to restore on different cluster, but still possible to restore into a different namespace.</del>
_Edit: After some discussions and reviews I dropped support for restoring on different cluster._
This is caused by Keycloak, that binds to OAuth user IDs in Openshift cluster. It was decided not to spent time on patching a lot of related values in Keycloak database as Keycloak will be removed in a mid/short term.

To create backup snapshots, encrypt/decrypt them and push/pull to/from a backup server, [third party tool](https://github.com/restic/restic) called `restic` is used. It provides encryption, versioning of snapshots (so, it is possible to restore not from the latest backup by setting `snapshotId` in restore CR, however by default the latest one is used), and support for different backends. In this PR, only 3 types of external backup servers are added:
 - REST
 - SFTP (via SSH)
 - AWS S3 (and compatible servers like minio, etc.)

Backup servers configuration are the same for backup and restore CRs. Under the hood, they use the same structures in source code. For backup/restore passwords a field in a CR or a separate secret might be used.

Both backup and restore CRs have property `triggerNow` that starts the process of backup / restore correspondingly. This is done to be able to request backups without recreating of the whole CR.

#### Known bugs

- Inability to restore on a different Openshift cluster (won't fix)
- Custom CA certificates are not used when connect to a backup server (will be implemented in a separate PR)
- SFTP client doesn't check fingerprint of configured backup server
- Sometimes, after restore, it is needed to reopen private browser window in order to login into restored Che.

N/A

https://github.com/eclipse/che/issues/18703

### How to test this PR?

In order to run operator with changes from this PR the following steps are required:
1. Checkout to the PR branch
2. There are 2 ways how to apply templates (use only one: a *or* b):
2.a. Use templates from this branch, when deploying Che
2.b. Manually apply resources:
```shell
oc apply -f deploy/role.yaml
oc apply -f deploy/crds/org.eclipse.che_checlusterbackups_crd.yaml
oc apply -f deploy/crds/org.eclipse.che_checlusterrestores_crd.yaml
```
and also manually apply `deploy/operator.yaml` from the PR branch to create operator deployment.
3. Deploy Che using operator image built from the PR branch:
```shell
chectl server:deploy --platform=**** --installer=operator --che-operator-image=mm4eche/che-operator:18703
```

#### Testing with internal backup server

1. Deploy Eclipse Che
2. Make some changes in its configuration: change Che CR, add custom CA certs, create user workspaces, etc.
3. Request backup by creating CR of `CheClusterBackup` type:
```yaml
kind: CheClusterBackup
metadata:
  name: eclipse-che-backup
spec:
  triggerNow: true
  useInternalBackupServer: true
```
4. Wait until backup CR status says that backup successfully finished
5. Change installation: edit Che CR, create/delete workspaces, etc.
6. Request restore by creating CR of `CheClusterRestore` type:
```yaml
kind: CheClusterRestore
metadata:
  name: eclipse-che-restore
spec:
  copyBackupServerConfiguration: true
```
7. Wait until restore CR status shows that restore successfully finished.
8. Check installation, that everything is in the state on the moment of backup.

#### Testing with an external backup server

In this test scenario we'll use REST backup server, but any other supported backup server might be used. Just setup the server and provide correct configuration (if not, Che operator will tell about occurred error in backup CR status).

1. Setup [REST backup server](https://github.com/restic/rest-server) to be available from within the cluster.
2. Install and configure Eclipse Che, amek some configuration changes, create user workspaces, etc.
3. Request backup by creating backup CR (adjust values according to your settings):
```yaml
kind: CheClusterBackup
metadata:
  name: eclipse-che-backup
spec:
  triggerNow: true
  servers:
    rest:
      hostname: mybackup-rest-server.host.net
      protocol: http
      port: 8000
      repositoryPath: che
      repositoryPasswordSecretRef: secretName
```
4. Wait until backup CR status says that backup successfully finished
5. Make changes to installation
6. Request restore by creating restore CR (adjust values according to backup server configuration):
```yaml
kind: CheClusterRestore
metadata:
  name: eclipse-che-restore
spec:
  servers:
    rest:
      hostname: 192.168.50.191
      protocol: http
      port: 8000
      repositoryPath: che
      username: cheuser
      password: chebackup
      repositoryPasswordSecretRef: secretName
```
or:
```yaml
kind: CheClusterRestore
metadata:
  name: eclipse-che-restore
spec:
  copyBackupServerConfiguration: true
```
7. Wait until restore finishes
8. Check that Eclipse Che is in state as on the moment of creating the backup

### PR Checklist

### Reviewers


mmorhun:
Rebased

Dockerfile
```diff
@@ -14,6 +14,12 @@ FROM registry.access.redhat.com/ubi8-minimal:8.4-200 as builder
 RUN microdnf install -y golang unzip && \
     go version
 
+# get restic. Needed for backup / restore capabilities
+ENV RESTIC_VERSION=0.12.0
+RUN microdnf install -y bzip2 && \
+    curl -sLO https://github.com/restic/restic/releases/download/v${RESTIC_VERSION}/restic_${RESTIC_VERSION}_linux_amd64.bz2 && \
+    bzip2 -d restic_${RESTIC_VERSION}_linux_amd64.bz2 && mv restic_* /tmp/restic && chmod +x /tmp/restic
```
tolusha, line 21: Should we care about architecture ?
Does licence allow to include resting in an image?

deploy/crds/org.eclipse.che_checlusterbackups_crd.yaml
```diff
@@ -0,0 +1,214 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
tolusha, line 1: must be v1

deploy/crds/org.eclipse.che_checlusterrestores_crd.yaml
```diff
@@ -0,0 +1,230 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
tolusha, line 1: must be v1

deploy/operator.yaml
```diff
@@ -79,6 +79,8 @@ spec:
               value: quay.io/che-incubator/devworkspace-che-operator:ci
             - name: RELATED_IMAGE_devworkspace_controller
               value: quay.io/devfile/devworkspace-controller:next
+            - name: RELATED_IMAGE_internal_rest_backup_server
+              value: quay.io/eclipse/che-backup-server-rest:eeacd92
```
tolusha, line 83: We have to contact Nick how to productize this image.

deploy/role.yaml
```diff
@@ -1,22 +1,12 @@
-#
```
tolusha, line 1: Pls. add license ;)

olm/update-resources.sh
```diff
@@ -173,6 +173,9 @@ updateNighltyBundle() {
       incrementNightlyVersion "${platform}"
     fi
 
+    cp -f "${ROOT_PROJECT_DIR}/deploy/crds/org.eclipse.che_checlusterbackups_crd.yaml" "${NIGHTLY_BUNDLE_PATH}/manifests"
+    cp -f "${ROOT_PROJECT_DIR}/deploy/crds/org.eclipse.che_checlusterrestores_crd.yaml" "${NIGHTLY_BUNDLE_PATH}/manifests"
```
tolusha, line 177: 1. crd generation should be done in `generateCRD` function
2. I believe references to crds must be added into `csv-config.yaml` files

tolusha: 3. Test files existence during release https://github.com/eclipse-che/che-operator/blob/main/make-release.sh#L235-L236
4. Update `olm/check-resources.sh` to check if crds are up to date like here https://github.com/eclipse-che/che-operator/blob/main/.github/bin/check-resources.sh#L52
5. Update `olm/prepare-community-operator.sh` to copy crds to PR like here https://github.com/eclipse-che/che-operator/blob/main/olm/prepare-community-operators-update.sh#L123-L126
6. Update `release-olm-files.sh` to create CRDs diffs on release

ibuziuk (commented):
@mmorhun have a design question - is it possible to configure periodic backups somehow?

mmorhun:
@ibuziuk directly in the backup CR or Che - no. But one can setup a cron job in a Kubernetes cluster which will trigger backup in the CR.
Alternatively, we could think about implementation on operator side as an improvement later.

Dockerfile
mmorhun: Thanks, will build the link depending on the container arhitecture

deploy/role.yaml
mmorhun: ok

deploy/operator.yaml
mmorhun: I've created [UBI based dockerfile](https://github.com/che-dockerfiles/che-backup-server-rest) too. @nickboldt please let me know if any other steps are needed from my side.

deploy/crds/org.eclipse.che_checlusterbackups_crd.yaml
```diff
@@ -0,0 +1,214 @@
...
+          description: CheClusterBackupSpec defines the desired state of CheClusterBackup
+          properties:
+            autoconfigureRestBackupServer:
+              description: Automatically setup container with REST backup server and
+                use the server in this configuration. Note, this will overwrite existing
+                configuration.
+              type: boolean
+            serverType:
+              description: If more than one backup server configured, should specify
+                which one to use. Allowed values are fields names form BackupServers
```
sleshchenko, line 41: I think `BackupServers` struct is a reference to go struct, but with CRD it's not really clear which field you refer to. Please try to rewrite in a way that would work for go and Json Schema.

deploy/crds/org.eclipse.che_checlusterbackups_crd.yaml
```diff
@@ -0,0 +1,214 @@
...
+                    true
+                  properties:
+                    credentialsSecretRef:
+                      description: Secret that contains username and password fields
+                      type: string
+                    hostname:
+                      description: Backup server host
+                      type: string
+                    password:
+                      description: Password to authenticate the user
```
sleshchenko, line 99: it's just an alternative to credentialsSecretRef, right? If so - let's make it clear in JSON Schema

pkg/apis/org/v1/backuprestorecommon_types.go
```diff
@@ -0,0 +1,112 @@
...
+	// User login on the remote server
+	Username string `json:"username,omitempty"`
+	// Private ssh key under 'ssh-privatekey' field for passwordless login
+	SshKeySecretRef string `json:"sshKeySecretRef,omitempty"`
+}
+
+// +k8s:openapi-gen=true
+// REST backup server configuration
+// Example: https://user:password@host:5000/repo/
+// Mandatory fields are: RepoPassword, Hostname
```
sleshchenko, line 68: this info does not seem to be populated to JSON Schema. I'm not sure if it's possible but would be ideal to get right required field in JSON Schema

pkg/backup_servers/sftp_server.go
```diff
@@ -0,0 +1,162 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// SftpServer implements BackupServer
+type SftpServer struct {
+	config orgv1.SftpServerConfing
+	ResticClient
+}
+
+func (s *SftpServer) PrepareConfiguration(client client.Client, namespace string) (bool, error) {
```
sleshchenko, line 35:
```suggestion
func (s *SftpServer) PrepareConfiguration(client client.Client, namespace string) (done bool, err error) {
```
it would help to understand what is `return true, fmt.Errorf("SFTP server username must be configured")` and similar

pkg/backup_servers/util.go
```diff
@@ -0,0 +1,69 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// getResticRepoPassword checks if the password for restic repository is specified and retrieves it.
+// It doesn't check the password correctness.
+// Returns:
+//  - password or empty string if password is not set.
```
sleshchenko, line 29:
```suggestion
//  - password or empty string if password is not set
```
just to be consistent

pkg/backup_servers/util.go
```diff
@@ -0,0 +1,69 @@
...
+//  - done status
+//  - error if any
+// Password from CR takes precedence over the password from secret.
```
sleshchenko, line 32: Is it specified in JSON Schema?

pkg/controller/add_checlusterrestore.go
```diff
@@ -0,0 +1,10 @@
+package controller
```
sleshchenko, line 1: some files miss license header

.github/bin/minikube/test-backup-restore.sh
```diff
@@ -0,0 +1,138 @@
...
+createBackupCR() {
+  kubectl apply -f - <<EOF
+apiVersion: org.eclipse.che/v1
+kind: CheClusterBackup
+metadata:
+  name: eclipse-che-backup
+  namespace: ${NAMESPACE}
+spec:
+  triggerNow: true
+  autoconfigureRestBackupServer: true
```
sleshchenko, line 37: Is it backup server under `servers[rest]` or `servers[internal]`?

.github/bin/minikube/test-backup-restore.sh
```diff
@@ -0,0 +1,138 @@
...
+
+createBackupCR() {
+  kubectl apply -f - <<EOF
+apiVersion: org.eclipse.che/v1
+kind: CheClusterBackup
+metadata:
+  name: eclipse-che-backup
+  namespace: ${NAMESPACE}
+spec:
+  triggerNow: true
```
sleshchenko, line 36: I'm not entirely sure about this usage.
Does controller trigger backup and reset this field to false?

sleshchenko (commented):
I'm not able to do a detailed review but I hope you'll find useful left minor comments.

ibuziuk:
```
Manually apply resources:
oc apply -f deploy/role.yaml
oc apply -f deplot/crds/org.eclipse.che_checlusterbackups_crd.yaml
oc apply -f deplot/crds/org.eclipse.che_checlusterrestores_crd.yaml
```

Just for the record that this trick will not be possible on OSD with dedicated / enhanced dedicated admins (no permissions for manual CRD creation).

olm/update-resources.sh
mmorhun: Thank you for the full list of places where I should patch the scripts

ibuziuk:
> Request backup by creating CR of CheClusterBackup type:

this CR should be created in the same namespace where the `eclipse-che` is installed right?

mmorhun:
@ibuziuk yes. More general: it should be crated where a Che operator deployment exists. So, you can create a namespace, deploy Che operator manually (using `deploy/operator.yaml` (but not forget about creating CRDs) ) and create restore CR. It will restore Che (even if it is not installed there yet).

mmorhun:
Rebased

pkg/apis/org/v1/backuprestorecommon_types.go
mmorhun: Probably that's not possible, but I've added duplicate commentary to the fields, so they will be propagated to CRDs yamls.

pkg/backup_servers/sftp_server.go
mmorhun: Thamk you for the remark. However, by default, if a function returns `(bool, error)` this means it returns done status and error if any. This is a convention for the backup and restore controllers. In case when a function returns something else or have, say, another return sequence/parameters, it is always described in a comment or variable name.

pkg/backup_servers/util.go
mmorhun: Now yes

pkg/controller/add_checlusterrestore.go
mmorhun: They are autogenerated, so I didn't do it, but will add the license, makes sense.

l0rd:
This is good work @mmorhun , thank you! I have a few comments:

1. Why `autoconfigureRestBackupServer` is in CheClusterBackup CR and not in CheClusterCR?
2. Is Che Operator config excluded from backup / restored? Restoring Che v7.40 with Che operator v7.41 looks fragile, it would be better if versions of Che operator and che/keycloak/postgres/registries versions match.
3. Why are you restoring the databases after CheCluster CR? That makes the flow more complicated.
4. Because of the problem of users IDs not being the same on different cluster, and the fact that we are going to require Che user == Cluster user even on plain Kubernetes, I don’t think we should support restore on a different cluster even on Kubernetes (unless user ids match on both clusters)
5. Plain text passwords in a Kubernetes CR is a no-no, we should only support secrets
6. `triggerNow` seams awkward: backup should be triggered when creating a CheClusterBackup CR.
7. How do the SFTP and REST servers store the backups? In volumes? Is it possible to configure it? 
8. Why are we supporting multiple backup severs configured in a CheClusterBackup (Servers)? From where does the requirement come? It makes things more complicated for an admin. Can we limit the CR to have one Server?

deploy/olm-catalog/nightly/eclipse-che-preview-openshift/csv-config.yaml
```diff
@@ -1,3 +1,3 @@
 operator-path: deploy/operator.yaml
 role-paths: [ "deploy/olm-catalog/nightly/eclipse-che-preview-openshift/generated/roles/role.yaml", "deploy/olm-catalog/nightly/eclipse-che-preview-openshift/generated/roles/cluster_role.yaml"]
-crd-cr-paths: ["deploy/crds/org_v1_che_crd.yaml"]
+crd-cr-paths: ["deploy/crds/org_v1_che_crd.yaml", "deploy/crds/org.eclipse.che_checlusterbackups_crd.yaml", "deploy/crds/org.eclipse.che_checlusterrestores_crd.yaml"]
```
tolusha, line 3: There is one more file to update
https://github.com/eclipse-che/che-operator/blob/main/deploy/olm-catalog/nightly/eclipse-che-preview-kubernetes/csv-config.yaml

olm/prepare-community-operators-update.sh
```diff
@@ -121,9 +121,11 @@ do
   > "${folderToUpdate}/${lastPackagePreReleaseVersion}/eclipse-che.v${lastPackagePreReleaseVersion}.clusterserviceversion.yaml"
 
   echo
-  echo "   - Update the CRD file"
+  echo "   - Update the CRD files"
   cp "${packageBaseFolderPath}/${lastPackagePreReleaseVersion}/org_v1_che_crd.yaml" \
   "${folderToUpdate}/${lastPackagePreReleaseVersion}/checlusters.org.eclipse.che.crd.yaml"
+  cp "${packageBaseFolderPath}/${lastPackagePreReleaseVersion}/org.eclipse.che_checlusterbackups_crd.yaml" "${folderToUpdate}/${lastPackagePreReleaseVersion}/org.eclipse.che_checlusterbackups_crd.yaml"
+  cp "${packageBaseFolderPath}/${lastPackagePreReleaseVersion}/org.eclipse.che_checlusterrestores_crd.yaml" "${folderToUpdate}/${lastPackagePreReleaseVersion}/org.eclipse.che_checlusterrestores_crd.yaml"
```
tolusha, line 128: naming convention is different. Let's have something common.
`checlusters.org.eclipse.che.crd.yaml`
`org.eclipse.che_checlusterbackups_crd.yaml`
`org.eclipse.che_checlusterrestores_crd.yaml`

olm/release-olm-files.sh
```diff
@@ -35,19 +35,26 @@ setLatestReleasedVersion() {
...
 
   compareResult=$(pysemver compare "${LAST_RELEASE_VERSION}" "7.27.2")
   if [ "${compareResult}" == "1" ]; then
+    # Version is newer than 7.27.2
     wget "https://raw.githubusercontent.com/eclipse-che/che-operator/${LAST_RELEASE_VERSION}/deploy/olm-catalog/stable/eclipse-che-preview-${platform}/manifests/che-operator.clusterserviceversion.yaml" \
         -q -O "${PRE_RELEASE_CSV}"
     wget "https://raw.githubusercontent.com/eclipse-che/che-operator/${LAST_RELEASE_VERSION}/deploy/olm-catalog/stable/eclipse-che-preview-${platform}/manifests/org_v1_che_crd.yaml" \
-        -q -O "${PRE_RELEASE_CRD}"
+        -q -O "${PRE_RELEASE_CHE_CRD}"
+    wget "https://raw.githubusercontent.com/eclipse-che/che-operator/${LAST_RELEASE_VERSION}/deploy/olm-catalog/stable/eclipse-che-preview-${platform}/manifests/org.eclipse.che_checlusterbackups_crd.yaml" \
```
tolusha, line 49: wget fails if there is no target file => release-olm-file.sh fails, so

1. We can remove version comparison and remove everything inside else block
2. Download file only if it is available.

mmorhun:
Hello @l0rd, thank you for the review!
Let me briefly explain thought on why these decisions were made.

On a discussion before the implementation, we decided to not to create a separate operator for backup and/or restore, but extend capabilities of existing Che operator. To not to impact performance of the main functionality, backup and restore capabilities are implemented in their own controllers, so they are run in separate threads and because of that, they don't directly overlap with the main Che operator functionality. Another andvantage of it is that we don't mix codebases of Che reconcile loop and backup / restore logic. However, we still can reuse some common code between controllers without any external dependencies.

About your points:

1. `autoconfigureRestBackupServer` is a property of `CheClusterBackup` because managing of a backup server is related to backup functionality and all backup related stuff are done in separate backup controller and so should be configured in backup CR.
2. Yes, it is excluded, because operator cannot manage itself. To restore Che we need to deploy Che operator first (Che CR is not needed). I think, that could be done from `chectl`. And this way we ensure that all components (postgres, keycloak, registries) are of the same version.
Also, as it was discussed, we do not support restoring of different version (at least now). It would be much more complicated as it might require some migrations in different places and problem with component versions arises as you mentioned.
3. To be 100% safe and compatible we have to get rid of previous database and database server and use database and its server of the same version as operator, so we can safely patch database then. If try to use existing (if any!) database, we don't know its version and level of it's damage, that makes everything fragile...
4. Yes, you are completely right. It just works now because there is no bindings to user IDs on Kubernetes infrastructure.
5. I added support for passwords in CR fields as we already have such abilities in the main Che operator CR. @l0rd should I remove plain passwords fields from backup/restore CRDs completely?
6. I tried to use backup/restore functionality while I was testing my code and to me `triggerNow` field looks useful. This way a user can pre-configure backup and use it when needed. Moreover, it allows to reuse the same configuration many times by just changing single flag and do not messing with a yaml and configs. And last, but not least, user can automate backups. Just a single patch query is needed.
7. A good strategy is to keep backups outside the cluster on an external backup server. But it is responsibility of admins to configure a backup server somewhere and how they configure backed storage is their own decision. It might be a pod with a volume or a server that stores data on hard drive in a directory or something else...
But there is also internal backup server. The main purpose of it is to be able to easily save backup data before version upgrade. Additionally, it makes testing easier. The only place we can safely deploy a backup pod is the Che server namespace. I haven't added a volume, so now backups are gone together with the backup pod, but it can be easily changed. Adding a volume shouldn't be a difficult task. Just let me know if you think we need it and I'll implement that in a follow-up PR.
8. Different backup servers require different configuration. For example, SFTP server requires SSH key to be provided, AWS S3 requires AWS secret (user token), etc. Using a big data object with all fields of all possible servers looks very confusing, so the configurations divided by backup server type. Anyway we have to have a field for each server type. A user may configure a few backup servers, but, I think, it is a rare case. In most cases only one is used and so configured. And if user provides only one configuration, `serverType` field is autodetected and user don't have to edit it manually. But for a tricky case it is still possible to have several servers of different types (more like a bonus, but we may restrict it artificially. However it will not prevent user from configuring several servers in the backup CR while editing it. In the end, it is just changes in validation part if needed, but I see no point in it right now).

Thank you, Mario, again, any improvement suggestions are welcomed!

.github/bin/minikube/test-backup-restore.sh
mmorhun: `autoconfigureRestBackupServer` configures `internal` server that has the same type as `rest` one.

.github/bin/minikube/test-backup-restore.sh
mmorhun: Yes, controller triggers backup and when it successfully finished, the flag is set to `false`.
The flag is useful. It allows to retrigger backup without creating entire CR again. Also it allows patching the CR by automated tool to have scheduled backups.

olm/release-olm-files.sh
mmorhun: I think we can run these two `wget`s failsafe

olm/prepare-community-operators-update.sh
mmorhun: I'd leave this improvement for a refactoring PR

pkg/apis/org/v1/backuprestorecommon_types.go
```diff
@@ -0,0 +1,99 @@
...
+// Mandatory fields are: ResticRepoPasswordSecretRef, Hostname, Repo, Username, SshKeySecretRef
+type SftpServerConfing struct {
+	// Holds reference to a secret with restic repository password under 'repo-password' field to encrypt / decrypt its content.
+	ResticRepoPasswordSecretRef string `json:"resticRepoPasswordSecretRef,omitempty"`
+	// Backup server host
+	Hostname string `json:"hostname,omitempty"`
+	// Backup server port
+	Port int `json:"port,omitempty"`
+	// Restic repository path, relative or absolute, e.g. /srv/repo
+	Repo string `json:"repo,omitempty"`
```
tolusha, line 52: `RepositoryPath`

pkg/apis/org/v1/backuprestorecommon_types.go
```diff
@@ -0,0 +1,99 @@
...
+	// Mandatory fields are: RepoPassword, Hostname.
+	Rest RestServerConfig `json:"rest,omitempty"`
+	// Amazon S3 or alternatives.
+	// Mandatory fields are: RepoPassword, Repo, AWS key+id or secret with it.
+	AwsS3 AwsS3ServerConfig `json:"awss3,omitempty"`
+}
+
+// +k8s:openapi-gen=true
+// SFTP backup server configuration
+// Example: user@host://srv/repo
```
tolusha, line 42: `:port`

pkg/apis/org/v1/backuprestorecommon_types.go
```diff
@@ -0,0 +1,99 @@
...
+// Mandatory fields are: ResticRepoPasswordSecretRef, Hostname, Repo, Username, SshKeySecretRef
+type SftpServerConfing struct {
+	// Holds reference to a secret with restic repository password under 'repo-password' field to encrypt / decrypt its content.
+	ResticRepoPasswordSecretRef string `json:"resticRepoPasswordSecretRef,omitempty"`
```
tolusha, line 46:
```suggestion
	RepositoryPasswordSecretRef string `json:"resticRepoPasswordSecretRef,omitempty"`
```

pkg/apis/org/v1/backuprestorecommon_types.go
```diff
@@ -0,0 +1,99 @@
...
+	// Backup server host
+	Hostname string `json:"hostname,omitempty"`
+	// Backup server port
+	Port int `json:"port,omitempty"`
+	// Restic repository path, relative or absolute, e.g. /srv/repo
+	Repo string `json:"repo,omitempty"`
+	// User login on the remote server
+	Username string `json:"username,omitempty"`
+	// Private ssh key under 'ssh-privatekey' field for passwordless login
+	SshKeySecretRef string `json:"sshKeySecretRef,omitempty"`
```
tolusha, line 56: `CredentialsSecretRef`

pkg/apis/org/v1/backuprestorecommon_types.go
```diff
@@ -0,0 +1,99 @@
...
+
+// +k8s:openapi-gen=true
+// REST backup server configuration
+// Example: https://user:password@host:5000/repo/
+// Mandatory fields are: ResticRepoPasswordSecretRef, Hostname
+type RestServerConfig struct {
+	// Holds reference to a secret with restic repository password under 'repo-password' field to encrypt / decrypt its content.
+	ResticRepoPasswordSecretRef string `json:"resticRepoPasswordSecretRef,omitempty"`
+	// Protocol to use when connection to the server
+	// Defaults to https.
```
tolusha, line 67: // +optional

pkg/apis/org/v1/checlusterbackup_types.go
```diff
@@ -0,0 +1,72 @@
...
+	AutoconfigureRestBackupServer bool `json:"autoconfigureRestBackupServer,omitempty"`
+	// Set to true to start backup process.
+	TriggerNow bool `json:"triggerNow"`
+	// If more than one backup server configured, should specify which one to use.
+	// Allowed values are keys from sibling Servers field.
+	ServerType string `json:"serverType,omitempty"`
+	// List of backup servers.
+	// Usually only one is used.
+	// In case of several available, ServerType should contain server to use.
+	Servers BackupServers `json:"servers,omitempty"`
```
tolusha, line 36: `BackupServersConfigs`
`Configs`

pkg/apis/org/v1/checlusterbackup_types.go
```diff
@@ -0,0 +1,72 @@
...
+// CheClusterBackupSpec defines the desired state of CheClusterBackup
+type CheClusterBackupSpec struct {
+	// Automatically setup container with REST backup server and use the server in this configuration.
+	// Note, this will overwrite existing configuration.
+	AutoconfigureRestBackupServer bool `json:"autoconfigureRestBackupServer,omitempty"`
+	// Set to true to start backup process.
+	TriggerNow bool `json:"triggerNow"`
+	// If more than one backup server configured, should specify which one to use.
+	// Allowed values are keys from sibling Servers field.
+	ServerType string `json:"serverType,omitempty"`
```
tolusha, line 32:
```suggestion
```

pkg/apis/org/v1/checlusterbackup_types.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:openapi-gen=true
+// CheClusterBackupSpec defines the desired state of CheClusterBackup
+type CheClusterBackupSpec struct {
+	// Automatically setup container with REST backup server and use the server in this configuration.
+	// Note, this will overwrite existing configuration.
+	AutoconfigureRestBackupServer bool `json:"autoconfigureRestBackupServer,omitempty"`
```
tolusha, line 27: `useInternalBackupServer`

pkg/apis/org/v1/checlusterbackup_types.go
```diff
@@ -0,0 +1,72 @@
...
+	// Usually only one is used.
+	// In case of several available, ServerType should contain server to use.
+	Servers BackupServers `json:"servers,omitempty"`
+}
+
+// +k8s:openapi-gen=true
+// CheClusterBackupStatus defines the observed state of CheClusterBackup
+type CheClusterBackupStatus struct {
+	// Backup result or error message
+	Message string `json:"message,omitempty"`
```
tolusha, line 43: message, phase, status, reason...

pkg/apis/org/v1/checlusterrestore_types.go
```diff
@@ -0,0 +1,84 @@
...
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CheClusterRestoreSpec defines the desired state of CheClusterRestore
+type CheClusterRestoreSpec struct {
+	// If true, deletes the CR after successful restore.
+	DeleteConfigurationAfterRestore bool `json:"deleteConfigurationAfterRestore,omitempty"`
```
tolusha, line 25:
```suggestion
```

pkg/apis/org/v1/checlusterrestore_types.go
```diff
@@ -0,0 +1,84 @@
...
+	// If true, copies backup servers configuration from backup CR
+	CopyBackupServerConfiguration bool `json:"copyBackupServerConfiguration,omitempty"`
+	// Snapshit it to restore from.
+	// If omitted, latest snapshot will be used.
+	SnapshotId string `json:"snapshotId,omitempty"`
+	// Set to true to start backup process.
+	TriggerNow bool `json:"triggerNow"`
+	// If more than one backup server configured, should specify which one to use.
+	// Allowed values are keys from sibling Servers field.
+	ServerType string `json:"serverType,omitempty"`
```
tolusha, line 35:
```suggestion
```

pkg/apis/org/v1/backuprestorecommon_types.go
mmorhun: No, `:` here is a delimiter between server address and repository path. In case if port is present, the delimiter is `/`. See [docs](https://restic.readthedocs.io/en/latest/030_preparing_a_new_repo.html#sftp).

pkg/apis/org/v1/checlusterrestore_types.go
mmorhun: I would keep it, but since you insist, I'll remove this field.

pkg/apis/org/v1/checlusterrestore_types.go
```diff
@@ -0,0 +1,84 @@
...
+
+// CheClusterRestoreSpec defines the desired state of CheClusterRestore
+type CheClusterRestoreSpec struct {
+	// If true, deletes the CR after successful restore.
+	DeleteConfigurationAfterRestore bool `json:"deleteConfigurationAfterRestore,omitempty"`
+	// If true, copies backup servers configuration from backup CR
+	CopyBackupServerConfiguration bool `json:"copyBackupServerConfiguration,omitempty"`
+	// Snapshit it to restore from.
+	// If omitted, latest snapshot will be used.
+	SnapshotId string `json:"snapshotId,omitempty"`
```
tolusha, line 30: Add info to backup status

pkg/apis/org/v1/checlusterrestore_types.go
```diff
@@ -0,0 +1,84 @@
...
+	// Set to true to start backup process.
```
tolusha, line 31:
```suggestion
	// Set to true to start restore process.
```

pkg/apis/org/v1/checlusterrestore_types.go
```diff
@@ -0,0 +1,84 @@
...
+	TriggerNow bool `json:"triggerNow"`
+	// If more than one backup server configured, should specify which one to use.
+	// Allowed values are keys from sibling Servers field.
+	ServerType string `json:"serverType,omitempty"`
+	// List of backup servers.
+	// Usually only one is used.
+	// In case of several available, serverType should contain server to use.
+	Servers BackupServers `json:"servers,omitempty"`
+	// Amendments for CR from backup
+	CROverrides CROverrides `json:"crOverrides,omitempty"`
```
tolusha, line 41:
```suggestion
```

pkg/apis/org/v1/checlusterrestore_types.go
```diff
@@ -0,0 +1,84 @@
...
+type CheClusterRestoreSpec struct {
+	// If true, deletes the CR after successful restore.
+	DeleteConfigurationAfterRestore bool `json:"deleteConfigurationAfterRestore,omitempty"`
+	// If true, copies backup servers configuration from backup CR
+	CopyBackupServerConfiguration bool `json:"copyBackupServerConfiguration,omitempty"`
+	// Snapshit it to restore from.
+	// If omitted, latest snapshot will be used.
+	SnapshotId string `json:"snapshotId,omitempty"`
+	// Set to true to start backup process.
+	TriggerNow bool `json:"triggerNow"`
```
tolusha, line 32:
```suggestion
```

pkg/apis/org/v1/backuprestorecommon_types.go
```diff
@@ -0,0 +1,99 @@
...
+	AWS_ACCESS_KEY_ID_SECRET_KEY     = "awsAccessKeyId"
+	AWS_SECRET_ACCESS_KEY_SECRET_KEY = "awsSecretAccessKey"
+)
+
+// +k8s:openapi-gen=true
+// List of supported backup servers
+type BackupServers struct {
+	// Rest server within the cluster.
+	// The server and configuration are created by operator when AutoconfigureRestBackupServer is true.
+	Internal RestServerConfig `json:"internal,omitempty"`
```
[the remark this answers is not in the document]
mmorhun: ok, will use `rest` instead since we agreed to delete `serverType`

pkg/apis/org/v1/checlusterrestore_types.go
mmorhun: Sigh... :disappointed:  , ok, I will remove restore support on a different cluster...

pkg/apis/org/v1/checlusterbackup_types.go
tolusha, line 43:
```suggestion
	// Message explaining the state of the backup or an error message
	Message string `json:"message,omitempty"`
	// InProgress, Failed, Successed
	State string `json:"state,omitempty"`
	// 
	SnapshotId string `json:"snapshotId,omitempty"`
```

pkg/apis/org/v1/checlusterbackup_types.go
```diff
@@ -0,0 +1,72 @@
...
+	// Shows when backup was done last time
+	LastBackupTime string `json:"lastBackupTime,omitempty"`
```
tolusha, line 45:
```suggestion
	BackupTime string `json:"lastBackupTime,omitempty"`
```

pkg/apis/org/v1/checlusterrestore_types.go
```diff
@@ -0,0 +1,84 @@
...
+	// that has ingress domain different from the cluster on which the backup was done.
+	IngressDomain string `json:"ingressDomain,omitempty"`
+}
+
+// CheClusterRestoreStatus defines the observed state of CheClusterRestore
+type CheClusterRestoreStatus struct {
+	// Backup result or error message
+	Message string `json:"message,omitempty"`
+	// Describes restore progress
+	Stage string `json:"stage,omitempty"`
```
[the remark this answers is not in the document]
tolusha: Phase string `json:"phase,omitempty"`

pkg/apis/org/v1/checlusterrestore_types.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CheClusterRestoreSpec defines the desired state of CheClusterRestore
+type CheClusterRestoreSpec struct {
+	// If true, deletes the CR after successful restore.
+	DeleteConfigurationAfterRestore bool `json:"deleteConfigurationAfterRestore,omitempty"`
+	// If true, copies backup servers configuration from backup CR
```
tolusha, line 26:
```suggestion
	// If true, copies backup servers configuration from backup CR to `BackupServerConfig` field
```

deploy/crds/__crd.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: apiextensions.k8s.io/v1
```
tolusha, line 1: pls remove.
or rebase on master and rerun olm/update-resources.sh

deploy/crds/org.eclipse.che_v1_checlusterbackup_cr.yaml
```diff
@@ -0,0 +1,35 @@
...
+metadata:
+  name: eclipse-che-backup
+spec:
+  # If true will setup REST backup server pod and add configuration to use it
+  autoconfigureRestBackupServer: false
+  # Set to true to start backup process
+  triggerNow: false
+  # Might be omitted if only one server configured
+  serverType: 'sftp'
+  # Usually only one server is needed
```
tolusha, line 12: Example is incorrect. Since it must contain only one backup configuration

deploy/crds/org.eclipse.che_v1_checlusterbackup_cr.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: org.eclipse.che/v1
```
tolusha, line 1: Add license pls

deploy/crds/org.eclipse.che_v1_checlusterrestore_cr.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: org.eclipse.che/v1
```
tolusha, line 1: Add license and adopt example

olm/release-olm-files.sh
```diff
@@ -35,19 +35,29 @@ setLatestReleasedVersion() {
 downloadLatestReleasedBundleCRCRD() {
   mkdir -p "${STABLE_BUNDLE_PATH}/manifests" "${STABLE_BUNDLE_PATH}/generated/${platform}" "${STABLE_BUNDLE_PATH}/metadata"
   PRE_RELEASE_CSV="${STABLE_BUNDLE_PATH}/generated/${platform}/che-operator.clusterserviceversion.yaml"
-  PRE_RELEASE_CRD="${STABLE_BUNDLE_PATH}/generated/${platform}/org_v1_che_crd.yaml"
+  PRE_RELEASE_CHE_CRD="${STABLE_BUNDLE_PATH}/generated/${platform}/org_v1_che_crd.yaml"
+  PRE_RELEASE_CHE_BACKUP_CRD="${STABLE_BUNDLE_PATH}/generated/${platform}/org.eclipse.che_checlusterbackups_crd.yaml"
+  PRE_RELEASE_CHE_RESTORE_CRD="${STABLE_BUNDLE_PATH}/generated/${platform}/org.eclipse.che_checlusterrestores_crd.yaml"
 
   compareResult=$(pysemver compare "${LAST_RELEASE_VERSION}" "7.27.2")
```
tolusha, line 42: Pls remove version comparison. We way ahead of 7.27.2

pkg/apis/org/v1/backuprestorecommon_types.go
```diff
@@ -0,0 +1,121 @@
...
+	// Amazon S3 or compatible alternatives.
+	// Mandatory fields are: RepositoryPasswordSecretRef, RepositoryPath, CredentialsSecretRef.
+	// +optional
+	AwsS3 *AwsS3ServerConfig `json:"awss3,omitempty"`
+}
+
+// +k8s:openapi-gen=true
+// SFTP backup server configuration
+// Example: user@host://srv/repo
+// Mandatory fields are: Username, Hostname, RepositoryPath, RepositoryPasswordSecretRef, SshKeySecretRef
```
tolusha, line 47: If field is mandatory then neither  `+optional` nor `omitempty`

pkg/apis/org/v1/backuprestorecommon_types.go
```diff
@@ -0,0 +1,121 @@
...
+	// +optional
+	RepositoryPasswordSecretRef string `json:"repositoryPasswordSecretRef,omitempty"`
+	// Secret that contains username and password fields to login into restic server.
+	// Note, each repository is encrypted with own password. See ResticRepoPasswordSecretRef field.
+	// +optional
+	CredentialsSecretRef string `json:"credentialsSecretRef,omitempty"`
+}
+
+// +k8s:openapi-gen=true
+// Mandatory fields are: RepositoryPasswordSecretRef, RepositoryPath, CredentialsSecretRef.
```
tolusha, line 97: There is no a such field like CredentialsSecretRef

pkg/apis/org/v1/checlusterbackup_types.go
```diff
@@ -0,0 +1,77 @@
...
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+// Important: when any changes are made, CRD must be regenerated.
+// Please use olm/update-resources.sh script for that.
```
tolusha, line 14: It is covered by olm/check-resources.sh
We can remove it here

pkg/controller/checlusterbackup/backup_data_collector.go
```diff
@@ -0,0 +1,286 @@
...
+	CreationDate        string `json:"creationDate"`
+	CheVersion          string `json:"cheVersion"`
+	Infrastructure      string `json:"infrastructure"`
+	AppsDomain          string `json:"appsDomain"`
+	Namespace           string `json:"namespace"`
+}
+
+func createBackupMetadataFile(bctx *BackupContext, destDir string) (bool, error) {
+	infra := "Kubernetes"
+	isOpenShift, isOpenShift4, _ := util.DetectOpenShift()
```
tolusha, line 52: You don't have to detect it here. 
It has been already detected.

pkg/util/util.go
```diff
@@ -442,3 +516,37 @@ func ComputeHash256(yamlFile string) (string, error) {
 	sha := base64.URLEncoding.EncodeToString(hasher.Sum(nil))
 	return sha, nil
 }
+
+func ReloadCheCluster(client client.Client, cheCluster *orgv1.CheCluster) error {
```
tolusha, line 520: It is already here 
https://github.com/eclipse-che/che-operator/blob/main/pkg/deploy/checluster.go#L81

pkg/util/util.go
```diff
@@ -312,6 +317,75 @@ func getAPIUrlsForOpenShiftV4() (apiUrl string, apiInternalUrl string, err error
 	return apiUrl, apiInternalUrl, nil
 }
 
+func GetRouterCanonicalHostname(client client.Client, namespace string) (string, error) {
```
tolusha, line 320: Why do we need this?

deploy/crds/org.eclipse.che_v1_checlusterbackup_cr.yaml
mmorhun: Maybe we should keep all servers configured as an example and just warn user in comment that only one configuration at a time allowed?

olm/release-olm-files.sh
mmorhun: ok

pkg/apis/org/v1/backuprestorecommon_types.go
mmorhun: They are mandatory for the server configuration, but they are not mandatory for the CR

deploy/crds/org.eclipse.che_v1_checlusterbackup_cr.yaml
tolusha: ok

pkg/util/util.go
mmorhun: To support database patching in case of different restoring namespace within the same cluster.

pkg/util/util.go
mmorhun: I cannot use that version, because it requires deploy context which is main controller specific.

pkg/apis/org/v1/backuprestorecommon_types.go
mmorhun: Thanks, I see.

mmorhun:
CQs https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23352 and https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23353 approved.

tolusha (approved):
Well done.

ibuziuk (commented):
@mmorhun indeed great job. 
Could you please clarify yet another design / flow  question for the following use-case:

- Eclipse Che is installed in namespace A
- Need to migrate the workloads to namespace B on the very same cluster

How the backup / restore procedure for this case is going to look like?

flacatus:
I've finished to test locally the changes and looks really nice!
Awesome job @mmorhun !

flacatus (approved):

mmorhun:
@ibuziuk you can do it, but the automatic copying of the backup configuration is not possible and should be done manually.
You need:
1. Create a backup CR:
```yaml
kind: CheClusterBackup
metadata:
  name: eclipse-che-backup
spec:
  triggerNow: true
  useInternalBackupServer: true
```
2. Wait until backup is finished
3. Deploy Che operator in target (for new Che) namespace.
The easiest way to do it is to start a fresh Che installation using chectl, but break it on creating Che CR step (or later to be safe, just delete Che CR manually then).
4. Create restore CR in the target namespace, for example:
```yaml
kind: CheClusterRestore
metadata:
  name: eclipse-che-restore
spec:
  backupServerConfig:
    rest:
      hostname: backup-rest-server-service.eclipse-che.svc.cluster.local
      port: 8000
      protocol: http
      repositoryPasswordSecretRef: backup-rest-server-repo-password
      repositoryPath: che
```
Replace `eclipse-che` in `hostname` with existing Che installation namespace.

5. Wait until restore is done.

mmorhun:
@ibuziuk for your use-case:
1. No need to do any manipulations with templates, everything is merged. Just make sure you have latest version installed.
2. Create a backup by creating backup CR:
```yaml
kind: CheClusterBackup
metadata:
  name: eclipse-che-backup
spec:
  useInternalBackupServer: true
```
2. Manually copy `backup-rest-server-configuration` of type `CheBackupServerConfiguration to target namespace and patch service url as describe [above](https://github.com/eclipse-che/che-operator/pull/844#issuecomment-861544821).
3. Request restore by creating restore CR:
```yaml
kind: CheClusterRestore
metadata:
  name: eclipse-che-restore
spec:
  backupServerConfigRef: backup-rest-server-configuration
```

deploy/operator.yaml
nickboldt: lots of steps required to productize a new container. See https://issues.redhat.com/browse/CRW-1878

PR #894, opened 2021-07-08
feat: Make DevWorkspaceRouting a pre-requisite of che-operator.

metlos:
### What does this PR do?
In order to be able to merge codebases of DWCO and CO, we need to pre-install the devworkspacerouting CRD into the cluster, if supported by the target platform (i.e. this should NOT be done on OpenShift 3).

This PR includes the the DevWorkspaceRouting CRD in the deployment files of the che-operator and updates (or at least tries to) the OLM deployment scripts to include it there, too.

I am personally unfamiliar with OLM so I would appreciate any and all help with testing this PR.

Note that currently, the DevWorkspaceRouting CRD is just contained in the repo. Alternatively, we could think about downloading it from DWO, but then we should be very careful about downloading the correct version (should we use the latest version during or should we stick to some released version?).

N/A

eclipse/che#20088

### How to test this PR?
Run `./deploy.sh`. The cluster should contain the `devworkspaceroutings.controller.devfile.io` CRD after that.

### PR Checklist
          
### Reviewers


tolusha:
To be honest shipping crd that is not related to che-operator is not good.
I have another option. What about creating devworkspacerouting CRD in dev_workspage.go ?

metlos:
There is a chicken egg problem there. If we want to have a single operator for che and devworkspace, we need the controller manager to reconcile all the necessary objects. I.e. both CheCluster and DevWorkspaceRouting. For the controller manager to successfully start, those objects need to be present in the cluster. 

I.e. if we want to install DevWorkspaceRouting lazily (as we do atm), we need to keep the codebases separate. Alternatively we could try running 2 controller managers in the same process in parallel, but I'm not sure how that would behave wrt caching, etc.

Note that @l0rd is ok with preinstalling DevWorkspaceRouting, see https://github.com/eclipse/che/issues/20088#issuecomment-875543533

tolusha:
@metlos 
I see your point.
So, regarding with the PR. Don't modify csv-config files. They must only Eclipse Che operator CRD.

deploy/olm-catalog/nightly/eclipse-che-preview-openshift/manifests/devworkspaceroutings.controller.devfile.io.CustomResourceDefinition.yaml
```diff
@@ -0,0 +1,3708 @@
...
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.3.0
+  creationTimestamp: null
+  labels:
+    app.kubernetes.io/name: devworkspace-controller
+    app.kubernetes.io/part-of: devworkspace-operator
+  name: devworkspaceroutings.controller.devfile.io
```
sleshchenko, line 20: I prefer to run discussions in threads, so starting it instead of commenting PR.

DevWorkspace Routing CRD must not be shipped by Che Operator and even by DevWorkspace Che Operator.
It may work somehow now, but it's supposed to fully fail later when Descoping Operators is done.

So, devworkspace routing CRD must be installed only by one operator, and it's DevWorkspace Operator who really owns that API. Others (like DWCO) just depend on it.

The only proper solution I see is depending with OLM dependencies https://github.com/eclipse-che/che-operator/pull/890 in a dedicated channel ATM.
If we need the same Che Operator deployment in both channels, DWCO should be no-op is DevWorkspaceRouting CRD is not available, I think.

metlos:
I'm closing this in favor of a design where che-operator will be still in charge of the installation of DWR and will have alternative watch sets based on the CRD presence in the cluster, as discussed with @sleshchenko.

PR #1027, opened 2021-08-20
feat: Syncing of proxy settings and self-signed cert to the user's workspace namespace

metlos:
### What does this PR do?
This adds the ability to sync settings to the workspace namespaces of the users.

The new controller watches namespaces (or projects on OpenShift). If a namespace has appropriate label (`org.eclipse.che/workspace-namespace-owner-uid`), it is considered a user workspace namespace and is synced to.

The value of the label is supposed to be the UID of the user. On OpenShift, this is the UID of the user object and on Kubernetes, the UID of the user as obtained from the OAuth token of the configured OIDC provider.

eclipse/che#20168

### How to test this PR?
create a namespace, label it and check that the proxy settings are synced to it in a config map and self signed certificates in a secret.

### PR Checklist

### Reviewers


controllers/cheuser/controller.go
```diff
@@ -0,0 +1,336 @@
...
+
+	return bld.Complete(r)
+}
+
+func (r *CheUserNamespaceReconciler) watchRulesForSecrets(ctx context.Context) handler.EventHandler {
+	return &handler.EnqueueRequestsFromMapFunc{
+		ToRequests: handler.ToRequestsFunc(func(mo handler.MapObject) []reconcile.Request {
+			if isLabeledAsUserSettings(mo.Meta) && r.isInManagedNamespace(ctx, mo.Meta) {
+				return asReconcileRequestForNamespace(mo.Meta)
+			} else if true {
```
AndrienkoAleksandr, line 82: Sorry, maybe simple `else {` will be enough ?

controllers/cheuser/controller.go
```diff
@@ -0,0 +1,336 @@
...
+			NamespacedName: types.NamespacedName{Name: obj.GetNamespace()},
+		},
+	}
+}
+
+func (r *CheUserNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	ctx := context.Background()
+
+	info, err := r.namespaceCache.ExamineNamespace(ctx, req.Name)
+	if err != nil || info == nil || info.OwnerUid == "" {
```
AndrienkoAleksandr, line 155: Maybe err, we should log?

controllers/cheuser/controller.go
```diff
@@ -0,0 +1,336 @@
...
+	return ctrl.Result{}, nil
+}
+
+func findManagingCheCluster(key types.NamespacedName) *v2alpha1.CheCluster {
+	instances := devworkspace.GetCurrentCheClusterInstances()
+	if len(instances) == 0 {
+		return nil
+	}
+
+	if len(instances) == 1 {
```
AndrienkoAleksandr, line 203:
```suggestion
	if len(instances) > 1 {
```

metlos: Nope, the logic is this:

If there is a single instance of `CheCluster` in the k8s cluster, then we don't need to look for it using the labels on the namespace (the `key`). If there are more than `CheCluster` instances, we're only able to figure out which one is in charge of this user's namespace by looking at the labels of the namespace.

controllers/cheuser/controller.go
metlos: huh, where did that come from?  🤭

pkg/deploy/sync.go
```diff
@@ -28,13 +28,23 @@ import (
...
 	if !deployContext.CheCluster.ObjectMeta.DeletionTimestamp.IsZero() {
 		return true, nil
 	}
 
+	return DoSync(deployContext, blueprint, diffOpts...)
+}
+
+// Sync syncs the blueprint to the cluster in a generic (as much as Go allows) manner.
+// Returns true if object is up to date otherwiser returns false
+func DoSync(deployContext *DeployContext, blueprint metav1.Object, diffOpts ...cmp.Option) (bool, error) {
```
tolusha, line 47: Why do we need to sync if checluster is being deleted. What is a case?

controllers/devworkspace/controller.go
```diff
@@ -87,6 +87,13 @@ func GetCurrentCheClusterInstances() map[client.ObjectKey]v2alpha1.CheCluster {
 	return ret
 }
 
+func CleanCheClusterInstancesForTest() {
```
tolusha, line 90: Could we move this to test file?

controllers/cheuser/namespacecache.go
```diff
@@ -0,0 +1,121 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	ownerUidLabel            string = "che.eclipse.org/workspace-namespace-owner-uid"
```
tolusha, line 18: To be consistent with const below
```suggestion
	cheClusterOwnerUidLabel            string = "che.eclipse.org/workspace-namespace-owner-uid"
```

controllers/cheuser/controller.go
```diff
@@ -0,0 +1,344 @@
...
+	scheme         *runtime.Scheme
+	namespaceCache namespaceCache
+}
+
+var _ reconcile.Reconciler = (*CheUserNamespaceReconciler)(nil)
+
+var (
+	configMapDiffOpts cmp.Option = cmp.Options{
+		cmpopts.IgnoreFields(corev1.ConfigMap{}, "TypeMeta", "ObjectMeta"),
+	}
```
tolusha, line 46: It is out of scope of this PR but we already have `configMapDiffOpts`. It would be nice to reuse
https://github.com/eclipse-che/che-operator/blob/main/pkg/deploy/configmap.go#L23

controllers/cheuser/controller.go
```diff
@@ -0,0 +1,344 @@
...
+	secretDiffOpts cmp.Option = cmp.Options{
```
tolusha, line 47: `secretDiffOpts`
https://github.com/eclipse-che/che-operator/blob/main/pkg/deploy/secret.go#L30

controllers/cheuser/controller.go
```diff
@@ -0,0 +1,344 @@
...
+		CheCluster: org.AsV1(checluster),
+		ClusterAPI: deploy.ClusterAPI{
+			Client:          r.client,
+			NonCachedClient: r.client,
+			DiscoveryClient: nil,
+			Scheme:          r.scheme,
+		},
+	}
+
+	isSelfSignedCertUsed, err := deploy.IsSelfSignedCertificateUsed(deployContext)
```
tolusha, line 184: It is better to check if corresponding secret exists without invoking `IsSelfSignedCertificatdUsed`

tolusha: A common problem is that reconciling self-signed certificate is not enough. 
Others certificates are located in `ca-certs-merged` config map
see https://github.com/eclipse-che/che-operator/blob/main/pkg/deploy/tls.go#L505

tolusha: Besides this, we have `che-git-self-signed-cert` configmap which store self-signed certificate and host for git repository. see https://github.com/eclipse-che/che-operator/blob/main/pkg/deploy/server/server_deployment.go#L84-L109

pkg/deploy/sync.go
metlos: I just wanted to use this in a generic way and the shortcut logic in a generic sync method seemed a little bit out of place to me, following the "principle of least astonishment". Otherwise you're right, all the callers in the current codebase should behave equivalently regardless of the method used.

controllers/devworkspace/controller.go
metlos: This needs to access data private to the devworkspace package but needs to be called from test code in other packages. Moving it to `controller_test.go` makes it invisible to other packages unfortunately.

controllers/cheuser/controller.go
metlos: `IsSelfSignedCertificateUsed` is no longer used and `ca-certs-merged` should be handled. I chose not to handle `che-git-self-signed-cert` just to not further increase the scope of this issue. We can handle it separately later (I am also glad, we managed to make the CI (almost) pass after quite some time, so I don't want to go through that ordeal again ;) ).

metlos: Btw handling of the git cert is more complex, because it only makes sense if we also sync gitconfig - for that we not only need the git host (that should be present in the `che-git-self-signed-cert` configmap, but also user name and email. This is intentionally out of scope of this PR so we'll need to remember to handle that in a separate issue in the future.

controllers/usernamespace/controller.go
```diff
@@ -0,0 +1,393 @@
...
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "ConfigMap",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      targetConfigMapName,
+			Namespace: targetNs,
+			Labels: defaults.AddStandardLabelsForComponent(checluster, userSettingsComponentLabelValue, map[string]string{
+				constants.DevWorkspaceMountLabel:              "true",
+				"config.openshift.io/inject-trusted-cabundle": "true",
```
tolusha, line 300: You don't have to add this label. `deploy.CheAllCACertsConfigMapName` already contains all OpenShift certificates.

tolusha (commented):
Other then that sounds good.

.github/bin/common.sh
```diff
@@ -167,24 +167,49 @@ installYq() {
...
   oc get events -n ${DEVWORKSPACE_CONTROLLER_TEST_NAMESPACE} > ${ARTIFACTS_DIR}/events-${DEVWORKSPACE_CONTROLLER_TEST_NAMESPACE}.txt
   oc get events -n ${DEVWORKSPACE_CHE_OPERATOR_TEST_NAMESPACE} > ${ARTIFACTS_DIR}/events-${DEVWORKSPACE_CHE_OPERATOR_TEST_NAMESPACE}.txt
-  set -x
+  set -e
+}
+
+collectDevworkspaceOperatorLogs() {
+  mkdir -p ${ARTIFACTS_DIR}/devworkspace-operator
+
+  oc get events -n devworkspace-controller > ${ARTIFACTS_DIR}/events-devworkspace-controller.txt
```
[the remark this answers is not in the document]
metlos: I think it makes sense to do both - collect events from the DWO namespace as well as from the test namespaces where DWO and CO deploy the workspaces.

tolusha (approved):

.github/bin/common.sh
```diff
@@ -167,24 +167,49 @@ installYq() {
...
+
+  #determine the name of the devworkspace controller manager pod
+  local CONTROLLER_POD_NAME=$(oc get pods -n devworkspace-controller -l app.kubernetes.io/name=devworkspace-controller -o json | jq -r '.items[0].metadata.name')
+  local WEBHOOK_SVR_POD_NAME=$(oc get pods -n devworkspace-controller -l app.kubernetes.io/name=devworkspace-webhook-server -o json | jq -r '.items[0].metadata.name')
+  
+  # save the logs of all the containers in the DWO pod
+  for container in $(oc get pod -n devworkspace-controller ${CONTROLLER_POD_NAME} -o json | jq -r '.spec.containers[] | .name'); do
+    mkdir -p ${ARTIFACTS_DIR}/devworkspace-operator/${CONTROLLER_POD_NAME}
```
sleshchenko, line 190: What if we just bump everything in the namespace? Like we do on dashboard/devworkspace operator side https://github.com/devfile/devworkspace-operator/blob/main/.ci/common.sh#L18.
You just need to specify namespace as an argument.

Also, please note that chectl:logs extract only certain pods, (not sure if che-gateway is respected), so in additional to chectl:logs, we also do bummPodsInfo eclipse-che, and we have all jobs, all containers info.

controllers/devworkspace/util.go
```diff
@@ -0,0 +1,26 @@
+package devworkspace
+
+import (
+	chev2alpha1 "github.com/eclipse-che/che-operator/api/v2alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type DevworkspaceState int
```
sleshchenko, line 8:
```suggestion
type DevWorkspaceState int
```
it would be a consistent naming I think, but that's not critical

controllers/devworkspace/util.go
```diff
@@ -0,0 +1,26 @@
...
+
+const (
+	DevworkspaceStateNotPresent = 0
```
sleshchenko, line 11:
```suggestion
	DevworkspaceStateNotPresent DevworkspaceState = 0
	DevworkspaceStateDisabled   DevworkspaceState = 1
	DevworkspaceStateEnabled    DevworkspaceState = 2
```
it would be a bit better from go typing perspective.

controllers/usernamespace/controller.go
```diff
@@ -0,0 +1,392 @@
+package usernamespace
```
sleshchenko, line 1: Seems new files miss license header

controllers/usernamespace/controller.go
```diff
@@ -0,0 +1,392 @@
...
+		ToRequests: handler.ToRequestsFunc(func(mo handler.MapObject) []reconcile.Request {
+			return asReconcileRequestForNamespaceIf(mo.Meta, func() bool {
+				return (isLabeledAsUserSettings(mo.Meta) && r.isInManagedNamespace(ctx, mo.Meta)) ||
+					r.hasNameAndIsColocatedWithCheCluster(ctx, mo.Meta, deploy.CheAllCACertsConfigMapName)
+			})
+		}),
+	}
+}
+
+func (r *CheUserNamespaceReconciler) hasNameAndIsColocatedWithCheCluster(ctx context.Context, obj metav1.Object, names ...string) bool {
```
sleshchenko, line 89:
```suggestion
func (r *CheUserNamespaceReconciler) hasNameAndIsCollocatedWithCheCluster(ctx context.Context, obj metav1.Object, names ...string) bool {
```

controllers/usernamespace/controller.go
```diff
@@ -0,0 +1,392 @@
...
+	ret, ok := instances[key]
+
+	if ok {
+		return &ret
+	} else {
+		return nil
+	}
+}
+
+func (r *CheUserNamespaceReconciler) reconcileSelfSignedCert(ctx context.Context, deployContext *deploy.DeployContext, targetNs string, checluster *v2alpha1.CheCluster) error {
```
sleshchenko, line 226: I know, these methods (`reconcileTrustedCerts`, `reconcileSelfSignedCert`, ...) are not too big  but in the end it blows controller.go and later it will grow and grow. Which I think makes it more difficult to understand what is supposed to happen here.
So, I think it makes sense to move them to separate three files (trustedCerts, selfSignedCerts, proxySettings), but it's up to you.

controllers/usernamespace/controller_test.go
```diff
@@ -0,0 +1,361 @@
...
+			"trusted1": "trusted cert 1",
+			"trusted2": "trusted cert 2",
+		},
+	}
+	if err := cl.Create(ctx, caCerts); err != nil {
+		t.Fatal(err)
+	}
+
+	r := devworkspace.New(cl, scheme)
+	// the reconciliation needs to run twice for it to be trully finished - we're setting up finalizers etc...
```
sleshchenko, line 92:
```suggestion
	// the reconciliation needs to run twice for it to be truly finished - we're setting up finalizers etc...
```

controllers/usernamespace/controller_test.go
```diff
@@ -0,0 +1,361 @@
...
+
+func TestRequiresLabelsToMatchOneOfMultipleCheCluster(t *testing.T) {
+	test := func(t *testing.T, infraType infrastructure.Type, namespace metav1.Object) {
+		ctx := context.TODO()
+		scheme, cl, r := setup(infraType, namespace.(runtime.Object))
+		setupCheCluster(t, ctx, cl, scheme, "che1", "che")
+		setupCheCluster(t, ctx, cl, scheme, "che2", "che")
+
+		res, err := r.Reconcile(reconcile.Request{NamespacedName: types.NamespacedName{Name: namespace.GetName()}})
+		if err != nil {
```
sleshchenko, line 177: these three lines can be replaced with
```suggestion
assert.NilError(t, err)
```

controllers/usernamespace/controller_test.go
```diff
@@ -0,0 +1,361 @@
...
+		}
+
+		proxySettings := corev1.ConfigMap{}
+		if err := cl.Get(ctx, client.ObjectKey{Name: "che-eclipse-che-proxy-settings", Namespace: namespace.GetName()}, &proxySettings); err != nil {
+			t.Fatal(err)
+		}
+		if proxySettings.GetAnnotations()[constants.DevWorkspaceMountAsAnnotation] != "env" {
+			t.Errorf("proxy settings should be annotated as mount as 'env' but was '%s'", proxySettings.GetLabels()[constants.DevWorkspaceMountAsAnnotation])
+		}
+		if proxySettings.GetLabels()[constants.DevWorkspaceMountLabel] != "true" {
```
sleshchenko, line 276: what do you think about using asserts, like the following:
```go
assert.Equal(t, proxySettings.GetLabels()[constants.DevWorkspaceMountLabel], "true",
			"proxy settings label %s is not matched", constants.DevWorkspaceMountLabel)
```

controllers/usernamespace/controller_test.go
```diff
@@ -0,0 +1,361 @@
...
+					workspaceNamespaceOwnerUidLabel: "uid",
+					cheNameLabel:                    "che",
+					cheNamespaceLabel:               "che1",
+				},
+			},
+		})
+	})
+}
+
+func TestCreatesDataInNamespace(t *testing.T) {
```
sleshchenko, line 253: I believe tests readability would benefit if you separate it into three tests: TestPropagatingTrutedCerts, TestPropagatingSelfSignedCert, TestPropagatingProxySettings.

sleshchenko (approved):

.github/bin/common.sh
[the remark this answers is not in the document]
sleshchenko: it also makes sense to collect DW, DWT, DWR if they are not there: https://github.com/devfile/devworkspace-operator/blob/main/.ci/oci-devworkspace-happy-path.sh#L31

controllers/devworkspace/util.go
metlos: Hmm, from looking at CO codebase, we're consistently "Devworkspace" :) So I'd like to keep it out of this PR and deal with it in another issue. But you're of course right, we should use the same terminology...

controllers/usernamespace/controller_test.go
metlos: True, but let's wait with that until we have even more data synced....

.github/bin/common.sh
metlos: Created https://github.com/eclipse/che/issues/20504 for that.

controllers/usernamespace/controller.go
metlos: Agreed, but we can easily move then later, once it becomes a problem.

max-cx:
Hi, a Yes/No question about this issue:

Will the outcome require any changes to the relevant content of the _Installation Guide_ or _Administration Guide_ or _End-user Guide_?

Yes/No?

max-cx:
@tolusha, a Yes/No question about this issue:

Will the outcome require any changes to the relevant content of the _Installation Guide_ or _Administration Guide_ or _End-user Guide_?

Yes/No?

max-cx:
@meltos, could you please clarify what the user experience impact of this is (for the sake of writing the Release Notes text)?
You can reply here or in https://issues.redhat.com/browse/CRW-2397

tolusha:
@max-cx 
No

PR #1045, opened 2021-09-01
feat: secure workspace services

sparkoo:
### What does this PR do?
Secures workspace services with devworkspace engine on OpenShift. 

There is a slight change in exposing logic. The ingresses/routes are not merged into one when exposing same port.

Beside that, it also always deploys with gateway on kubernetes when devworkspaces are enabled.

### How it works?

Nowdays, workspace service is open to the cluster. It exposes all workspace ports and communication goes directly to target container, so anyone in the cluster can access it.

This patch adds traefik container into workspace pod and narrows the service to only single port (targeting to the traefik) for all endpoints that exposes on subpath. That means that all communication that is coming to the workspace service, must go through the traefik. The traefik is configured so it checks the requests with kube-rbac-proxy (in main che-gateway) using forwardAuth middleware (https://doc.traefik.io/traefik/middlewares/http/forwardauth/). Kube-rbac-proxy is checking whether token in authorization header can access the service in workspace's namespace. If this check is OK, traefik route the request to correct port inside the workspace pod.

https://github.com/eclipse/che/issues/20190

### How to test this PR?

1. deploy with che-operator image `quay.io/mvala/che-operator:gh20190-secureWsServicesTraefik`
2. start a workspace with user1 and get the service name
3. start a workspace with user2 and try curl to the service of user1's workspace. It is now not possible without user1's token in authorization header 

### PR Checklist

### Reviewers

### DIFF
--- a/controllers/devworkspace/solver/che_routing.go
+++ b/controllers/devworkspace/solver/che_routing.go
@@ -18,6 +18,9 @@ import (
 	"path"
 	"strings"
 
+	"github.com/eclipse-che/che-operator/pkg/deploy/gateway"
+	"k8s.io/apimachinery/pkg/util/intstr"
+
 	dw "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	dwo "github.com/devfile/devworkspace-operator/apis/controller/v1alpha1"
 	"github.com/devfile/devworkspace-operator/controllers/controller/devworkspacerouting/solvers"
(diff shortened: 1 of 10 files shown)


sparkoo:
I'm investigating CI failures. Some are probably real

controllers/devworkspace/solver/che_routing.go
```diff
@@ -208,212 +270,297 @@ func isSecureScheme(scheme string) bool {
...
+	prefix := fmt.Sprintf("/%s/%s", componentName, routeName)
+	rulePrefix := fmt.Sprintf("PathPrefix(`%s`)", prefix)
 
-			rtrs[name] = traefikConfigRouter{
-				Rule:        fmt.Sprintf("PathPrefix(`%s`)", prefix),
-				Service:     name,
-				Middlewares: calculateMiddlewares(name),
-				Priority:    100,
-			}
+	// skip if exact same route is already exposed
```
metlos, line 517: Under what circumstances does this happen? Can we eliminate the possibility sooner?

controllers/devworkspace/solver/che_routing.go
```diff
@@ -208,212 +270,297 @@ func isSecureScheme(scheme string) bool {
...
-							To:     "Authorization",
-							Prefix: "Bearer ",
-						},
-					},
-				}
-			}
+	if infrastructure.IsOpenShift() {
+		cfg.HTTP.Middlewares[name+"-auth"] = traefikConfigMiddleware{
+			ForwardAuth: &traefikConfigForwardAuth{
+				Address: fmt.Sprintf("http://%s.%s:8089?namespace=%s", gateway.GatewayServiceName, cheCluster.Namespace, routing.Namespace),
```
metlos, line 551: is https possible here? would prefer it otherwise potentially valid token can be sniffed from the unecrypted traffic.

metlos (commented):
I've not tested it in a running cluster yet, but here's a couple of small comments...

pkg/deploy/defaults.go
```diff
@@ -360,7 +360,7 @@ func DefaultPullPolicyFromDockerImage(dockerImage string) string {
 }
 
 func GetSingleHostExposureType(cr *orgv1.CheCluster) string {
-	if util.IsOpenShift {
+	if util.IsOpenShift || cr.Spec.DevWorkspace.Enable {
```
tolusha, line 363: Why do we need this changes? 
So, it means that single-host will be default for k8s when devworkspace is enabled.

pkg/deploy/gateway/gateway.go
```diff
@@ -56,8 +58,9 @@ var (
 
 // SyncGatewayToCluster installs or deletes the gateway based on the custom resource configuration
 func SyncGatewayToCluster(deployContext *deploy.DeployContext) error {
-	if util.GetServerExposureStrategy(deployContext.CheCluster) == "single-host" &&
-		(deploy.GetSingleHostExposureType(deployContext.CheCluster) == "gateway") {
+	if (util.GetServerExposureStrategy(deployContext.CheCluster) == "single-host" &&
```
tolusha, line 61: `a && b || c` let's be more clear here: `(a && b) || c`

pkg/deploy/defaults.go
metlos: Exactly. There was the agreement on the last cabal that we're doing that, if I remember correctly.

tolusha: Then `DefaultOpenShiftSingleHostExposureType` const has no sense anymore.
Let's remove it and just return `gateway`

controllers/devworkspace/solver/che_routing.go
sparkoo: multiple non-unique subpath endpoints. There's test for it https://github.com/eclipse-che/che-operator/blob/main/controllers/devworkspace/solver/che_routing_test.go#L152

skabashnyuk:
@sparkoo can you describe in the description how in general protection has been made.

controllers/devworkspace/solver/che_routing.go
sparkoo: In theory, yes we can eliminate sooner and it is how it's currently implemented. However, I think this is much more straightforward. It was really hard to do even small change without breaking it completely, because endpoints were organized to complicated maps by various properties with custom structs etc. So small change to one of those things lead to completely different result. This just accepts the original endpoint from devworkspace-controller and expose it as needed (or not if it is already exposed). That's why I ended up with doing this bigger refactoring.

pkg/deploy/defaults.go
sparkoo: there will be only `gateway` with devworkspaces. `native` does not really work with native auth. Everything we want to protect with oauth-proxy and kube-rbac-proxy has to go through the gateway.

controllers/devworkspace/solver/che_routing.go
sparkoo: We're on http since beginning of the gateway https://github.com/eclipse-che/che-operator/blob/main/controllers/devworkspace/solver/che_routing.go#L506 and even with che-server workspaces https://github.com/eclipse-che/che-server/blob/main/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/external/TraefikGatewayRouteConfigGenerator.java#L60

It's communication inside of the cluster, so I would assume it's safe, but maybe I'm too naive. We can definitely do it, but it would need more work as we need all gateway components to listen on https. I would do it in separate issue.

pkg/deploy/gateway/gateway.go
sparkoo: I don't follow, it's now written as `(a && b) || c` => `(single-host && gateway) || devworkspace`

controllers/devworkspace/solver/che_routing.go
```diff
@@ -100,42 +132,77 @@ func (c *CheRoutingSolver) cheSpecObjects(cheManager *v2alpha1.CheCluster, routi
...
+		ImagePullPolicy: corev1.PullAlways,
+		VolumeMounts: []corev1.VolumeMount{
+			{
+				Name:      wsGatewayName,
+				MountPath: "/etc/traefik",
+			},
+		},
+	})
+
+	defaultMode := int32(420)
```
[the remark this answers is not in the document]
sparkoo: added comment

controllers/devworkspace/solver/che_routing.go
```diff
@@ -208,212 +270,297 @@ func isSecureScheme(scheme string) bool {
...
 	}
+}
 
-	if len(config.HTTP.Routers) > 0 {
-		contents, err := yaml.Marshal(config)
-		if err != nil {
-			return []corev1.ConfigMap{}, err
+func getCommonService(objs *solvers.RoutingObjects, dwId string) *corev1.Service {
+	for i, svc := range objs.Services {
+		if svc.Name == common.ServiceName(dwId) {
```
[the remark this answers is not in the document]
sparkoo: fixed

sparkoo:
> @sparkoo can you describe in the description how in general protection has been made.

added "How it works" paragraph

tolusha (approved):

metlos (approved):

metlos:
Btw I've read the description again and am I correct in saying that the request is checked using kube-rbac-proxy also in the workspace pod (essentially again?). This is important so that we guard against in-cluster access by other users. I thought I saw it in the code but I understood your description otherwise. Could you please clarify?

sparkoo:
> Btw I've read the description again and am I correct in saying that the request is checked using kube-rbac-proxy also in the workspace pod (essentially again?). This is important so that we guard against in-cluster access by other users. I thought I saw it in the code but I understood your description otherwise. Could you please clarify?

Yes, it's now checked twice. First in main che-gateway https://github.com/eclipse-che/che-operator/pull/1045/files#diff-ebca2eefe12f7ba4a722c53d574ba1b2adee412909da8cdbc974c8f7fcbfb02fR468 then second time in workspace gateway https://github.com/eclipse-che/che-operator/pull/1045/files#diff-ebca2eefe12f7ba4a722c53d574ba1b2adee412909da8cdbc974c8f7fcbfb02fR554

sleshchenko (approved):
Nothing really meaningful I can bring after reviewing changes.
Works fine for user1/user2 scenario 
Screenshot_20210917_130846

but seems cluster admin is able to open any workspace, I assume it's expected but it should be made clear in the documentation, because it's a change from the previous auth model

PR #1324, opened 2022-02-17
feat: add conversion webhook

tolusha:

### What does this PR do?
- add conversion weebhook

N/A

https://github.com/eclipse/che/issues/20758

### How to test this PR?

N/A

### PR Checklist

### Reviewers


l0rd:
I did a review of https://github.com/eclipse-che/che-operator/blob/20758/api/v2/checluster_types.go. Here are a few comments:

--- 

> the Operator automatically creates and maintains several ConfigMaps
  
 Can we list them? 

---

At line 58 you should replace `displayName="Devfile registry"` with `displayName="Dashboard"`

---

At line 79 you should not you should remove ", the plugin and devfile registries"

---

What happens if `server.clusterRoles` is omitted? What are the default ClusterRoles? 

---

In `server.workspaceServiceAccountClusterRole` description it's mentioned "The default roles are used when omitted or left blank." What are the default roles?

---

In `server.WorkspaceNamespaceName` description: "for a case when a user does not override it.". I don't think a user can override it: users cannot select the workspace namespace anymore. I would rather say: "If users namespaces are not pre-created this field defines the Kubernetes namespace created when a user starts his first workspace....". And also what happens when omitted or left blank?

---

Maybe `server.serverTrustStoreConfigMapName` should be renamed `serverTrustStoreConfigMapName.javaTrustStoreConfigMapName`?

---

It would be better if instead of the boolean `server.gitSelfSignedCert` we had `server.gitServerTLSCertificateConfigMapName`

---

For `storage.pvcStrategy`: We should mention that currently only `common` is supported and add a link to the issue: https://github.com/eclipse/che/issues/21185.

---

Isn't `CheCluster.storage` only related to workspaces (and database.pvc is server side)? In that case we should probably make it explicit in the comments, call it `workspaceStorage` and probably move it under `CheCluster.server` since everything related to workspaces is there.

---

In `server.workspacesDefaultPlugins` workspace**s** is plural whereas other fields use the singular (`workspacePodNodeSelector` etc...). For consistency we should rather use singular everywhere.

---

Do we really need `database.postgresVersion`? What's the default by the way?

---

`server.workspacesDefaultPlugins` should be renamed `server.workspaceDefaultPluginsPerEditor`

---

How an admin can specify the default editor?

---

How an admin can specify the default container image (if for instance there is no devfile.yaml in a github repository)

---

All server components Deployment can be overridden using DeploymentCustomSettings except the Gateway (server side) and the 3 sidecar (workspace side) where only image can be overriden. This is inconsistent.

---

We should mention that when `gateway.ingress` is specified, `gateway.route` should not. And that `gateway.route` is for OpenShift only. It would be even better if both could be merged together so that we only have `gateway.ingress`

---

The sentence "The generated host name will follow this pattern: `<route-name>-<route-namespace>.<domain>`" is confusing because an admin cannot specify `<route-name>` and `<route-namespace>`. I would replace that with "The generated hostname will follow this pattern: `che-<che-namespace>.<domain>`, where `<che-namespace>` is the namespace where the `CheCluster` CRD is created".

---

What about renaming `route.host` and `ingress.host` as `route.hostname` and `ingress.hostname` respectively?

---

We can safely replace every instance of `host name` with `hostname`

---

Setting the hostname is a common task. With the new model it's hidden under the `auth` section: `auth>gateway>ingress>hostname`. I am wondering if we should not rather organize things like this: 
```
ingress
  > labels 
  > annotations 
  > domain 
  > hostname 
  > tlsSecret
  > auth
    > identityProviderURL
    > oAuthClientName
    > oAuthSecret
    > gateway
      > ...
```

---

In `CheClusterSpecMetrics` there is the following description: "/ Enables `metrics` the Che server endpoint. Default to `true`.". Doesn't it control DevWorkspace controller metrics too? If it's only for che-server it should be moved under `server` if it's for `DevWorkspace` it should be mentioned. And it shoudl be also mentioned that this is distinct from workspace telemetry that can be enabled through a plugin editor only.

---

The server.proxy is described as "// Proxy server settings" but then in the Proxy `struct`: This drives the appropriate changes in the `JAVA_OPTS` and `https(s)_proxy` variables in the Che server **and** workspaces containers. So are the proxy settings propagated to the che-server, to the workspaces or both? What about other server side components as the dashboard, the registries etc...?

---

`server.customCheProperties` should be `server.customCheServerProperties`

---

`server.debug` is a string but should be a boolean

---

`workspaceServiceAccountClusterRole` is described as "Custom cluster role bound to the user for the Che workspaces": is it bound to the user or is bound to the SA?

---

Finally I think it makes sense to have all the workspace related fields grouped under workspaces so I have imagined the following structure:
```
> workspaces
    > defaultPlugins
    > defaultNamespace
        > nameTemplate
        > userClusterRoles (or serviceAccountClusterRoles?)
    > storage
    > nodeSelector
    > tolerations
    > trustedCerts
        > trustStoreConfigMapName
        > gitTrustedCertsConfigMapName
> serverComponents
    > dashboard
    > cheServer
        > clusterRoles
        > deployment
        > loglevel
        > debug
    > database
    > pluginRegistry
    > devfileRegistry
    > imagePuller
    > devWorkspace
    > metrics?
> containerRegistry
    > hostname
    > organization
```

bundle/next/eclipse-che-preview-openshift/manifests/che-operator.clusterserviceversion.yaml
```diff
@@ -75,12 +99,127 @@ metadata:
...
+            name: ''
+            version: v1
+          - kind: Secret
+            name: ''
+            version: v1
+          - kind: Service
+            name: ''
+            version: v1
+        specDescriptors:
+          - description: Container registry configuration.
```
l0rd, line 149: Can we have it displayed in this order:
- workspaces
- serverComponents
- ingress
- containerRegistry

l0rd, line 149:
```suggestion
          - description: Configuration of an alternative registry that stores Che images.
```

bundle/next/eclipse-che-preview-openshift/manifests/che-operator.clusterserviceversion.yaml
```diff
@@ -75,12 +99,127 @@ metadata:
...
+            displayName: Container registry
+            path: containerRegistry
+          - description: DevWorkspace operator configuration
```
l0rd, line 152: The DevWorksapce should be in the server components section in my opinion

bundle/next/eclipse-che-preview-openshift/manifests/che-operator.clusterserviceversion.yaml
```diff
@@ -75,12 +99,127 @@ metadata:
...
+            displayName: Dev Workspace operator
+            path: devWorkspace
+          - description: Ingress configuration.
```
l0rd, line 155:
```suggestion
          - description: Che ingress hostname, authentication and TLS configuration.
```

bundle/next/eclipse-che-preview-openshift/manifests/che-operator.clusterserviceversion.yaml
```diff
@@ -75,12 +99,127 @@ metadata:
...
+            displayName: Ingress
+            path: ingress
+          - description: Server components configuration.
```
l0rd, line 158:
```suggestion
          - description: Che server components configuration.
```

bundle/next/eclipse-che-preview-openshift/manifests/che-operator.clusterserviceversion.yaml
```diff
@@ -75,12 +99,127 @@ metadata:
...
+            displayName: Server components
+            path: serverComponents
+          - description: Workspaces configuration.
```
l0rd, line 161:
```suggestion
          - description: Development environment default configuration options.
```

l0rd:
I don't know if that's related to this PR but testing it I noticed that the `Eclipse Che URL` is shown even if Che is not ready yet:

image

After a minute or so the application becomes available.

l0rd:
Another thing I have noticed: all che-server components deployed in `openshift-operators` namespace.

```
$ kubectl get deploy -n openshift-operators
NAME                              READY   UP-TO-DATE   AVAILABLE   AGE
che                               1/1     1            1           87m
che-dashboard                     1/1     1            1           87m
che-gateway                       1/1     1            1           87m
che-operator                      1/1     1            1           168m
devfile-registry                  1/1     1            1           88m
devworkspace-controller-manager   1/1     1            1           168m
devworkspace-webhook-server       1/1     1            1           167m
```

bundle/next/eclipse-che-preview-openshift/manifests/che-operator.clusterserviceversion.yaml
tolusha: Yes we can.

tolusha:
/test v9-devworkspace-happy-path
--


<br class="Apple-interchange-newline">

deploy/deployment/openshift/objects/che-operator.ClusterRoleBinding.yaml
```diff
@@ -0,0 +1,29 @@
+#
```
mmorhun, line 1: Does it makes sense to use kustomize to not to duplicate for Kubernetes and Openshift?

mmorhun (approved):
My review is very shallow. Just a few thought on the topic (feel free to ignore).
You've done really great job!

nickboldt:
This change has a downstream impact on https://issues.redhat.com/browse/CRW-2949 and https://issues.redhat.com/browse/RHDEVDOCS-4043

ibuziuk (commented):
I have some concerns about the `CheClusterServerComponents` semantics:

- the naming mimics and also includes the CheServer component which makes it a bit misleading
- the struct includes the Dashboard component that is actually meant to be a client application initially

I would propose to make the naming more generic to drop the 'Server' part and stick to 'CheClusterComponents'.

https://github.com/eclipse-che/che-operator/blob/3815a248cca8be70c173814469e0a3e977fb8899/api/v2/checluster_types.go#L42.

api/v2/conversion.go
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2019-2021 Red Hat, Inc.
```
amisevsk, line 2: I believe the usual naming format for this file is `<crd>_conversion.go`, e.g. `checluster_conversion.go`, in order to match `checluster_types.go` and `checluster_webhook.go`.

I also can't easily find the similar `conversion.go` bits in the v1 apiVersion

bundle/next/eclipse-che-preview-openshift/manifests/che-operator.clusterserviceversion.yaml
```diff
@@ -75,12 +103,125 @@ metadata:
...
+            name: ''
+            version: v1
+          - kind: Service
+            name: ''
+            version: v1
+        specDescriptors:
+          - description: Development environment default configuration options
+            displayName: Workspaces
+            path: workspaces
+          - description: Che operands configuration
```
ibuziuk, line 156: @tolusha @l0rd was thinking about the `operands` and technically not all the containg properties are operands e.g. `devWorkspace`, `imagePuller`, `extraProperties`, `metrics`
maybe it would be better to stick to nutaral `components` from the semantic perspective?

deploy/deployment/openshift/org_v2_checluster.yaml
```diff
@@ -0,0 +1,38 @@
...
+metadata:
+  name: eclipse-che
+  namespace: eclipse-che
+spec:
+  workspaces:
+    defaultNamespace:
+      template: '<username>-che'
+    storage:
+      pvcStrategy: 'common'
+      preCreateSubPaths: true
```
ibuziuk, line 24: @amisevsk @l0rd Wondering if we still need `preCreateSubPaths` property? as I recall it was needed only for the old version of k8s / openshift

bundle/next/eclipse-che-preview-openshift/manifests/che-operator.clusterserviceversion.yaml
tolusha: `components` works for me

l0rd: Ok let's go with `components` then.

I got 2 feedbacks: 
1) `networking` would be better than `ingress`
2) `workspaces` is an overloaded term, maybe `devEnvironments` or `remoteEnvironments` would be better

What do you think?

deploy/deployment/openshift/org_v2_checluster.yaml
amisevsk: By default, DWO has logic to work around the same issue to this day. However, now that PVC provisioning is managed by DWO, perhaps this option is no longer relevant to the CheCluster CR.

tolusha: Thank you. I will remove it then.

tolusha: But what about `PvcJobsImage` then? Does somebody use it?
Description says:
```
	// Overrides the container image used to create sub-paths in the Persistent Volumes.
	// This includes the image tag. Omit it or leave it empty to use the default container image provided by the Operator.
```

amisevsk: I think it makes sense to drop as well -- DWO handles PVC jobs and has its own related image for this, so this field will do nothing currently. If one day there's a mechanism within Che to configure DWO, then it might be useful then, but we can always add it at that point.

For the CheCluster API, I would err on the side of fewer fields, as adding fields is backwards compatible but removing them is not.

config/samples/org_v2_checluster.yaml
```diff
@@ -0,0 +1,37 @@
...
+  name: eclipse-che
+  namespace: eclipse-che
+spec:
+  workspaces:
+    defaultNamespace:
+      template: '<username>-che'
+    storage:
+      pvcStrategy: 'common'
+      pvc:
+        claimSize: '10Gi'
```
amisevsk, line 25: For DWO workspaces, I believe this field is currently a no-op -- DWO has its own configuration for PVC size so unless Che is configuring DWO this will generally be ignored.

I've created https://github.com/devfile/devworkspace-operator/issues/843 to maybe enable something like this in the future.

api/v2/checluster_types.go
```diff
@@ -0,0 +1,639 @@
...
+	Spec imagepullerv1alpha1.KubernetesImagePullerSpec `json:"spec"`
+}
+
+// Settings for installation and configuration of the DevWorkspace operator
+// See https://github.com/devfile/devworkspace-operator
+// +k8s:openapi-gen=true
+type DevWorkspace struct {
+	// Deployment override options.
+	// +optional
+	Deployment Deployment `json:"deployment,omitempty"`
```
amisevsk, line 298: Is this meant to override the DevWorkspace Operator deployment itself? Does the Che Operator support this field currently?

api/v2/checluster_types.go
```diff
@@ -0,0 +1,639 @@
...
+	// Persistent volume claim strategy for the Che server.
+	// The only `common` strategy is supported (all workspaces PVCs in one volume).
+	// See for details https://github.com/eclipse/che/issues/21185.
+	// +optional
+	// +kubebuilder:default:="common"
+	PvcStrategy string `json:"pvcStrategy,omitempty"`
+	// Overrides the container image used to create sub-paths in the Persistent Volumes.
+	// This includes the image tag. Omit it or leave it empty to use the default container image provided by the Operator.
+	// +optional
+	PvcJobsImage string `json:"pvcJobsImage,omitempty"`
```
amisevsk, line 341: Creating subpaths is handled by DWO now, this should be safely removable.

api/v2/checluster_types.go
```diff
@@ -0,0 +1,639 @@
...
+	// Note, this ConfigMap must have `app.kubernetes.io/part-of=che.eclipse.org` label.
+	// +optional
+	GitTrustedCertsConfigMapName string `json:"gitTrustedCertsConfigMapName,omitempty"`
+}
+
+// Configuration settings related to the workspaces persistent storage.
+type WorkspaceStorage struct {
+	// PVC settings.
+	// +optional
+	Pvc PVC `json:"pvc,omitempty"`
```
amisevsk, line 331: If this is for the `claim-che-workspace` PVC in each user's namespace, it's unclear how easy this will be to support. DWO creates PVCs, but will use a `claim-che-workspace` PVC if it is present, so Che could still do this, but I don't know if the functionality exists currently.

api/v2/checluster_types.go
```diff
@@ -0,0 +1,639 @@
...
+}
+
+// List of resources.
+type ResourceList struct {
+	// Memory, in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
+	// +optional
+	Memory string `json:"memory,omitempty"`
+	// CPU, in cores. (500m = .5 cores)
+	// +optional
+	Cpu string `json:"cpu,omitempty"`
```
amisevsk, line 473: These could be specified as `resource.Quantity` to get the standard k8s validation automatically and avoid the need to parse them in code.

```go
import "k8s.io/apimachinery/pkg/api/resource"
//...
	// Memory, in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
	// +optional
	Memory resource.Quantity `json:"memory,omitempty"`
	// CPU, in cores. (500m = .5 cores)
	// +optional
	Cpu resource.Quantity `json:"cpu,omitempty"`
```

amisevsk:
I believe the approach for the initial v2 CRD version should be to include only fields that are directly supported/handled now unless there's a good reason to have them. For CRDs, we can add fields in a future release and that is treated as a backward compatible change, but we cannot remove fields. This means that if there's a field in the CRD that is currently unused, and we decide in the future to support it in a different way, we'll still be stuck with it until v3.

api/v2/checluster_types.go
tolusha: It is used only when DevWorkspace is deployed by templates (not olm mode)

l0rd:
I had provided this comment https://github.com/eclipse-che/che-operator/pull/1324#discussion_r876415916. I think that `networking` is better than ingress and the term `workspaces` is used in [other (Red Hat) tools](https://coreos.slack.com/archives/C011LSD1R4M/p1652744640749659) with a different meaning so that can be a good opportunity to change that and start calling them `devEnvironments`. cc @ibuziuk @themr0c

tolusha:
@l0rd 
Thank you for reminding. I completely forgot about that comment. ^(

ibuziuk (approved):
@l0rd `devEnvironments` / `networking` LGTM
no more suggestions from my end

api/checluster_conversion_to_test.go
```diff
@@ -0,0 +1,412 @@
...
+	assert.Nil(t, err)
+
+	assert.Equal(t, map[string]string{"a": "b", "c": "d", "kubernetes.io/ingress.class": "k8s.IngressClass"}, checlusterv2.Spec.Networking.Annotations)
+	assert.Equal(t, "k8s.IngressDomain", checlusterv2.Spec.Networking.Domain)
+	assert.Equal(t, "CheHost", checlusterv2.Spec.Networking.Hostname)
+	assert.Equal(t, map[string]string{"a": "b", "c": "d"}, checlusterv2.Spec.Networking.Labels)
+	assert.Equal(t, "CheHostTLSSecret", checlusterv2.Spec.Networking.TlsSecretName)
+}
+
+func TestConvertTo(t *testing.T) {
```
amisevsk, line 144: It may also be worth including a test cases for
1. Round trip conversion (convert v1 -> v2 -> v1 and assert that the before and after are DeepEqual) -- this is the underlying k8s API requirement for conversion
2. Fuzzed objects with round-trip conversion, to ensure e.g. a field added to v2 isn't dropped if that CR is converted v2 -> v1 -> v2

api/v2/checluster_types.go
```diff
@@ -0,0 +1,637 @@
...
+	ImagePuller ImagePuller `json:"imagePuller"`
+	// Che server metrics configuration.
+	// +optional
+	Metrics ServerMetrics `json:"metrics"`
+}
+
+// Configuration settings related to the Networking used by the Che installation.
+// +k8s:openapi-gen=true
+type CheClusterSpecNetworking struct {
+	// Comma separated list of labels that can be used to organize and categorize objects by scoping and selecting.
```
amisevsk, line 111: Doc looks to be out of date -- says "comma separated" but field is `map[string]string`

api/v2/checluster_types.go
```diff
@@ -0,0 +1,637 @@
...
+	// - `<che-namespace>-cheworkspaces-clusterrole`
+	// - `<che-namespace>-cheworkspaces-devworkspace-clusterrole`
+	// where <che-namespace> is the namespace where the CheCluster CRD is created.
+	// Each role must have `app.kubernetes.io/part-of=che.eclipse.org` label.
+	// Be aware that the Che Operator has to already have all permissions in these ClusterRoles to grant them.
+	// +optional
+	ClusterRoles []string `json:"clusterRoles,omitempty"`
+	// Proxy server settings.
+	// +optional
+	Proxy Proxy `json:"proxy"`
```
amisevsk, line 181: Consider specifying how this field should be used -- the Operator will automatically read proxy configs on OpenShift.

api/v2/checluster_types.go
```diff
@@ -0,0 +1,637 @@
...
+}
+
+// Configuration settings related to the database used by the Che installation.
+// +k8s:openapi-gen=true
+type Database struct {
+	// Instructs the Operator on whether to deploy a dedicated database.
+	// By default, a dedicated PostgreSQL database is deployed as part of the Che installation. When `externalDb` is `true`, no dedicated database will be deployed by the
+	// Operator and you will need to provide connection details to the external DB you are about to use. See also all the fields starting with: `chePostgres`.
+	// +optional
+	ExternalDb bool `json:"externalDb"`
```
amisevsk, line 240:
```suggestion
	// Instructs the Operator on whether to deploy a dedicated database.
	// By default, a dedicated PostgreSQL database is deployed as part of the Che installation. When `externalDb` is `true`, no dedicated database will be deployed by the
	// Operator and you will need to provide connection details to the external DB you are about to use. See also all the fields starting with: `postgres`.
	// +optional
	ExternalDb bool `json:"externalDb"`
```

api/v2/checluster_types.go
```diff
@@ -0,0 +1,637 @@
...
+// Settings for installation and configuration of the DevWorkspace operator
+// See https://github.com/devfile/devworkspace-operator
+// +k8s:openapi-gen=true
+type DevWorkspace struct {
+	// Deployment override options.
+	// +optional
+	Deployment Deployment `json:"deployment,omitempty"`
+	// Maximum number of the running workspaces per user.
+	// +optional
+	RunningLimit string `json:"runningLimit,omitempty"`
```
amisevsk, line 303: This might make more sense as a dashboard option (rather than DevWorkspace). Currently, the dashboard is what imposes the limit on running workspaces, and DWO does not have functionality to do this.

api/v2/checluster_types.go
```diff
@@ -0,0 +1,637 @@
...
+}
+
+type DefaultNamespace struct {
+	// if users namespaces are not pre-created this field defines the Kubernetes namespace created when a user starts his first workspace.
```
amisevsk, line 307:
```suggestion
	// if users namespaces are not pre-created this field defines the Kubernetes namespace created when a user starts their first workspace.
```

api/v2/checluster_types.go
```diff
@@ -0,0 +1,637 @@
...
+	// The only `common` strategy is supported (all workspaces PVCs in one volume).
+	// See for details https://github.com/eclipse/che/issues/21185.
+	// +optional
+	// +kubebuilder:default:="common"
+	PvcStrategy string `json:"pvcStrategy,omitempty"`
+}
+
+type WorkspaceDefaultPlugins struct {
+	// The editor id to specify default plug-ins for.
+	Editor string `json:"editor,omitempty"`
```
amisevsk, line 344: Is this only an ID or can it also be a URI?

api/v2/checluster_types.go
```diff
@@ -0,0 +1,637 @@
...
+// of the various components of the installation. These generated ConfigMaps must NOT be updated manually.
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +k8s:openapi-gen=true
+// +operator-sdk:csv:customresourcedefinitions:displayName="Eclipse Che instance Specification"
+// +operator-sdk:csv:customresourcedefinitions:order=0
+// +operator-sdk:csv:customresourcedefinitions:resources={{Ingress,v1},{Route,v1},{ConfigMap,v1},{Service,v1},{Secret,v1},{Deployment,apps/v1},{Role,v1},{RoleBinding,v1},{ClusterRole,v1},{ClusterRoleBinding,v1}}
+// +kubebuilder:storageversion
+type CheCluster struct {
```
amisevsk, line 581: I might have missed it in reading this file, but consider adding `+kubebuilder:printcolumn:name` annotations here to automatically print the dashboard URL, etc. when listing objects on the cluster. For example, something like 
```
// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.cheClusterRunning",description="Status for the CheCluster"
// +kubebuilder:printcolumn:name="Dashboard URL",type="string",JSONPath=".status.cheURL",description="The URL for the Che Dashboard"
```
If this is handled by the `+operator-sdk` annotations, then ignore this comment.

amisevsk (approved):
Approving to not block merge, a few comments below but generally looks good and comments can be addressed in a separate PR if need be.

api/v2/checluster_types.go
amisevsk: An example of how this is used in the DevWorkspace CR: [[link]](https://github.com/devfile/api/blob/d6e86811be170f92c70a15fbfea79028866448a9/pkg/apis/workspaces/v1alpha2/devworkspace_types.go#L70-L78)

l0rd (approved):

api/v2/checluster_types.go
tolusha: I think it is covered by +operator-sdk annotations

api/v2/checluster_types.go
```diff
@@ -0,0 +1,638 @@
...
+	HeaderMessage DashboardHeaderMessage `json:"HeaderMessage,omitempty"`
+}
+
+// Configuration settings related to the Plugin Registry used by the Che installation.
+// +k8s:openapi-gen=true
+type PluginRegistry struct {
+	// Deployment override options.
+	// +optional
+	Deployment Deployment `json:"deployment,omitempty"`
+	// Disables internal Plugin registry.
```
deerskindoll, line 208:
```suggestion
	// Disables internal plug-in registry.
```

api/v2/checluster_types.go
```diff
@@ -0,0 +1,638 @@
...
+// See https://github.com/che-incubator/kubernetes-image-puller-operator
+// +k8s:openapi-gen=true
+type ImagePuller struct {
+	// Install and configure the Community Supported Kubernetes Image Puller Operator. When set to `true` and no spec is provided,
+	// it will create a default KubernetesImagePuller object to be managed by the Operator.
+	// When set to `false`, the KubernetesImagePuller object will be deleted, and the Operator will be uninstalled,
+	// regardless of whether a spec is provided.
+	// If the `spec.images` field is empty, a set of recommended workspace-related images will be automatically detected and
+	// pre-pulled after installation.
+	// Note that while this Operator and its behavior is community-supported, its payload may be commercially-supported
```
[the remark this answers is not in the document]
tolusha: cc @ibuziuk

tolusha:
@deerskindoll 
Thank you for review!

PR #1467, opened 2022-08-03
fix: Improve a way to check if DevWorkspace resources must be managed…

tolusha:
… by Che Operator

### What does this PR do?
Checks if Che Operator or Dev Workspace operator deployment has an owner.
It means these resources are managed by someone else (OLM). 
So, Dev Workspace resources mustn't be managed by Che Operator as well.

N/A

https://github.com/eclipse/che/issues/21603

### How to test this PR?

Catalog source image to test: `docker.io/abazko/catalog:21603`

### PR Checklist

- [ ] [Custom resource definition file is up to date](https://github.com/eclipse-che/che-operator/blob/main/README.md#updating-custom-resource-definition-file)
- [ ] [OLM bundles are up to date](https://github.com/eclipse-che/che-operator/blob/main/README.md#update-olm-bundle)

### Reviewers

### DIFF
--- a/pkg/deploy/dev-workspace/dev_workspace.go
+++ b/pkg/deploy/dev-workspace/dev_workspace.go
@@ -62,27 +62,19 @@ func NewDevWorkspaceReconciler() *DevWorkspaceReconciler {
 
 func (d *DevWorkspaceReconciler) Reconcile(ctx *chetypes.DeployContext) (reconcile.Result, bool, error) {
 	if infrastructure.IsOpenShift() {
-		// Do nothing if Eclipse Che operator installed by OLM
-		// In this case DevWorkspace operator should be managed by OLM as well
-		cheOperatorInstalledByOLM, err := isCheOperatorInstalledByOLM(ctx)
-		if cheOperatorInstalledByOLM {
+		// Do nothing if Che operator has owner.
+		// In this case DevWorkspace operator resources mustn't be managed by Che operator.
+		cheOperatorHasOwner, err := isCheOperatorHasOwner(ctx)
+		if cheOperatorHasOwner {
 			return reconcile.Result{}, true, nil
 		} else if err != nil {
 			return reconcile.Result{Requeue: true}, false, err
 		}
 
-		// Do nothing if Dev Workspace operator installed by OLM
-		devWorkspaceInstalledByOLM, err := isDevWorkspaceOperatorInstalledByOLM(ctx)
-		if devWorkspaceInstalledByOLM {
-			return reconcile.Result{}, true, nil
-		} else if err != nil {
-			return reconcile.Result{Requeue: true}, false, err
-		}
-
-		// Do nothing if WTO exists since it should bring or embeds DWO
-		// TODO check if it is needed
-		wtoInstalledByOlm, err := isWebTerminalInstalledByOlm(ctx)
-		if wtoInstalledByOlm {
+		// Do nothing if Dev Workspace operator has owner.
+		// In this case DevWorkspace operator resources mustn't be managed by Che operator.
+		devWorkspaceOperatorHasOwner, err := isDevWorkspaceOperatorHasOwner(ctx)
+		if devWorkspaceOperatorHasOwner {
 			return reconcile.Result{}, true, nil
 		} else if err != nil {
 			return reconcile.Result{Requeue: true}, false, err
--- a/pkg/deploy/dev-workspace/dev_workspace.go
+++ b/pkg/deploy/dev-workspace/dev_workspace.go
@@ -37,7 +37,6 @@ const (
 
 	SubscriptionResourceName          = "subscriptions"
 	ClusterServiceVersionResourceName = "clusterserviceversions"
-	DevWorkspaceCSVNamePrefix         = "devworkspace-operator"
 
 	WebTerminalOperatorSubscriptionName = "web-terminal"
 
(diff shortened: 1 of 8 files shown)


pkg/deploy/dev-workspace/dev_workspace.go
```diff
@@ -62,19 +61,23 @@ func NewDevWorkspaceReconciler() *DevWorkspaceReconciler {
 }
 
 func (d *DevWorkspaceReconciler) Reconcile(ctx *chetypes.DeployContext) (reconcile.Result, bool, error) {
-	if isDevWorkspaceOperatorCSVExists(ctx) {
```
ibuziuk, line 65: I was thinking that as a workaround we would improve this check e.g. 
`ifCSVExists || devworkspaceDeploymentExists { Do nothing }`

pkg/deploy/dev-workspace/dev_workspace.go
```diff
@@ -62,19 +61,23 @@ func NewDevWorkspaceReconciler() *DevWorkspaceReconciler {
...
-		// Do nothing if DevWorkspace has been already deployed via OLM
-		return reconcile.Result{}, true, nil
-	}
-
 	if infrastructure.IsOpenShift() {
-		wtoInstalled, err := isWebTerminalSubscriptionExist(ctx)
-		if err != nil {
```
ibuziuk, line 72: So, we do not care about WTO anymore from Che / DevSpaces operator pov? It is actually interesting that on Developer Sandbox we had both DWO and WTO installed in the `crw` namespace  so looks like both isDevWorkspaceOperatorCSVExists and isWebTerminalSubscriptionExist returned false

tolusha: We don't need WTO check anymore, since if WTO installed then DWO is installed as well. And we already have improved DWO check. 

`isWebTerminalSubscriptionExist` returned false because it checked `openshift-operators` namespace for existed WTO subscription but not `crw` namespace.

pkg/deploy/dev-workspace/dev_workspace.go
tolusha: `ifCSVExists` is not robust, so I've got rid of it.
And I added `isDevWorkspaceOperatorHasOwner` func to check if DWO deployment exists and managed by OLM.

tolusha:
@ibuziuk 
Updated

ibuziuk (commented):
in the process of PR verification, but with this change does it mean that when the DevSpaces operator is installed via OperatorHub it will not install DWO even if it does not exist on the cluster?

tolusha:
@ibuziuk 
Exactly, DWO must be be installed as OLM dependency or admin can install it via yaml templates. 
But in both cases che operator won't update DWO resources.

ibuziuk:
Awesome, so on a fresh OpenShift 4.x cluster when Eclipse Che / Dev Spaces is installed by admin via OperatorHub, DWO will be installed automatically by OLM since we have it as a dependency [1], but the operator itself would not install it ever now :+1: 

[1] https://github.com/eclipse-che/che-operator/blob/main/bundle/next/eclipse-che-preview-openshift/metadata/dependencies.yaml

pkg/deploy/dev-workspace/dev_workspace.go
```diff
@@ -62,19 +54,22 @@ func NewDevWorkspaceReconciler() *DevWorkspaceReconciler {
...
-			// Do nothing if WTO exists since it should bring or embeds DWO
+
+		// Do nothing if Dev Workspace operator has owner (installed via OLM).
+		// In this case Dev Workspace operator resources mustn't be managed by Che operator.
+		isDevWorkspaceOperatorHasOwner, err := isDevWorkspaceOperatorHasOwner(ctx)
+		if isDevWorkspaceOperatorHasOwner {
 			return reconcile.Result{}, true, nil
+		} else if err != nil {
+			return reconcile.Result{Requeue: true}, false, err
 		}
```
ibuziuk, line 73: Do we really care now about the owner?

ibuziuk (approved):
Verified against 7.52.0-643.next from docker.io/abazko/catalog:21603 
DWO is installed and managed by olm during the EclipseChe installation:

image

pkg/deploy/dev-workspace/dev_workspace.go
tolusha: We need this for the following case:
1. OpenShift cluster with preinstalled DevWorkspace operator from OperatorHub
2. Eclipse Che operator is deployed via yaml templates (for instance for testing purpose), so `NO_OPT_DWO=false`

so, che operator must not managed DWO resources here.

ibuziuk: 1. OpenShift cluster with preinstalled DevWorkspace operator from OperatorHub

but in this case, it will still be managed by olm, no? My understanding is that olm should discover this instance during the Eclipse Che / Devspaces installation

2. Eclipse Che operator is deployed via yaml templates (for instance for testing purpose), so NO_OPT_DWO=false

This is actually an interesting part, In this case wouldn't olm try to install it as a dependency again?

tolusha: 1. Yes DWO will be still managed by olm. 
2. Since che operator is installed via yaml templates, there is no olm bundle and there is no any dependencies.

l0rd (approved):
👍

pkg/deploy/dev-workspace/dev_workspace.go
```diff
@@ -62,19 +54,22 @@ func NewDevWorkspaceReconciler() *DevWorkspaceReconciler {
...
+			return reconcile.Result{}, true, nil
+		} else if err != nil {
 			return reconcile.Result{Requeue: true}, false, err
 		}
-		if wtoInstalled {
-			// Do nothing if WTO exists since it should bring or embeds DWO
+
+		// Do nothing if Dev Workspace operator has owner (installed via OLM).
+		// In this case Dev Workspace operator resources mustn't be managed by Che operator.
+		isDevWorkspaceOperatorHasOwner, err := isDevWorkspaceOperatorHasOwner(ctx)
```
amisevsk, line 68: I worry that down the line we may run into similar issues in non-OLM installs if 

1. I manually install DWO to namespace A
2. I install Che to namespace B

From reading the code, it looks like what will happen in this case is that Che find the DWO deployment, see it doesn't have an owner, and try to deploy/update DWO. However, since we're not looking for an install namespace, there's a chance Che will attempt to deploy an second instance of the operator, breaking workspace reconcile. 

This PR does fix the main issue we're facing though, so :+1: to the solution.

amisevsk (commented):
LGTM, one comment for future refinement.

pkg/deploy/dev-workspace/dev_workspace.go
tolusha: Yes, your are right. 
There is one assumption behind this.
DWO is always installed into `devworkspace-controller` namespace for non-OLM installs [1]

[1] https://github.com/devfile/devworkspace-operator/blob/main/deploy/deployment/openshift/objects/devworkspace-controller-manager.Deployment.yaml#L8

PR #1520, opened 2022-09-21
feat: Move handling DevWorkspace operator resources out of Che Operator

tolusha:
### What does this PR do?
feat: Move handling DevWorkspace operator resources out of Che Operator

N/A

https://github.com/eclipse/che/issues/21710

### How to test this PR?

N/A

### PR Checklist

### Reviewers


api/v2/checluster_types.go
```diff
@@ -337,9 +337,6 @@ type ImagePuller struct {
 // See https://github.com/devfile/devworkspace-operator
 // +k8s:openapi-gen=true
 type DevWorkspace struct {
-	// Deployment override options.
-	// +optional
-	Deployment *Deployment `json:"deployment,omitempty"`
```
amisevsk, line 342: To double check -- this field is intended only for use on Kubernetes, right? 

We had a discussion earlier about removing CRD fields and whether OLM would permit the change. We figured it was generally fine at the time but in further testing I've found that OLM blocks operator upgrades if the new CRD would render any existing CR invalid, meaning that if anyone has an OLM install and has set this field, we would need to document that this field has to be dropped before the upgrade can take place.

tolusha: Yes, it was just for Kubernetes.
I see, do I unserstand correctly that you suggest keeping this field for a while?

amisevsk: I'm not sure the best way forward -- probably the best approach is to just remove it, assuming not many users have configured the field in their CheCluster CRs. If the field isn't set in the CheCluster, then the update can proceed with no issues. The bigger issue is OLM where the installPlan would fail, requiring deleting + reinstalling the Operator, but that's extremely unlikely since it's a field meant for Kubernetes.

If we do get reports on failing to upgrade in the future, we should keep this in mind as a potential cause. The fix is relatively simple (delete the field before attempting the install).

build/scripts/minikube-tests/common.sh
```diff
@@ -45,19 +45,29 @@ initTemplates() {
...
   LAST_OPERATOR_VERSION_CLONE_PATH=${OPERATOR_REPO}/tmp/${LAST_PACKAGE_VERSION}
-  git clone --depth 1 --branch ${LAST_PACKAGE_VERSION} https://github.com/eclipse-che/che-operator/ ${LAST_OPERATOR_VERSION_CLONE_PATH}
+  git clone --quiet --depth 1 --branch ${LAST_PACKAGE_VERSION} https://github.com/eclipse-che/che-operator/ ${LAST_OPERATOR_VERSION_CLONE_PATH}
+
+  git clone --quiet https://github.com/devfile/devworkspace-operator ${OPERATOR_REPO}/tmp/dwo
+
+  local xFlag="+x"; [[ $- =~ x ]] && xFlag="-x"
+  set +x # suppress output
+  pushd ${OPERATOR_REPO}/tmp/dwo
+  DWO_STABLE_VERSION=$(git describe --tags $(git rev-list --tags) | sort --version-sort | tail -1)
```
amisevsk, line 58: Executing this command in a freshly-cloned DWO repo has the error log

```
❯ git describe --tags $(git rev-list --tags) | sort --version-sort | tail -1     
fatal: No tags can describe '4b1a86952b6493f5541fa4a93aee4366d66d26fc'.
Try --always, or create some tags.
v0.16.0
```

I tried `git tag | sort -V` but the newest tag there is `v1.0.0-alpha1` from two years ago. However, I think another way to get the most recent tag is `git describe --tags $(git rev-list --tags --max-count=1)`

config/rbac/role_binding.yaml
```diff
@@ -14,6 +14,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: che-operator
+  namespace: eclipse-che
```
amisevsk, line 17: Is there a reason to start including the namespace in these resources?

Also I'm not clear whether the leader-election role/rolebindings are required for this PR or if they're just updates.

amisevsk (approved):

build/scripts/minikube-tests/common.sh
tolusha: Thank you, make sense to me.

config/rbac/role_binding.yaml
tolusha: 1. No reason, I just wanted to explicitly specify namespace into resources.

2. Initially DWO roles were included into Che Operator ones [1]. Since I removed them, I had to add missing ones.

[1] https://github.com/eclipse-che/che-operator/pull/1520/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L376-L423

PR #1687, opened 2023-05-25
Add fields for configuring project clone container to CheCluster CRD

amisevsk:
### What does this PR do?
Adds fields to `checluster.spec.devEnvironments`:
```yaml
spec:
  devEnvironments:
    projectCloneContainer:
      name: <unused name field from container struct>
      image: <image used for project clone>
      imagePullPolicy: <image pull policy for project clone>
      env:
      - name: <env var name>
        value: <env var value
      resources:
        limits:
          cpu: <cpuLimit>
          memory: <memoryLimit>
        request:
          cpu: <cpuRequest>
          memory: <memoryRequest>
```
The `name` field is left over from reusing the existing Container struct as described in https://github.com/eclipse/che/issues/22036 and is consistent with `.spec.devEnvironments.gatewayContainer` added in https://github.com/eclipse-che/che-operator/pull/1639.

N/A

Project clone part of https://github.com/eclipse/che/issues/22036

### How to test this PR?
Deploy changes (including updated CRDs) and create a CheCluster with spec 
```yaml
spec:
  devEnvironments:
    projectCloneContainer:
      image: test-image
      imagePullPolicy: IfNotPresent
      env: 
      - name: test-env
        value: test-val
      resources:
        limits:
          memory: 1234Mi
          cpu: 1234m
        request:
          memory: 123Mi
          cpu: 123m
```
and verify that 
1. Fields are propagated into the DevWorkspaceOperatorConfig created by Che (the image used in the sample above will cause the workspace to fail to start)
2. When starting a workspace, fields defined in the CheCluster are used for the project clone init container
3. Editing/removing fields removes them from the DevWorkspaceOperatorConfig
### PR Checklist

- [ ] [Development resource are up to date](https://github.com/eclipse-che/che-operator/blob/main/README.md#update-development-resources)

### Reviewers


pkg/deploy/dev-workspace-config/dev_workspace_config_test.go
```diff
@@ -1328,3 +1409,194 @@ func TestReconcileDevWorkspaceConfigDeploymentStrategy(t *testing.T) {
...
+	testMemLimit := resource.MustParse("2Gi")
+	testCpuLimit := resource.MustParse("1000m")
+	testMemRequest := resource.MustParse("1Gi")
+	testCpuRequest := resource.MustParse("500m")
+
+	type testCase struct {
+		name                       string
+		cheProjectCloneConfig      *chev2.Container
+		expectedDevWorkspaceConfig *controllerv1alpha1.ProjectCloneConfig
+		existingDevWorkspaceConfig *controllerv1alpha1.ProjectCloneConfig
```
AObuchow, line 1424: This is a nitpick and very subjective, but it might be worth renaming these two fields to something like `expectedDWProjectCloneConfig` and `existingDWProjectCloneConfig`.

I got a bit confused later on in the test when you did a [diff check](https://github.com/eclipse-che/che-operator/pull/1687/commits/e7159d0d30f915a14c354b979c65133f5a367cd8#diff-698b095064585b533923485dccc65654810ebbf08649fc81eb10ec4c99a7cc77R1597):

`diff := cmp.Diff(testCase.expectedDevWorkspaceConfig, dwoc.Config.Workspace.ProjectCloneConfig) `

At a glance, it looks like we're comparing the DWOC and the project clone config.

tolusha (approved):

pkg/deploy/dev-workspace-config/dev_workspace_config.go
```diff
@@ -146,19 +148,60 @@ func updateWorkspaceStorageConfig(devEnvironments *chev2.CheClusterDevEnvironmen
...
+	}
+	container := devEnvironments.ProjectCloneContainer
+
+	workspaceConfig.ProjectCloneConfig.Image = container.Image
+	workspaceConfig.ProjectCloneConfig.ImagePullPolicy = container.ImagePullPolicy
+	workspaceConfig.ProjectCloneConfig.Env = container.Env
+	workspaceConfig.ProjectCloneConfig.Resources = cheResourcesToCoreV1Resources(container.Resources)
+}
+
+func cheResourcesToCoreV1Resources(resources *chev2.ResourceRequirements) *corev1.ResourceRequirements {
```
tolusha, line 182: I have some concerns:
* No value (`nil`) in `resources` should lead to setting defaults
* Zero value (`"0"`) should not lead to setting any limits/requests

amisevsk: I believe that's the current behavior here, though maybe indirectly. Since DWO controls the project-clone container, it also holds the default values and handling of zero values -- implementing this logic here would work against us:
* Che Operator does not know defaults for project-clone, and shouldn't set them
* If we convert `"0"` into "do not set field here", it will lead to the DWOC not having the field set, which will in turn lead to DWO using the default value

Basically, DWO and Che implement similar logic for handling resources overrides, and so on the Che Operator side, we need to just pass all fields from the CheCluster into the DevWorkspaceOperatorConfig, unchanged (which requires copying fields in Go since the structs come from different packages).

I think maybe it would be clearer if the function had a better name -- this is intended mostly for converting from the `chev2` struct to the `corev1` struct directly to allow passing it (unmodified) to the DevWorkspace Operator.

tolusha: > I believe that's the current behavior here, though maybe indirectly. Since DWO controls the project-clone container, it also holds the default values and handling of zero value

Yes, I found it [1], not sure about settings defaults. Any way, I think it make sense to put some comments here:
`Since DWO controls the project-clone container, it also holds the default values and handling of zero value`

[1] https://github.com/devfile/devworkspace-operator/blob/68fbeb46519c59b6786e4716382913c6fea8272a/pkg/library/projects/clone.go#L97

tolusha (approved):

tolusha (approved):

pkg/deploy/dev-workspace-config/dev_workspace_config.go
amisevsk: Good idea -- applied

ibuziuk (approved):

PR #1729, opened 2023-07-11
add pod & container security context Che Cluster Fields

AObuchow:
### What does this PR do?

This PR introduces a couple of changes:

Adds 3 new Che Cluster CR fields: `spec.devEnvironments.security`, `spec.devEnvironments.security.containerSecurityContext` and `spec.devEnvironments.security.podSecurityContext`. The latter two fields allow configuring the pod and security contexts used by workspaces by setting the corresponding DWOC fields.

When the `devEnvironments.security.containerSecurityContext` field is used and `devEnvironments.disableContainerBuildCapabilities` is set to `false`, the container security context required for the `container-builds` SCC will be used, overriding the security context set in `devEnvironments.security.containerSecurityContext`.  

Some house-keeping:
- Moved all the DWOC reconciler tests into separate functions
- Cleaned up some DWOC reconciler test cases by removing expected fields that are unrelated to the test cases, and ignoring these fields in the `cmp.diff` options
- Cleaned up some DWOC reconciler test case names for accuracy
- Removed an unused error in the DWOC reconciler `updateWorkspacePodSchedulerNameConfig` function

n/a

Fix eclipse/che#22307 

### How to test this PR?
Install Che with the changes from this PR. On OpenShift, run `build/scripts/olm/test-catalog-from-sources.sh`

#### Normal usage
Disable container builds by setting `spec.devEnvironments.disableContainerBuildCapabilities: true` in the Che Cluster CR: `kubectl edit checluster eclipse-che -n eclipse-che`

Configure `spec.devEnvironments.containerSecurityContext` and `spec.devEnvironments.podSecurityContext` in the Che Cluster CR: `kubectl edit checluster eclipse-che -n eclipse-che`

For example:

```YAML
spec:  
(...)
 devEnvironments:  
   disableContainerBuildCapabilities: true  
   containerBuildConfiguration:  
     openShiftSecurityContextConstraint: container-build  
   containerSecurityContext:  
     allowPrivilegeEscalation: false  
     capabilities:  
       add:  
       - SYS_TIME  
       - CHOWN  
       drop:  
       - SETUID  
       - KILL  
       - SETGID  
     privileged: false  
     runAsUser: 1000690001
   podSecurityContext:  
     runAsGroup: 0  
     runAsNonRoot: false  
     runAsUser: 1000690001
```

Check that the containerSecurityContext and podSecurityContext are propagated to the Che-owned DWOC: `kubectl describe dwoc -n eclipse-che`

```YAML
Kind:                    DevWorkspaceOperatorConfig
API Version:  controller.devfile.io/v1alpha1  
Config:  
 Workspace:  
   Container Security Context:  
     Allow Privilege Escalation:  false  
     Capabilities:  
       Add:  
         SYS_TIME  
         CHOWN  
       Drop:  
         SETUID  
         KILL  
         SETGID  
     Privileged:         false  
     Run As User:        1000690001  
   Deployment Strategy:  Recreate  
   Pod Security Context:  
     Run As Group:     0  
     Run As Non Root:  false  
     Run As User:      1000690001  
   Progress Timeout:   300s  
   Service Account:  
     Disable Creation:  false
```

#### Container builds enabled case
Enable container builds by setting `spec.devEnvironments.disableContainerBuildCapabilities: false` in the Che Cluster CR: `kubectl edit checluster eclipse-che -n eclipse-che`

Any configuration to `spec.devEnvironments.containerSecurityContext` in the Che Cluster CR should be ignored, and instead, the security context required for container builds should be used. This can be verified by checking the DWOC's containerSecurityContext, it should have `allowPrivilegeEscalation: true` and the `SETGID`, `SETUID` add capabilities : 

```YAML
config:  
 workspace:  
   containerSecurityContext:  
     allowPrivilegeEscalation: true  
     capabilities:  
       add:  
       - SETGID  
       - SETUID  
   deploymentStrategy: Recreate  
   progressTimeout: 300s  
   serviceAccount:  
     disableCreation: false
```

### PR Checklist

- [ ] [Development resource are up to date](https://github.com/eclipse-che/che-operator/blob/main/README.md#update-development-resources)

### Reviewers


pkg/deploy/dev-workspace-config/dev_workspace_config.go
```diff
@@ -180,6 +193,78 @@ func updateProjectCloneConfig(devEnvironments *chev2.CheClusterDevEnvironments,
...
+	toDropCapabilitiesMap := getCapabilitiesSet(to.Drop)
+
+	// Get Add Capabilities that will be merged from
+	fromAddCapabilitiesMap := getCapabilitiesSet(from.Add)
+
+	// Get Drop Capabilities that will be merged from
+	fromDropCapabilitiesMap := getCapabilitiesSet(from.Drop)
+
+	// Remove add capabilities that are being dropped
+	for dropCapability := range fromDropCapabilitiesMap {
```
AObuchow, line 236: Note: for this PR, we don't _need_ to be removing add capabilities that are dropped in `from`, since `from` is `constants.DefaultWorkspaceContainerSecurityContext.Capabilities` which does not drop any capabilities.

However, I figured it'd make more sense to handle all cases, even if they aren't needed.

api/v2/checluster_types.go
```diff
@@ -121,8 +121,17 @@ type CheClusterDevEnvironments struct {
 	// +kubebuilder:default:=-1
 	SecondsOfRunBeforeIdling *int32 `json:"secondsOfRunBeforeIdling,omitempty"`
 	// Disables the container build capabilities.
+	// DEPRECATED: Usage of the containerSecurityContext and podSecurityContext fields is recommended instead to configure privileges required for container builds.
```
tolusha, line 124: I am against marking this field as DEPRECATED. 
It is still a valid and easy way to enable container build capabilities.

AObuchow: > It is still a valid and easy way to enable container build capabilities.

I agree @tolusha and am open to leaving it as-is.
@l0rd WDYT?

Perhaps its better to just modify the disableContainerBuildCapabilities documentation to mention that it overrides the containerSecurityContext (or have that noted on the containerSecurityContext documentation)?

l0rd:
We currently have:

```
devEnvironments:
  containerBuildConfiguration:
    openShiftSecurityContextConstraint: container-build
  disableContainerBuildCapabilities: true
```

I think we should have instead:

```
devEnvironments:
  openShiftSecurityContextConstraint: container-build
  containerSecurityContext:
    ...
  podSecurityContext:
   ...
```

or even 

```
devEnvironments:
  security:
    openShiftSecurityContextConstraint: container-build
    containerSecurityContext:
      ...
    podSecurityContext:
     ...
```

with the defaults securityContext being the bare minimum to enable building containers.

AObuchow:
> 
> or even
> 
> ```
> devEnvironments:
>   security:
>     openShiftSecurityContextConstraint: container-build
>     containerSecurityContext:
>       ...
>     podSecurityContext:
>      ...
> ```
> 
> with the defaults securityContext being the bare minimum to enable building containers.

I'm +1 for this, though I'm not sure how the process for removing Kubernetes CRD fields works.
@tolusha what are your thoughts? 

Won't existing Che Cluster CR instances that have `disableContainerBuildCapabilities` set become invalid if `disableContainerBuildCapabilities` is removed from the CRD? Maybe this issue was alleviated by https://github.com/eclipse/che/issues/22064, as `disableContainerBuildCapabilities` doesn't appear in the Che Cluster CR spec when instantiated, even though it's set to `true`.

Moreover, moving `devEnvironments.openShiftSecurityContextConstraint` to `devEnvironments.security.openShiftSecurityContextConstraint` would also break some (currently valid) Che Cluster CR instances I think.

@tolusha if you have any advice on any precautions I need to take in order to accomplish this, please let me know.

tolusha:
We can easily remove fields from a CRD. In this case the corresponding field will be removed from a CR a well.
Notice, that container build capabilities are disabled by default on K8S, otherwise workspaces fail to start. So, removing this field my affect k8s.

Unfortunately, we can't move fields. In order to do that, we need to introduce a new CheCluster API version. So, let's not introduce `security` section for now.

l0rd:
I am ok with @tolusha proposal. I just want to make sure we are on the same page:
- Today the admin has the choice to enable or disable "container-build capabilities". We need to replace that with a way to specify fine-grained and arbitrary pods/containers security context and users SCC.
- When the admin doesn't customize them in the CheCluster then the default security contexts and SCC should identical to today defaults.
- The admin should be also able to specify in the CheCluster an "empty" securityContext for workspace pods/containers and no SCC.
- Whatever customizaztion the admin does we should assume he knows what he is doing: we should not check if the SCC is compatible with the security contexts.
- For existing installations: if `disableContainerBuildCapabilities: true` we should set "empty" securityContext and no SCC. If `disableContainerBuildCapabilities: false` we should apply the default securityContext and use the SCC specified (if any) or use the `container-build` SCC. 
- The SCC is OpenShift-only but pods/container security context should be set on kubernetes too.

AObuchow:
I've updated the PR & the PR description:
- The `containerSecurityContext` and `podSecurityContext` fields are placed under `devEnvironments.security`, but `devEnvironments.containerBuildConfiguration` and `devEnvironments.disableContainerBuildCapabilities` remain unchanged. 
- The workspace security context can be configured by setting `devEnvironments.disableContainerBuildCapabilities: true`  and specifying the security context in `devEnvironments.security.containerSecurityContext`. This is a bit awkward, so I've made mention of it in the field's documentation.
- The default container security context used is the one required for container builds. It can be unset by setting  `devEnvironments.disableContainerBuildCapabilities: true` and an empty value for `devEnvironments.security.containerSecurityContext`.
- The default container security context is now defined with the environment variable `CHE_DEFAULT_SPEC_DEVENVIRONMENTS_CONTAINERSECURITYCONTEXT`


I did not mark `devEnvironments.disableContainerBuildCapabilities` as deprecated in its documentation, as it is still required to be set to false in order for `devEnvironments.security.containerSecurityContext` to take effect. @l0rd should we still mark it as deprecated?

api/v2/checluster_types.go
```diff
@@ -862,6 +879,18 @@ func (c *CheCluster) GetIdentityToken() string {
 	return constants.IdToken
 }
 
+// Returns the default container security context required for container builds.
+// Returns an error if the default container security context could not be retrieved.
+func (c *CheCluster) GetDefaultContainerSecurityContext() (*corev1.SecurityContext, error) {
```
tolusha, line 884: Could you move this func to `dev_workspace_config.go` ?

tolusha:
> I did not mark devEnvironments.disableContainerBuildCapabilities as deprecated in its documentation.

I believe we shouldn't mark this field as deprecated.

api/v2/checluster_types.go
AObuchow: Will do 👍

amisevsk (approved):
LGTM apart from @tolusha's comments

tolusha (approved):

AObuchow:
> > @l0rd should we still mark it as deprecated?
> 
> @AObuchow no I am ok with that. But I would mention in the doc that, when set to false (the default), containerSecurityContext field will be ignored and that a specific containerSecurityContext will be applied (with the details of the containerSecurityContext).

Sounds good, I've [updated](https://github.com/eclipse-che/che-operator/pull/1729/files#diff-10f5235ef9752eb68e9938878907ee064b366dd04540ac315d0b8912860a38dcR128-R137) the `disableContainerBuildCapabilities` field accordingly.

l0rd (approved):
👍 LGTM

tolusha:
@AObuchow 
Please merge.

AObuchow:
@tolusha I don't have merge right's, unfortunately. Could you please merge for me?

PR #1799, opened 2024-01-15
feat: sync Secrets, ConfigMaps and PersistentVolumesClaims to users n…

tolusha:
…amespaces

### What does this PR do?
sync Secrets, ConfigMaps and PersistentVolumesClaims to users namespases

N/A

https://github.com/eclipse/che/issues/20501

### How to test this PR?

1. Prepare a patch file if needed:
```bash
cat > /tmp/cr-patch.yaml <<EOF
apiVersion: org.eclipse.che/v2
kind: CheCluster
spec: {}
EOF
```

2. Deploy the operator:
 
#### OpenShift
```bash
./build/scripts/olm/test-catalog-from-sources.sh --cr-patch-yaml /tmp/cr-patch.yaml
```

#### on Minikube

```bash
./build/scripts/minikube-tests/test-operator-from-sources.sh --cr-patch-yaml /tmp/cr-patch.yaml
```
3. Create 500 namespaces
```
for i in {1..500}; do                                          
  oc apply -f - <<EOF
kind: Namespace
apiVersion: v1
metadata:
  name: test-$i
  labels:
    app.kubernetes.io/component: "workspaces-namespace"
    app.kubernetes.io/part-of: "che.eclipse.org"
    che.eclipse.org/che-namespace: "test-$i"
    che.eclipse.org/che-name: "che"
    che.eclipse.org/workspace-namespace-owner-uid: "owner-id"
  annotations:
    che.eclipse.org/username: "test-$i"
EOF
done
```

4. Create a ConfigMap in eclipse-che namespace
```
oc apply -f - <<EOF                                            
kind: ConfigMap   
apiVersion: v1
metadata:
  name: dev-env-variables
  namespace: eclipse-che
  labels:
    app.kubernetes.io/part-of: che.eclipse.org
    app.kubernetes.io/component: workspaces-config
EOF
```
5. Wait until reconcile is completed
```
oc logs deploy/che-operator -n eclipse-che -f
```

6. Check that CM exists in every namespace
```
for i in {1..500}; do                                  
  oc get cm -n test-$i dev-env-variables                 
done
```

### PR Checklist

- [x] [Development resource are up to date](https://github.com/eclipse-che/che-operator/blob/main/README.md#update-development-resources)

### Reviewers


controllers/che/checluster_controller.go
```diff
@@ -251,16 +250,11 @@ func (r *CheClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-			// Request object not found, could have been deleted after reconcile request.
-			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
-			// Return and don't requeue
-			return ctrl.Result{}, nil
-		}
+	checluster, err := deploy.FindCheClusterCRInNamespace(r.nonCachedClient, req.NamespacedName.Namespace)
+	if checluster == nil {
+		r.Log.Info("CheCluster Custom Resource not found.")
+		return ctrl.Result{}, nil
+	} else if err != nil {
```
amisevsk, line 257: We might want to add a secondary check here -- the previous functionality would get the specific CheCluster that triggered the reconcile (i.e. only return 0 or 1 CheClusters). Now it's possible for this section to return an error because there are 2+ CheClusters in the same namespace. If this occurs, the operator will enter a tight loop of 

> (queue reconcile for CheCluster A) -> (find two CheClusters, return error) -> (queue reconcile for CheCluster A).

If there are multiple CheClusters in a namespace, they should get some sort of error state and stop queueing reconciles until one of them is deleted.

controllers/usernamespace/usernamespace_controller.go
```diff
@@ -561,7 +521,7 @@ func (r *CheUserNamespaceReconciler) reconcileGitTlsCertificate(ctx context.Cont
 
 func (r *CheUserNamespaceReconciler) reconcileNodeSelectorAndTolerations(ctx context.Context, targetNs string, checluster *chev2.CheCluster, deployContext *chetypes.DeployContext) error {
 	ns := &corev1.Namespace{}
-	if err := r.client.Get(ctx, client.ObjectKey{Name: targetNs}, ns); err != nil {
+	if err := r.nonCachedClient.Get(ctx, client.ObjectKey{Name: targetNs}, ns); err != nil {
```
amisevsk, line 524: Is there a reason for switching to the non-cached client? For `Get` requests, this will lead to a lot more API traffic.

tolusha: I see, I will revert my changes.
The idea behind those changes was to slightly reduce memory consumption since we don't often update namespace to set node selector and toleration.

controllers/che/checluster_controller.go
tolusha: Thank you for your comment.
che-operator doesn't support more than one checluster and the webhook [1] prevents creating it.
So, it is not even possible to bump into the such an error, that's why I don't think it worth changing anything.

[1] https://github.com/eclipse-che/che-operator/blob/a4e19ead9d324dce09f5ca400a5cb7eff3abd253/api/v2/checluster_webhook.go#L86

controllers/usernamespace/usernamespace_controller.go
amisevsk: If it's very infrequent, then non-cached would be fine. The main difference is that with the cached client, any read requests (e.g. get, list, etc.) are free since it only ever checks the internal cache -- for the non-cached client, every time we get or list in a reconcile it has to make a k8s API request, which is slower and also hits the API server. If this reconcile runs infrequently, it should be fine to choose either. "Infrequent" here means reconciles, however, rather than updates to namespaces. 

We might also be able to limit memory usage by restricting which namespaces are cached or doing something similar.

controllers/che/checluster_controller.go
amisevsk: Makes sense :+1:

artaleks9:
Verified on Eclipse Che using `chectl server:deploy -p openshift --che-operator-image=quay.io/abazko/operator:20501`
The functionality works as expected.

artaleks9 (approved):

ibuziuk (approved):

PR #2169, opened 2026-07-27
chore: Add NetworkPolicy reconciler

tolusha:
### What does this PR do?

Introduces operator-managed NetworkPolicy resources for both the Che namespace and user workspace namespaces on OpenShift clusters.

A new `spec.networking.networkPolicy.enabled` field controls the feature (disabled by default). When enabled, the operator creates fine-grained ingress and egress NetworkPolicy resources to restrict network traffic to only the flows required by Che components and workspaces.

**Che namespace policies:**
- `allow-from-same-namespace` — allows ingress traffic between Che pods in the same namespace
- `allow-from-workspaces` — allows ingress traffic from user workspace namespaces
- `allow-from-openshift-ingress` — allows ingress traffic from the OpenShift ingress namespace
- `allow-from-openshift-monitoring` — allows ingress traffic from the OpenShift monitoring namespace
- `allow-from-<flavor>-operator` — allows ingress traffic from the operator pod to Che components
- `allow-all-egress` — allows all egress traffic from Che pods

**User workspace namespace policies:**
- `allow-from-<che-namespace>` — allows ingress traffic from Che namespace pods
- `allow-from-same-namespace` — allows ingress traffic between pods in the same namespace
- `allow-from-devworkspace-operator` — allows ingress traffic from the DevWorkspace operator
- `allow-from-openshift-monitoring` — allows ingress traffic from the OpenShift monitoring namespace
- `allow-from-openshift-ingress` — allows ingress traffic from the OpenShift ingress namespace
- `allow-all-egress` — allows all egress traffic from Che pods

Additional changes:
- Refactored `DwoNamespaceReconciler` into a utility function `GetDevWorkspaceOperatorNamespace`
- Extended cache selector to also watch DevWorkspace operator pods (for namespace lookup)
- Added `NetworkPolicy` diff option, `DeleteIgnoreNotFound` k8s client method, and related constants

N/A — infrastructure change, no UI impact.

[CRW-10821](https://redhat.atlassian.net/browse/CRW-10821)

### How to test this PR?

1. Deploy the operator:

#### OpenShift
```bash
./build/scripts/olm/test-catalog-from-sources.sh
```

or

```bash
build/scripts/docker-run.sh /bin/bash -c "
  oc login \
    --token=<...> \
    --server=<...> \
    --insecure-skip-tls-verify=true && \
  build/scripts/olm/test-catalog-from-sources.sh
"
```

2. Enable network policies in the CheCluster CR:
```bash
kubectl patch checluster eclipse-che -n eclipse-che --type merge -p '{"spec":{"networking":{"networkPolicy":{"enabled":true}}}}'
```

3. Verify NetworkPolicy resources are created in the Che namespace:
```bash
kubectl get networkpolicies -n eclipse-che
```
Expected: 6 NetworkPolicy resources (allow-from-same-namespace, allow-from-workspaces, allow-from-openshift-ingress, allow-from-openshift-monitoring, allow-from-che-operator, allow-all-egress).

4. Start a workspace and verify NetworkPolicy resources are created in the user namespace:
```bash
kubectl get networkpolicies -n <user-namespace>
```
Expected: 6 NetworkPolicy resources (allow-from-eclipse-che, allow-from-same-namespace, allow-from-devworkspace-operator, allow-from-openshift-monitoring, allow-from-openshift-ingress, allow-all-egress).

5. Disable network policies and verify cleanup:
```bash
kubectl patch checluster eclipse-che -n eclipse-che --type merge -p '{"spec":{"networking":{"networkPolicy":{"enabled":false}}}}'
kubectl get networkpolicies -n eclipse-che
kubectl get networkpolicies -n <user-namespace>
```
Expected: all Che-managed NetworkPolicy resources are deleted.

#### Common Test Scenarios
- [x] Deploy Eclipse Che
- [x] Start an empty workspace
- [x] Open terminal and build/run an image
- [x] Stop a workspace
- [x] GitHub Device authentication

### PR Checklist

### Reviewers


controllers/usernamespace/usernamespace_controller.go
```diff
@@ -590,6 +622,243 @@ func (r *CheUserNamespaceReconciler) reconcileSCCPrivileges(
...
+							},
+						},
+					},
+				},
+			},
+			PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress},
+		},
+	}
+
+	allowFromOpenShiftIngress := &networkingv1.NetworkPolicy{
```
rohanKanojia: nit, I see a lot of repeated `NetworkPolicy` construction. Maybe we can create a small helper (or a builder for common metadata) to reduce duplication?

tolusha: They are slightly different, all NP for eclipse-che namespace has a pod selector:
```
		PodSelector: metav1.LabelSelector{
				MatchLabels: map[string]string{
					constants.KubernetesPartOfLabelKey: constants.CheEclipseOrg,
				},
			},
```

tolusha (commented):
## Review of PR #2169 - Add NetworkPolicy reconciler

Overall the feature is well-structured: disabled by default, OpenShift-only guard, proper owner references for Che-namespace policies, and reasonable test coverage. A few things worth addressing before merge.

---

### User-namespace NetworkPolicies orphaned on CheCluster CR deletion

When the CheCluster CR is deleted, the user namespace controller's `Reconcile` returns early without cleaning up NetworkPolicies in user namespaces. Cross-namespace owner references are not supported, so GC won't handle these. The Che-namespace policies are fine - owner references cover them. But user-namespace NPs will persist after Che is uninstalled, potentially blocking workspace traffic with no obvious origin (the CR and operator are both gone at that point).

Does it make sense to add cleanup logic when `checluster == nil`, or at least document this as a known limitation requiring manual cleanup after uninstall?

---

### `allow-all-egress` permits unrestricted outbound traffic - worth documenting

Both reconcilers create an `allow-all-egress` policy with an empty egress rule, which permits all outbound traffic. This is the right default for workspaces (git, extensions, etc.), but administrators enabling this feature in a hardened environment may not expect unrestricted egress. Consider noting in the API doc comment that egress is intentionally unrestricted.

controllers/usernamespace/usernamespace_controller.go
```diff
@@ -590,6 +622,243 @@ func (r *CheUserNamespaceReconciler) reconcileSCCPrivileges(
 	)
 }
 
+func (r *CheUserNamespaceReconciler) reconcileNetworkPolicies(
+	ctx context.Context,
+	targetNs string,
+	deployContext *chetypes.DeployContext,
+) error {
+	policies, err := r.getNetworkPolicies(targetNs, deployContext)
```
tolusha: `getNetworkPolicies` is called unconditionally here, before the `IsNetworkPoliciesEnabled()` guard at line 635. `getNetworkPolicies` calls `GetDevWorkspaceOperatorNamespace`, which returns an error when no DWO pod is found. Since network policies are disabled by default, user-namespace reconciliation will fail with a spurious error any time the DWO pod is temporarily unavailable - for example during cluster startup or DWO upgrades.

Consider moving the `IsNetworkPoliciesEnabled` check to the top of `reconcileNetworkPolicies`, before calling `getNetworkPolicies`, so that DWO discovery is skipped entirely when the feature is off.

api/v2/checluster_types.go
```diff
@@ -329,6 +329,28 @@ type CheClusterSpecNetworking struct {
 	// +optional
 	// +kubebuilder:default:={gateway: {configLabels: {app: che, component: che-gateway-config}}}
 	Auth Auth `json:"auth"`
+	// NetworkPolicy configures NetworkPolicy resources for the Che namespace
+	// and user workspace namespaces. For OpenShift clusters only.
+	// When enabled, the following policies are created:
+	// In the Che namespace:
+	//   - allow-from-same-namespace: allows ingress traffic between Che pods in the same namespace.
+	//   - allow-from-workspaces-namespaces-to-openvsx-registry: allows ingress traffic from user workspace
```
tolusha: The comment lists `allow-from-workspaces-namespaces-to-openvsx-registry` and `allow-from-workspaces-namespaces-to-plugin-registry` as two separate policies, but `networkpolicies.go` creates a single policy named `allow-from-workspaces` that allows ingress from all workspace namespaces to all Che pods. Would it be possible to update this to match the actual policy name?

api/v2/checluster_types.go
```diff
@@ -337,6 +359,14 @@ type DevEnvironmentNetworking struct {
 	ExternalTLSConfig *ExternalTLSConfig `json:"externalTLSConfig,omitempty"`
 }
 
+// NetworkPolicy configuration settings.
+// +k8s:openapi-gen=true
+type NetworkPolicy struct {
+	// Enabled controls whether the operator creates NetworkPolicy resources.
+	// +optional
+	Enabled *bool `json:"enabled"`
```
tolusha: Similar bool pointer fields in the API use `json:"enabled,omitempty"` (e.g., `PersistentHomeConfig.Enabled`). Consider adding `omitempty` for consistency:

```suggestion
	Enabled *bool `json:"enabled,omitempty"`
```

pkg/deploy/networkpolicies/init_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"github.com/eclipse-che/che-operator/pkg/common/infrastructure"
+	defaults "github.com/eclipse-che/che-operator/pkg/common/operator-defaults"
+	"github.com/eclipse-che/che-operator/pkg/common/test"
+)
+
+func init() {
+	test.EnableTestMode()
+
```
tolusha, line 23: The reconciler is guarded by `infrastructure.IsOpenShift()` at registration time, but the test infrastructure initializes as `Kubernetes`. The tests work because they call the reconciler directly and bypass that guard, but it would be more consistent to use `OpenShiftV4` here - the `init_test.go` in the `usernamespace` package does this correctly.

```suggestion
	infrastructure.InitializeForTesting(infrastructure.OpenShiftV4)
```

controllers/usernamespace/usernamespace_controller.go
```diff
@@ -590,6 +622,243 @@ func (r *CheUserNamespaceReconciler) reconcileSCCPrivileges(
...
+	}
+
+	return nil
+}
+
+func (r *CheUserNamespaceReconciler) getNetworkPolicies(
+	targetNs string,
+	deployContext *chetypes.DeployContext,
+) ([]*networkingv1.NetworkPolicy, error) {
+	devWorkspaceOperatorNamespace, err := devworkspace.GetDevWorkspaceOperatorNamespace(deployContext.ClusterAPI.ClientWrapper)
```
tolusha: `GetDevWorkspaceOperatorNamespace` does a pod list on every user-namespace reconciliation. The main controller already resolves and stores `DwoNamespace` in `deployContext.DwoNamespace` (see `checluster_controller.go`), but the user namespace controller builds its own `DeployContext` without populating that field, then calls `GetDevWorkspaceOperatorNamespace` again here.

Does it make sense to populate `deployContext.DwoNamespace` once in the user namespace controller's `Reconcile` function, and pass it into `getNetworkPolicies` as a parameter rather than re-discovering it on each call?

pkg/deploy/networkpolicies/networkpolicies_test.go
```diff
@@ -0,0 +1,132 @@
...
+	reconciler := NewNetworkPoliciesReconciler()
+
+	_, done, err := reconciler.Reconcile(ctx)
+	assert.True(t, done)
+	assert.NoError(t, err)
+
+	npList := &networkingv1.NetworkPolicyList{}
+	err = ctx.ClusterAPI.Client.List(context.TODO(), npList, &client.ListOptions{Namespace: "eclipse-che"})
+	assert.NoError(t, err)
+	assert.Equal(t, len(allNetworkPolicyNames), len(npList.Items))
```
tolusha: The test asserts on policy count and `Get` success, but never inspects the `Spec` of any created policy. A bug that creates `allow-from-operator` with the wrong namespace selector, or `allow-from-workspaces` with an empty ingress rule, would pass silently.

Would it be possible to add at least one assertion on the spec of a representative policy - for example, verifying that `allow-from-operator` has the correct `NamespaceSelector` for the operator namespace?

controllers/usernamespace/usernamespace_controller.go
rohanKanojia: okay, we can keep it as it is then.

tolusha: I will add some utility func

controllers/usernamespace/usernamespace_controller_network_policies_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+	_, err := r.Reconcile(context.TODO(), reconcile.Request{NamespacedName: types.NamespacedName{Name: "user-project"}})
+	assert.NoError(t, err)
+
+	npList := &networkingv1.NetworkPolicyList{}
+	err = cl.List(context.TODO(), npList, &client.ListOptions{Namespace: "user-project"})
+	assert.NoError(t, err)
+	assert.Equal(t, len(allUserNsNetworkPolicyNames), len(npList.Items))
+
+	for _, name := range allUserNsNetworkPolicyNames {
```
rohanKanojia, line 55: The test currently verifies that the expected NetworkPolicy objects exist, but not their contents. Would it be worth adding an assertion in the spec? That might help catch regressions where a policy is created but with an incorrect selector or rule.

rohanKanojia:
I tested these changes and can confirm it works as expected. ✔️ 

After installing Eclipse Che based on these changes, and enabling Network Policies I could see NetworkPolicies getting created:
```
kubectl get networkpolicies -n eclipse-che
NAME                              POD-SELECTOR                                                   AGE
allow-all-egress                  app.kubernetes.io/part-of=che.eclipse.org                      5m11s
allow-from-openshift-ingress      app.kubernetes.io/part-of=che.eclipse.org                      5m11s
allow-from-openshift-monitoring   app.kubernetes.io/part-of=che.eclipse.org                      5m11s
allow-from-operator               app.kubernetes.io/part-of=che.eclipse.org                      5m11s
allow-from-same-namespace         app.kubernetes.io/part-of=che.eclipse.org                      5m11s
allow-from-workspaces             app.kubernetes.io/part-of=che.eclipse.org                      5m11s
eclipse-che-grpc-server           olm.catalogSource=eclipse-che,olm.managed=true                 63m
eclipse-che-unpack-bundles        olm.managed in (true),operatorframework.io/bundle-unpack-ref   63m
```
After creating a workspace, I could see NetworkPolicies in user namespace:
```
kubectl get networkpolicies -nkubeadmin-che
NAME                               POD-SELECTOR   AGE
allow-all-egress                   <none>         26s
allow-from-devworkspace-operator   <none>         26s
allow-from-eclipse-che             <none>         26s
allow-from-openshift-ingress       <none>         26s
allow-from-openshift-monitoring    <none>         26s
allow-from-same-namespace          <none>         26s
```

After disabling Network Policies I could see created NetworkPolicies getting cleaned up:
```
kubectl patch checluster eclipse-che -n eclipse-che --type merge -p '{"spec":{"networking":{"networkPolicy":{"enabled":false}}}}'
checluster.org.eclipse.che/eclipse-che patched

kubectl get networkpolicies -n eclipse-che
NAME                         POD-SELECTOR                                                   AGE
eclipse-che-grpc-server      olm.catalogSource=eclipse-che,olm.managed=true                 68m
eclipse-che-unpack-bundles   olm.managed in (true),operatorframework.io/bundle-unpack-ref   68m

che-operator : $ kubectl get networkpolicies -nkubeadmin-che
No resources found in kubeadmin-che namespace.
```

rohanKanojia (approved):

tolusha:

**Available commands**:

tolusha:
⚠️ **Warning: IDE/tool configuration files detected**

This PR contains changes to files in directories that are typically not intended to be committed:

- `.claude/rules/api-types.md`
- `.claude/rules/build-and-test.md`
- `.claude/rules/code-style.md`
- `.claude/rules/k8s-client.md`
- `.claude/rules/kubebuilder-defaults.md`
- `.claude/rules/pull-requests.md`
- `.claude/rules/reconcilers.md`
- `.claude/rules/redhat-compliance.md`
- `.claude/rules/resource-naming.md`
- `.claude/rules/testing.md`

Please verify these changes are intentional.

tolusha (commented):
Thanks for this PR - the network policy design is sound and the feature being disabled by default is the right call. Sharing some findings below (items already raised by @tolusha and @rohanKanojia are not repeated).

**Two findings worth discussing before merging:**
- `IsOperatorManagedComponent` now checks a third label, which is a behavioral change that could affect existing operator-managed resources during upgrades (see inline).
- User-namespace NetworkPolicies are orphaned when the CheCluster CR is deleted. Unlike orphaned ConfigMaps, orphaned NetworkPolicies actively restrict traffic, which could affect workloads after Che is uninstalled.

The remaining findings are suggestions and testing gaps.

---

**Design note - duplicated policy construction:** `pkg/deploy/networkpolicies/networkpolicies.go` and `controllers/usernamespace/usernamespace_controller.go` both define `allow-from-same-namespace`, `allow-from-openshift-monitoring`, `allow-from-openshift-ingress`, and `allow-all-egress`. Despite identical names, these have different semantics: the Che-namespace versions scope both the PodSelector and `From` peer to `part-of=che.eclipse.org`, while the user-namespace versions use an empty PodSelector (all pods). Would it help to extract structurally identical policies into a shared helper and add comments where they intentionally diverge, to avoid a developer fixing one without noticing the counterpart?

**Performance note:** Every CheCluster spec change triggers `triggerAllNamespaces`, enqueuing reconciliation for all N workspace namespaces. Each cycle calls `getNetworkPolicies` (DWO pod cache query) and up to 6 `Sync` calls. The `Sync` short-circuits when no change is needed, so actual API writes are minimal, but the Get-then-compare overhead scales linearly with user count. Worth noting in the feature documentation for large deployments.

pkg/deploy/labels.go
```diff
@@ -71,6 +71,8 @@ func GetLabelsAndAnnotations(obj client.Object) ([]string, []string) {
 	return slices.Collect(maps.Keys(obj.GetLabels())), slices.Collect(maps.Keys(obj.GetAnnotations()))
 }
 
-func IsPartOfEclipseCheResourceAndManagedByOperator(labels map[string]string) bool {
-	return labels[constants.KubernetesPartOfLabelKey] == constants.CheEclipseOrg && labels[constants.KubernetesManagedByLabelKey] == GetManagedByLabel()
+func IsOperatorManagedComponent(labels map[string]string, component string) bool {
```
tolusha, line 74: Consider the upgrade scenario here. `IsOperatorManagedComponent` now checks three labels (`part-of`, `managed-by`, AND `component`) compared to the old `IsPartOfEclipseCheResourceAndManagedByOperator` which checked only two. Both callers pass a specific component value. If any existing operator-managed resource on an upgraded cluster is missing the `component` label (possible if it was created before this label was consistently applied), `IsOperatorManagedComponent` returns `false` and the operator stops managing that resource. Does it make sense to verify that all resources created by earlier operator versions have always included the `component` label before relying on it as a third discriminator?

controllers/usernamespace/usernamespace_controller.go
```diff
@@ -135,6 +138,28 @@ func (r *CheUserNamespaceReconciler) commonRules(ctx context.Context, namesInChe
 	}
 }
 
+func (r *CheUserNamespaceReconciler) watchRulesForNetworkPolicies(_ context.Context) handler.EventHandler {
+	return handler.EnqueueRequestsFromMapFunc(
+		func(ctx context.Context, obj client.Object) []reconcile.Request {
+			workspaceInfo, _ := r.namespaceCache.GetNamespaceInfo(ctx, obj.GetNamespace())
```
tolusha: The error from `GetNamespaceInfo` is discarded here. If the namespace cache returns a transient error, `workspaceInfo` will be `nil` and the handler returns no reconcile requests - the NetworkPolicy event is silently dropped and not re-processed until the next CheCluster or namespace event. Would it be worth logging the error to make this easier to diagnose? (The same pattern exists in `commonRules`, so this is consistent, but it's new code.)

controllers/usernamespace/usernamespace_controller.go
```diff
@@ -166,7 +191,7 @@ func (r *CheUserNamespaceReconciler) triggerAllNamespaces() handler.EventHandler
 	return handler.EnqueueRequestsFromMapFunc(
 		handler.MapFunc(func(ctx context.Context, obj client.Object) []reconcile.Request {
 			nss := r.namespaceCache.GetAllKnownNamespaces()
-			ret := make([]reconcile.Request, len(nss))
+			ret := make([]reconcile.Request, 0, len(nss))
```
tolusha, line 211: Good fix - the old code allocated `len(nss)` zero-valued elements then appended `len(nss)` more, producing twice as many requests with half being zero-valued (empty namespace names). Consider noting this as a separate bug fix in the PR description.

controllers/usernamespace/usernamespace_controller.go
```diff
@@ -590,6 +622,243 @@ func (r *CheUserNamespaceReconciler) reconcileSCCPrivileges(
 	)
 }
 
+func (r *CheUserNamespaceReconciler) reconcileNetworkPolicies(
```
tolusha: When the CheCluster CR is deleted, `Reconcile` returns early at the `FindCheCluster` check before ever reaching `reconcileNetworkPolicies`. The Che-namespace policies are cleaned up via owner references (garbage collection), but user-namespace policies are not - cross-namespace owner references are unsupported. Unlike orphaned ConfigMaps or Secrets which are inert, orphaned NetworkPolicies actively restrict traffic and would need to be cleaned up manually after Che is uninstalled. Is this a known limitation? Would it be worth adding a finalizer or documenting the manual cleanup step?

pkg/deploy/devworkspace/dwo_namespace.go
```diff
@@ -16,62 +16,45 @@ import (
...
 			constants.KubernetesPartOfLabelKey: constants.DevWorkspaceOperatorName,
 		},
 	)
 
-	items, err := ctx.ClusterAPI.NonCachingClientWrapper.List(
+	items, err := clientWrapper.List(
 		context.TODO(),
 		&corev1.PodList{},
 		&client.ListOptions{LabelSelector: selector},
 	)
```
tolusha, line 38: The function now uses the cached client instead of the old `NonCachingClientWrapper`. During a DWO pod restart, there may be a brief window where the cached pod list is stale, causing this function to return an error and fail the entire CheCluster reconciliation until the cache syncs. The old non-cached path would have gotten current state. Would it make sense to add a doc comment noting that this function expects a cache-backed client and that the cache must include DWO pods via the `partOfEclipseCheOrDevWorkspace` selector configured in `cmd/main.go`?

pkg/deploy/devworkspace/dwo_namespace.go
```diff
@@ -16,62 +16,45 @@ import (
...
+
 	for _, item := range items {
 		pod := item.(*corev1.Pod)
 		if pod.Spec.ServiceAccountName == constants.DevWorkspaceServiceAccountName {
-			return pod.Namespace, nil
+			if devWorkspaceOperatorNamespace != "" && devWorkspaceOperatorNamespace != pod.Namespace {
+				return "", fmt.Errorf("multiple DevWorkspace Operator pods were found across different namespaces")
+			}
+			devWorkspaceOperatorNamespace = pod.Namespace
 		}
```
tolusha, line 56: The multi-namespace error path has no test coverage. The test setup (in `usernamespace_controller_setup_test.go`) creates a single DWO pod, but there is no test that creates pods in two different namespaces to verify this error is returned. Consider adding a test case for this branch.

pkg/deploy/networkpolicies/networkpolicies.go
```diff
@@ -0,0 +1,311 @@
...
+							},
+						},
+					},
+				},
+			},
+			PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress},
+		},
+	}
+
+	allowFromWorkspaces := &networkingv1.NetworkPolicy{
```
tolusha: The `allow-from-workspaces` policy uses only a `NamespaceSelector` with no `PodSelector`, so any pod in a workspace namespace - not just workspace pods - can reach Che infrastructure. In a strict multi-tenant environment, would it make sense to also add a pod selector restricting traffic to DevWorkspace-managed pods (e.g., pods with the DevWorkspace label)? This would close the aperture on untrusted pods that happen to be co-located in a workspace namespace.

pkg/common/diffs/diffs.go
```diff
@@ -81,6 +81,10 @@ func ConfigMap(labelKeys []string, annotationKeys []string) cmp.Options {
 	}
 }
 
+var NetworkPolicy = cmp.Options{
```
tolusha, line 84: Ignoring `ObjectMeta` means label or annotation drift on existing NetworkPolicies will never trigger an update via `Sync`. This is consistent with other diff options here (e.g., `ServiceMonitor`). Does it make sense to add a brief comment documenting this trade-off, so future contributors know that adding labels to existing policies would require a separate migration step rather than just updating the policy object?

pkg/deploy/networkpolicies/init_test.go
```diff
@@ -0,0 +1,26 @@
...
+	infrastructure.InitializeForTesting(infrastructure.Kubernetes)
+	defaults.InitializeForTesting("../../../config/manager/manager.yaml")
+}
```
tolusha, line 26: The `init_test.go` initializes as `Kubernetes`, which means the Che-namespace reconciler runs in tests even though in production it is never registered on Kubernetes (`if infrastructure.IsOpenShift()` in `checluster_controller.go`). Additionally, there is no test verifying that the user-namespace controller's `if infrastructure.IsOpenShift()` guard prevents policy creation on Kubernetes. A regression removing that guard would not be caught. Would it be worth adding a `TestNetworkPoliciesNotCreatedOnKubernetes` case?

controllers/usernamespace/usernamespace_controller_network_policies_test.go
```diff
@@ -0,0 +1,199 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var allUserNsNetworkPolicyNames = []string{
+	"allow-from-eclipse-che",
+	"allow-from-same-namespace",
+	"allow-from-devworkspace-operator",
```
tolusha, line 37: The `allUserNsNetworkPolicyNames` list hardcodes `"allow-from-eclipse-che"`. The policy name is dynamically computed as `"allow-from-" + deployContext.CheCluster.Namespace`, so a CheCluster in a namespace other than `eclipse-che` (e.g., `my-org`) would produce `"allow-from-my-org"`. Would it be worth adding a test with a non-default Che namespace to verify the dynamic name construction and confirm the hardcoded list stays in sync?

controllers/che/checluster_controller.go
```diff
@@ -240,31 +246,38 @@ func (r *CheClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	deployContext.Authentication, err = ResolveAuthentication(deployContext)
 	if err != nil {
 		r.Log.Error(err, "Error on resolving authentication")
 		return ctrl.Result{}, err
 	}
-	deployContext.Authentication = authentication
+
+	deployContext.DWONamespace, err = devworkspace.GetDevWorkspaceOperatorNamespace(clusterAPI.ClientWrapper)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("could not get DevWorkspaceOperator namespace: %w", err)
```
tolusha: The project convention (`.claude/rules/code-style.md`) requires error messages to start with `"failed to"`. Consider:
```go
return ctrl.Result{}, fmt.Errorf("failed to get DevWorkspace operator namespace: %w", err)
```

controllers/che/checluster_controller.go
```diff
@@ -240,31 +246,38 @@ func (r *CheClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	}
+	if deployContext.DWONamespace == "" {
+		r.Log.Error(nil, "DevWorkspaceOperator namespace not found.")
```
tolusha: Two related points here.

First, `r.Log.Error(nil, ...)` logs at error severity with a nil error. DWO being temporarily unavailable is an expected transient condition, not an error. Consider `r.Log.Info("DevWorkspace operator namespace not found, requeuing")` instead.

Second, moving DWO discovery to the pre-reconcile phase (before `reconcilerManager.ReconcileAll`) means the entire CheCluster pipeline - dashboard, gateway, server, devfile registry, etc. - is blocked whenever DWO is unavailable, even though those reconcilers have no dependency on DWO namespace. Previously `DwoNamespaceReconciler` was positioned later in the pipeline so earlier reconcilers could still run. During DWO restarts (30+ seconds on resource-constrained clusters) this delays all of Che. Would it be possible to only block the network policy reconciler on DWO availability rather than the full pipeline?

pkg/common/chetypes/types.go
```diff
@@ -27,7 +29,8 @@ type DeployContext struct {
 	Authentication          *Authentication
 	IsSelfSignedCertificate bool
 	CheHost                 string
-	DwoNamespace            string
+	DWONamespace            string
+	Context                 context.Context
```
tolusha, line 33: Go documentation advises against storing a `context.Context` inside a struct type. `DeployContext` is short-lived and request-scoped, which limits the risk, but this creates a hazard if the context is ever shared across goroutines or outlives a reconcile cycle. Would it be possible to pass `context.Context` as an explicit parameter to `SyncNetworkPolicy` and `DeleteNetworkPolicy` instead? That would match how `Sync` works elsewhere in the codebase.

pkg/deploy/networkpolicies/networkpolicies.go
```diff
@@ -0,0 +1,389 @@
...
+func (r *NetworkPoliciesReconciler) Finalize(_ *chetypes.DeployContext) bool {
+	return true
+}
+
+func GetNetworkPolicies(ctx *chetypes.DeployContext, namespace string) ([]*networkingv1.NetworkPolicy, error) {
+	isWorkspaceNetworkPolicies := ctx.CheCluster.Namespace != namespace
+
+	operatorNamespace, err := infrastructure.GetOperatorNamespace()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get operator namespace: %w", err)
```
tolusha, line 69: `GetOperatorNamespace()` is called unconditionally at the top of `GetNetworkPolicies`, but the operator namespace is only used to build `allowFromCheOperator` - which is only included in the Che-namespace policy set (`!isWorkspaceNetworkPolicies`). If `GetOperatorNamespace()` fails, workspace-namespace policy creation also fails unnecessarily. Consider moving the call inside the `if !isWorkspaceNetworkPolicies` branch to decouple the two code paths.

pkg/deploy/networkpolicies/networkpolicies.go
```diff
@@ -0,0 +1,389 @@
...
+		},
+		Spec: networkingv1.NetworkPolicySpec{
+			PodSelector: *podSelector.DeepCopy(),
+			Ingress: []networkingv1.NetworkPolicyIngressRule{
+				{
+					From: []networkingv1.NetworkPolicyPeer{
+						{
+							NamespaceSelector: &metav1.LabelSelector{
+								MatchLabels: map[string]string{
+									constants.KubernetesComponentLabelKey: constants.WorkspacesNamespaceComponentName,
```
tolusha, line 185: The `allow-from-workspaces` policy selects namespaces by `app.kubernetes.io/component: workspaces-namespace`. Unlike `kubernetes.io/metadata.name` (an immutable Kubernetes-managed label used elsewhere in this PR for `allow-from-<che-namespace>`), `app.kubernetes.io/component` is a mutable label that anyone with namespace-edit RBAC can set. On clusters where users can label namespaces (e.g., via self-provisioner), a user could label an arbitrary namespace to gain ingress access to all Che components.

Is this accepted as a known risk given OpenShift's default RBAC prevents regular users from labeling namespaces? If so, a comment documenting the assumption would help future readers.

controllers/usernamespace/usernamespace_controller_dwo_namespace.go
```diff
@@ -0,0 +1,133 @@
...
+			newNamespace := pod.GetNamespace()
+			currentNamespace := r.getDWONamespace()
+
+			if currentNamespace == newNamespace {
+				return
+			}
+
+			// DWO is present in a different namespace.
+			// No action is required for now; wait for the delete event.
+			if currentNamespace != "" {
```
tolusha: When a DWO pod is created in a different namespace than the currently cached one, the handler returns silently. `GetDevWorkspaceOperatorNamespace` explicitly treats multiple DWO installations across namespaces as an error, but that signal is swallowed here. Consider logging a warning so operators are alerted to the invalid multi-DWO state. If neither DWO is deleted, the warning from the comment ("wait for the delete event") is never raised.

controllers/usernamespace/usernamespace_controller_dwo_namespace.go
```diff
@@ -0,0 +1,133 @@
...
+			if currentDWONamespace != pod.GetNamespace() {
+				return
+			}
+
+			// Find the new DWO namespace
+			newDWONamespace, err := devworkspace.GetDevWorkspaceOperatorNamespace(r.clientWrapper)
+			if err != nil {
+				logger.Error(err, "Failed to get DevWorkspaceOperator namespace")
+				return
+			}
```
tolusha, line 87: `GetDevWorkspaceOperatorNamespace(r.clientWrapper)` is called from an event handler and uses `context.TODO()` internally for its `List` call, so there is no timeout or deadline. If the API server is slow or unresponsive, this blocks the event handler goroutine indefinitely and can starve the work queue. Consider passing the `ctx` from the `DeleteFunc` signature through to `GetDevWorkspaceOperatorNamespace`, or wrapping it with a short-timeout context before the call.

controllers/usernamespace/usernamespace_controller_dwo_namespace.go
```diff
@@ -0,0 +1,133 @@
...
+
+			// No DWO is currently present in the cluster.
+			// Wait for it to be redeployed, as it will likely reappear in the same namespace.
+			if newDWONamespace == "" {
+				return
+			}
+
+			if currentDWONamespace == newDWONamespace {
```
tolusha, line 89: When `newDWONamespace == ""` (DWO fully removed), the handler returns early without clearing the cached namespace. This creates a permanent staleness risk: if DWO is later reinstalled in a different namespace (`ns-B`), the `CreateFunc` sees `currentNamespace == "ns-A"` (non-empty) and returns early without updating. The cache stays permanently on `ns-A`, and all `allow-from-devworkspace-operator` policies in workspace namespaces will reference a defunct namespace. The only remediation is restarting the operator.

Consider calling `r.setDWONamespace("")` here before returning, so the next `CreateFunc` event for the new namespace can update the cache correctly. Or leave the stale value intentionally but add a comment explaining that the assumption is DWO will reappear in the same namespace (which the current comment already hints at).

controllers/usernamespace/usernamespace_controller.go
```diff
@@ -216,8 +258,11 @@ func (r *CheUserNamespaceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		ClusterAPI: chetypes.ClusterAPI{
 			Client:           r.client,
 			NonCachingClient: r.nonCachedClient,
+			ClientWrapper:    r.clientWrapper,
 			Scheme:           r.scheme,
 		},
+		Context:      ctx,
```
tolusha, line 264: `NonCachingClientWrapper` is not set in this `DeployContext`. The field `r.nonCachedClientWrapper` exists on the struct, but it is omitted here. Currently `networkpolicies.SyncNetworkPolicy` and `DeleteNetworkPolicy` only use `ClientWrapper`, so there is no runtime failure today. However, the project's `k8s-client.md` documents `NonCachingClientWrapper` as the standard path for non-cached operations, making it likely to be accessed by future callers. Would it make sense to add `NonCachingClientWrapper: r.nonCachedClientWrapper` to maintain the contract?

pkg/deploy/labels.go
```diff
@@ -71,6 +71,13 @@ func GetLabelsAndAnnotations(obj client.Object) ([]string, []string) {
...
+	return labels[constants.KubernetesPartOfLabelKey] == constants.CheEclipseOrg &&
```
tolusha, line 75: `IsOperatorManagedComponent` adds a `component` label check that the old `IsPartOfEclipseCheResourceAndManagedByOperator` did not require. Resources created by a prior operator version with `part-of=che.eclipse.org` and `managed-by=<operator>` but without a `component` label (e.g., SCCs created by an older release) will no longer be recognized as operator-managed on upgrade. They would be treated as custom resources and skipped, potentially leaving stale SCC configurations or misidentified OpenVSX secrets. Is this an accepted upgrade path, or does it warrant a migration/detection step?

tolusha (commented):
Thanks for the thorough PR - the separation of concerns between the CheCluster and user namespace controllers is clean, the test coverage in `networkpolicies_test.go` is extensive, and the DWO namespace caching approach with pod watches is a solid improvement over per-reconcile cluster-wide lists. A few findings below, ranging from correctness concerns to suggestions.

**Warnings (correctness / operational risk):**
- DWO namespace cache can become permanently stale after a DWO namespace migration (see inline comment on `usernamespace_controller_dwo_namespace.go:106`)
- The entire CheCluster reconciliation pipeline is now blocked when DWO is unavailable (see inline on `checluster_controller.go:271`)
- The event handler `DeleteFunc` uses `context.TODO()` with no timeout (see inline on `usernamespace_controller_dwo_namespace.go:98`)

**Verification needed:**
- No RBAC changes appear in the diff for NetworkPolicy create/delete in user namespaces - see general comment G1 below.

**Suggestions:**
- Error message format, log level, unconditional `GetOperatorNamespace()` call, mutable label security boundary, missing `NonCachingClientWrapper`, upgrade compatibility of the new label check, and test coverage gaps - all detailed in inline and general comments below.

pkg/deploy/networkpolicies/networkpolicies.go
```diff
@@ -0,0 +1,389 @@
...
+	err := SyncNetworkPolicy(ctx, ctx.CheCluster.Namespace)
+	if err != nil {
+		return reconcile.Result{}, false, fmt.Errorf("failed to sync NetworkPolicy in namespace %s: %w", ctx.CheCluster.Namespace, err)
+	}
+
+	return reconcile.Result{}, true, nil
+}
+
+func (r *NetworkPoliciesReconciler) Finalize(_ *chetypes.DeployContext) bool {
+	return true
```
tolusha, line 61: `Finalize` returns `true` without cleaning up. When CheCluster is deleted, Che-namespace policies are garbage-collected via owner references, but user-namespace policies persist - cross-namespace owner references are not supported. Unlike orphaned ConfigMaps, these restrictive NetworkPolicies actively block traffic to workspace pods after Che is removed, requiring manual cleanup.

`DeleteNetworkPolicy` already implements label-based cleanup. Would it make sense to wire it into `Finalize` by iterating `namespaceCache.GetAllKnownNamespaces()`? That would require either passing the namespace cache to `NetworkPoliciesReconciler` or coordinating cleanup through a finalizer on the user-namespace controller side.

pkg/deploy/networkpolicies/networkpolicies.go
```diff
@@ -0,0 +1,389 @@
...
+			networkPolicies = append(networkPolicies, allowFromDevWorkspaceOperator)
+		}
+	} else {
+		networkPolicies = append(networkPolicies, allowFromWorkspaces, allowFromCheOperator)
+	}
+
+	return networkPolicies, nil
+}
+
+func SyncNetworkPolicy(ctx *chetypes.DeployContext, namespace string) error {
```
tolusha, line 338: `SyncNetworkPolicy` applies policies sequentially. If the third `Sync` call fails, the first two are already applied and the namespace is left with a partial set of network restrictions until the next reconcile.

Similarly in `DeleteNetworkPolicy`, a mid-loop failure leaves some policies deleted and others intact, continuing to restrict traffic after the feature is disabled.

Does it make sense to at least add a comment noting this transient inconsistency, so operators debugging connectivity issues know to check for a partial policy set?

pkg/deploy/networkpolicies/networkpolicies.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	return reconcile.Result{}, true, nil
+}
+
+func (r *NetworkPoliciesReconciler) Finalize(_ *chetypes.DeployContext) bool {
+	return true
+}
+
+func GetNetworkPolicies(ctx *chetypes.DeployContext, namespace string) ([]*networkingv1.NetworkPolicy, error) {
```
tolusha, line 64: `GetNetworkPolicies` constructs every policy object unconditionally (both `allowFromWorkspaces`/`allowFromCheOperator` and `allowFromChe`/`allowFromDevWorkspaceOperator`), then selects a subset at the end via `isWorkspaceNetworkPolicies`. This means `infrastructure.GetOperatorNamespace()` is called and the full workspace-only policy structs are allocated even when generating Che-namespace policies where they are discarded, and vice versa.

Consider splitting into `GetCheNamespacePolicies` and `GetWorkspaceNamespacePolicies`, or adding early branching, to make the code self-documenting and avoid unnecessary work.

pkg/deploy/networkpolicies/networkpolicies.go
```diff
@@ -0,0 +1,389 @@
...
+	isWorkspaceNetworkPolicies := ctx.CheCluster.Namespace != namespace
+
+	operatorNamespace, err := infrastructure.GetOperatorNamespace()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get operator namespace: %w", err)
+	}
+
+	var podSelector metav1.LabelSelector
+
+	if isWorkspaceNetworkPolicies {
```
tolusha, line 74: In the workspace namespace path (`isWorkspaceNetworkPolicies == true`), `podSelector` is `metav1.LabelSelector{}` (empty), so `allow-from-same-namespace` allows every pod in the namespace to communicate with every other pod. Any injected sidecar, admission webhook proxy, or service mesh pod gets full ingress access to all workspace pods.

The Che-namespace version is narrower (restricted to `part-of=che.eclipse.org`). Is the broader selector here intentional? If so, would it be worth documenting this design choice in the CRD field description or a code comment?

pkg/deploy/labels.go
```diff
@@ -71,6 +71,13 @@ func GetLabelsAndAnnotations(obj client.Object) ([]string, []string) {
 	return slices.Collect(maps.Keys(obj.GetLabels())), slices.Collect(maps.Keys(obj.GetAnnotations()))
 }
 
-func IsPartOfEclipseCheResourceAndManagedByOperator(labels map[string]string) bool {
-	return labels[constants.KubernetesPartOfLabelKey] == constants.CheEclipseOrg && labels[constants.KubernetesManagedByLabelKey] == GetManagedByLabel()
+func IsOperatorManagedComponent(labels map[string]string, component string) bool {
```
tolusha, line 74: The old `IsPartOfEclipseCheResourceAndManagedByOperator` checked two labels (`part-of` and `managed-by`). `IsOperatorManagedComponent` checks three (`part-of`, `managed-by`, AND `component`). Both callers pass a specific component value, so current behavior is correct.

The concern is upgrade safety: if any existing operator-managed resource on an upgraded cluster lacks the `component` label, this function returns `false`. For the SCC caller in `container_capabilities.go` this could cause the operator to skip applying SCC settings to existing resources. Would it make sense to verify that all pre-existing managed resources carry the `component` label before the upgrade, or to handle the missing-label case explicitly?

controllers/che/checluster_controller.go
```diff
@@ -240,31 +246,38 @@ func (r *CheClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 		r.Log.Error(err, "Error on resolving authentication")
 		return ctrl.Result{}, err
 	}
-	deployContext.Authentication = authentication
+
+	deployContext.DWONamespace, err = devworkspace.GetDevWorkspaceOperatorNamespace(ctx, clusterAPI.ClientWrapper)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get DevWorkspaceOperator namespace: %w", err)
+	}
+	if deployContext.DWONamespace == "" {
```
tolusha, line 270: When `DWONamespace == ""`, the controller returns and requeues after 10 seconds. Because this runs before `reconcilerManager.ReconcileAll`, the entire pipeline (dashboard, gateway, server, devfile registry, etc.) is blocked every time DWO is unavailable.

The old `DwoNamespaceReconciler` ran mid-pipeline, so earlier reconcilers could complete independently. Would it make sense to keep the early-return only when networkPolicy is enabled and DWO namespace is actually required, letting the rest of the pipeline proceed when DWO is absent?

controllers/usernamespace/usernamespace_controller_dwo_namespace.go
```diff
@@ -0,0 +1,116 @@
...
+				q.Add(reconcile.Request{
+					NamespacedName: types.NamespacedName{Name: namespace},
+				})
+			}
+		},
+		// Handle the delete event to resolve an invalid configuration where
+		// multiple DevWorkspace Operators are running in different namespaces.
+		DeleteFunc: func(
+			ctx context.Context,
+			evt event.DeleteEvent,
```
tolusha, line 67: The `DeleteFunc` handler calls `devworkspace.GetDevWorkspaceOperatorNamespace(ctx, r.clientWrapper)` where `r.clientWrapper` wraps the **cached** client. If the informer cache still contains the just-deleted pod when this handler fires, `GetDevWorkspaceOperatorNamespace` returns the old namespace. The `currentDWONamespace == newDWONamespace` check then early-returns without clearing the cache.

Would it be possible to use `r.nonCachedClientWrapper` here instead, to ensure the handler sees the post-deletion cluster state?

controllers/usernamespace/usernamespace_controller_dwo_namespace.go
```diff
@@ -0,0 +1,116 @@
...
+			currentDWONamespace := r.getDWONamespace()
+
+			// Find the new DWO namespace
+			newDWONamespace, err := devworkspace.GetDevWorkspaceOperatorNamespace(ctx, r.clientWrapper)
+			if err != nil {
+				logger.Error(err, "Failed to get DevWorkspaceOperator namespace")
+				return
+			}
+
+			if currentDWONamespace == newDWONamespace {
```
tolusha, line 89: When `GetDevWorkspaceOperatorNamespace` returns an error, the handler logs and returns without clearing `r.dwoNamespace`. The stale cached namespace persists. Network policies in user namespaces continue to reference the old (possibly deleted) DWO namespace and the reconciler is not triggered to correct the state.

Would it make sense to clear the cached namespace and enqueue all known namespaces on error, so the next reconcile re-discovers the correct DWO namespace? Event handlers cannot requeue themselves, but explicitly enqueueing all namespaces achieves the same effect.

cmd/main.go
```diff
@@ -335,73 +336,79 @@ func main() {
 
 // watch k8s objects with labels `app.kubernetes.io/part-of=che.eclipse.org`
 func getCacheFunc() (cache.NewCacheFunc, error) {
-	partOfCheObjectSelector, err := labels.Parse(fmt.Sprintf("%s=%s", constants.KubernetesPartOfLabelKey, constants.CheEclipseOrg))
+	partOfEclipseChe := labels.SelectorFromSet(map[string]string{constants.KubernetesPartOfLabelKey: constants.CheEclipseOrg})
+
+	partOfCheOrDWO, err := labels.NewRequirement(
```
tolusha, line 341: The Pod cache selector now uses `part-of IN (che.eclipse.org, devworkspace-operator)`. Since the cache has no namespace restriction, the informer watches and caches ALL pods across ALL namespaces carrying the DWO label. On large multi-tenant clusters or environments with many DWO instances, this increases operator memory and event processing proportionally.

The DWO controller set is typically small (1-3 pods) so the practical impact should be modest. Would it be worth adding a comment noting this trade-off, or restricting the DWO pod watch to a known namespace once it has been discovered?

pkg/deploy/networkpolicies/networkpolicies_test.go
```diff
@@ -0,0 +1,513 @@
...
+var cheClusterPolicyNames = []string{
+	"allow-from-same-namespace",
+	"allow-from-openshift-ingress",
+	"allow-from-openshift-monitoring",
+	"allow-from-workspaces",
+	"allow-from-che-operator",
+	"allow-all-egress",
+}
+
+var workspacePolicyNames = []string{
```
tolusha, line 42: `workspacePolicyNames` includes `allow-from-devworkspace-operator`, which is only created when `ctx.DWONamespace != ""`. There is no test verifying the expected policy count when `DWONamespace == ""` (5 instead of 6).

There is also a subtler issue: if a stale `allow-from-devworkspace-operator` policy exists from a previous reconcile (when DWO was available) and DWO is then removed, `SyncNetworkPolicy` will not create a new one (since it is absent from the returned list), but neither will it delete the stale one - `DeleteNetworkPolicy` is only called when the feature is fully disabled. The stale policy references a now-nonexistent namespace and persists indefinitely.

Would it be possible to add tests for:
1. The correct policy count when `DWONamespace == ""`
2. That transitioning from DWO present to absent correctly removes the `allow-from-devworkspace-operator` policy

tolusha (commented):
The feature design is sound - disabled by default, OpenShift-only guard, owner references for Che-namespace policies, and 700+ lines of test coverage. The following observations complement the existing reviewer comments.

**Three general findings:**

**G1: PR description policy count mismatch.** The "How to test" section says "Expected: 7 NetworkPolicy resources" for the Che namespace, but the code creates 6: allow-from-same-namespace, allow-from-workspaces, allow-from-openshift-ingress, allow-from-openshift-monitoring, allow-from-{flavor}-operator, allow-all-egress. The test expectations correctly assert 6. Consider updating the description to match.

**G2: `config/rbac/cluster_role.yaml` missing `list`, `watch`, `patch` for networkpolicies.** The kustomize source file grants only `create`, `delete`, `get`, `update`. The generated deploy manifests and CSV include the full verb set, so OLM-based installs work. Any workflow applying `config/rbac/cluster_role.yaml` directly via kustomize would fail when `DeleteNetworkPolicy` calls `List` or the informer tries to `Watch`. Would it be possible to add the missing verbs to the source file as well?

**G3: No structured logging for NetworkPolicy lifecycle events.** Neither `SyncNetworkPolicy` nor `DeleteNetworkPolicy` emit Info-level logs on successful create, update, or delete. For a security-sensitive feature, a log entry per policy operation (namespace, policy name, outcome) would make it much easier to audit operator behavior and debug connectivity issues.

dkwon17:
Core functionality seems to be working as expected. I've successfully verified steps 2 through 5 above.

I also created a regular user and workspace starts / workspace samples / plugin installation was working as expected:
image
image

PR #2188, opened 2026-08-31
Retrieve TLS config from apiserver for OpenShift 5.0

dkwon17:
### What does this PR do?
Configures operator webhook and metric server to apply the cluster TLS profile when `tlsAdherence` is set to `StrictAllComponents`. Watcher restarts the operator on profile or adherence policy changes.

### How to test this PR?

I have personally tested this functionality on a OCP 5.0 cluster with techpreview enabled, with this operator image: `quay.io/dkwon17/che-operator:tls-adherence`.

1. Deploy the operator:
 
#### OpenShift
```bash

OPERATOR_IMAGE=quay.io/dkwon17/che-operator:tls-adherence
```

For testing, I have followed this markdown: https://gist.github.com/dkwon17/71e6ede0f1f94f9fba8f57a21aa1e41d

#### on Minikube

```bash
```

#### Common Test Scenarios
- [ ] Deploy Eclipse Che
- [ ] Stop a workspace

### PR Checklist

### Reviewers


cmd/main.go
```diff
@@ -240,9 +244,15 @@ func main() {
 	}
 
 	mgr, err := ctrl.NewManager(config, ctrl.Options{
-		Scheme:                        scheme,
-		Metrics:                       server.Options{BindAddress: metricsAddr},
-		WebhookServer:                 webhook.NewServer(webhook.Options{Port: 9443}),
+		Scheme: scheme,
+		Metrics: server.Options{
+			BindAddress: metricsAddr,
+			TLSOpts:     serverTLS.TLSOpts,
```
rohanKanojia, line 250: `Metrics.TLSOpts` is set here, but I don't see `SecureServing` being set alongside it.

Is enabling `SecureServing` in scope for this PR, or is the intent to cover metrics separately later?

dkwon17: I think it is out of scope for this PR

pkg/tlssetup/server_tls.go
```diff
@@ -0,0 +1,124 @@
...
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package tlssetup
```
tolusha, line 13: can we move everything related to tls package?

cmd/main.go
tolusha, line 250: How does it behave if TLSOpts is empty?

tolusha:
@dkwon17 
Could you pls rebase on main to resolve conflicts

cmd/main.go
dkwon17: In this case, it will be a no-op ([webhook](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.24.1/pkg/webhook/server.go#L196-L199), [metrics](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.24.1/pkg/metrics/server/server.go#L281-L284)), and will behave how the operator currently behaves on main

pkg/tlssetup/server_tls.go
dkwon17: Yes, my latest commit moves this file into the tls package

tolusha:
Pls update clear-defined-test.sh with following changes:
ignored_paths:
```
  ["k8s.io/kube-aggregator"]="Harvesting is in progress"
  ["github.com/openshift/library-go"]="Harvesting is in progress"
  ["github.com/openshift/controller-runtime-common"]="Harvesting is in progress"
  ["github.com/openshift/client-go"]="Harvesting is in progress"
  ["github.com/coreos/go-oidc"]="Harvesting is in progress"
```

ignored_paths_licenses:
```
  # https://github.com/kubernetes/kube-aggregator?tab=Apache-2.0-1-ov-file
  ["k8s.io/kube-aggregator"]="Apache-2.0"
  # https://github.com/openshift/library-go?tab=Apache-2.0-1-ov-file
  ["github.com/openshift/library-go"]="Apache-2.0"
  # https://github.com/openshift/controller-runtime-common?tab=Apache-2.0-1-ov-file
  ["github.com/openshift/controller-runtime-common"]="Apache-2.0"
  # https://github.com/openshift/client-go?tab=Apache-2.0-1-ov-file
  ["github.com/openshift/client-go"]="Apache-2.0"
  # https://github.com/coreos/go-oidc?tab=Apache-2.0-1-ov-file
  ["github.com/coreos/go-oidc"]="Apache-2.0"
```

cmd/main.go
```diff
@@ -312,6 +322,14 @@ func main() {
 	}
 	sigHandler := signal.SetupSignalHandler(terminationPeriod)
 
+	ctx, cancelCtx := context.WithCancel(sigHandler)
```
tolusha, line 325: Probably not needed, because  of 
https://github.com/dkwon17/che-operator/blob/a4e19ead9d324dce09f5ca400a5cb7eff3abd253/pkg/common/signal/signal_handler.go#L37

pkg/deploy/tls/server_tls.go
```diff
@@ -0,0 +1,125 @@
...
+	serverTLS := ServerTLS{
+		InitialTLSProfileSpec:     profile,
+		InitialTLSAdherencePolicy: adherence,
+		profileFetched:            true,
+	}
+
+	if shouldHonorClusterTLSProfile(adherence) {
+		tlsConfigFn, unsupported := tlspkg.NewTLSConfigFromProfile(profile)
+		if len(unsupported) > 0 {
+			log.Info("TLS profile contains ciphers unsupported by Go", "unsupported", unsupported)
```
tolusha, line 73: should simply return if there is `unsupported` ?

dkwon17: I think we can continue, since unsupported ciphers are filtered out in `cipherCodes` function in: https://github.com/eclipse-che/che-operator/blob/3b1addd70b5d150a15c21f4cf7b5011fb690c044/vendor/github.com/openshift/controller-runtime-common/pkg/tls/tls.go#L128-L129

WDYT?

cmd/main.go
dkwon17: I think this is needed to get the cancel function for https://github.com/dkwon17/che-operator/blob/3b1addd70b5d150a15c21f4cf7b5011fb690c044/cmd/main.go#L328 which is used to restart the operator when there are TLS profile changes in the apiserver resource

tolusha:

Task completed.

tolusha (commented):
Reviewed the full non-vendor diff, and read the vendored `controller-runtime-common/pkg/tls` and `library-go/pkg/crypto` sources to verify the integration contracts rather than infer them.

The webhook path is well built and the reuse of `controller-runtime-common` is the right call. The main blockers are the metrics claim (see the reply on @rohanKanojia's thread) and, for a security feature, the silent fail-open with no recovery (see `pkg/deploy/tls/server_tls.go:41`).

**Things this gets right**, worth recording because several are traps that are easy to fall into:

- `configv1.APIServer` is deliberately absent from the `ByObject` selectors in `getCacheFunc`, so the `part-of=che.eclipse.org` label filter does not blind the informer.
- `BuildServerTLSOptions` and the upstream `Reconcile` both derive the spec through `GetTLSProfileSpec`, so `reflect.DeepEqual` at `controller.go:137` cannot fire a spurious restart on the first reconcile.
- The RBAC split (`get` + `resourceNames` for the direct client, unscoped `list`/`watch` for the informer) is the tightest grant Kubernetes allows, read-only, on a dedicated ServiceAccount. All six generated manifests are in sync, and adding OpenShift rules to the Kubernetes ClusterRole matches pre-existing generator output.
- `NeedLeaderElection: false` upstream means every replica reloads - correct for a process-wide concern rather than a reconcile concern.
- `IsOpenShift()`'s lazy `initializeIfNeeded()` makes the early call site at `main.go:226` safe despite running before the discovery client is built.
- Reusing `controller-runtime-common` avoids the TLS 1.3 "cipher suites are not configurable in Go" subtlety, which is easy to get wrong by hand.
- Commits are cleanly separated by intent.

**Smaller notes** that did not seem worth their own threads:

- `cmd/main.go:73` - the `tls` alias is redundant; the package is already named `tls` and `crypto/tls` is not imported in `main.go`.
- `pkg/deploy/tls/server_tls.go:81` - logging the full cipher list at `Info` on every startup is verbose; `V(1)` might suit it better. `MinTLSVersion` at `Info` is useful.
- `RegisterSecurityProfileWatcher` returns `nil` when it skips registration, so the caller cannot distinguish "watcher installed" from "watcher skipped because the fetch failed". One `log.Info` on that branch would close the gap.
- There is no metric, event, or status condition reporting whether the cluster profile is being honored. Combined with the fail-open path, a degraded operator is indistinguishable from a healthy one.
- Worth a release note: honoring an `Old` cluster profile sets `MinVersion: VersionTLS10` on the webhook, which is *lower* than what we serve today. That is the correct semantics of "honor the cluster profile", but the PR title reads as hardening only.

**Verified as already handled**, so no action needed: the `clear-defined-test.sh` entries are correct and not stale (`kube-aggregator`, `client-go` and `go-oidc` are absent from `vendor/` but do appear in `go list -m -mod=mod all`, which line 20 of the script uses); the `context.WithCancel` wrapper at `main.go:325` is necessary as @dkwon17 explained, since `SetupSignalHandler` keeps its `cancel` private; and the empty-`TLSOpts` no-op behaviour checks out against both vendored loops.

The PR also currently carries `needs-rebase` and a failing `ci/prow/v19-devworkspace-happy-path`.

pkg/deploy/tls/server_tls.go
```diff
@@ -0,0 +1,125 @@
...
+	TLSOpts                   []func(*cryptotls.Config)
+	InitialTLSProfileSpec     configv1.TLSProfileSpec
+	InitialTLSAdherencePolicy configv1.TLSAdherencePolicy
+	profileFetched            bool
+}
+
+// BuildServerTLSOptions fetches TLS profile and adherence policy from cluster.
+// Returns TLS config functions when adherence policy requires strict compliance.
+// Falls back to empty TLSOpts on non-OpenShift, RBAC failures, or legacy adherence policy.
+func BuildServerTLSOptions(ctx context.Context, cfg *rest.Config, scheme *k8sruntime.Scheme, log logr.Logger) ServerTLS {
```
tolusha: `BuildServerTLSOptions` has no `error` return, so four distinct failure modes collapse into one indistinguishable zero value. The caller at `cmd/main.go:226` cannot tell "this is plain Kubernetes, nothing to do" from "this is an OpenShift cluster with `StrictAllComponents` and we could not read the profile" - two situations that call for opposite responses.

The result is that on a cluster which explicitly asked for strict adherence, an RBAC gap or a transient API error leaves the webhook on Go defaults, and because `profileFetched` stays `false` the watcher is never registered either - so the pod never recovers for its entire lifetime. Nothing in the cluster reports this: no metric, no event, no status condition.

There is a concrete trigger. On the `oc apply -f combined.yaml` upgrade path in the PR description, the new image and the new ClusterRole land in the same apply with no ordering guarantee. If the pod wins the race, the fetch 403s and the operator runs non-compliant permanently. The OLM path is safe because CSV permissions are applied before the deployment.

Would it make sense to return `(ServerTLS, error)` and handle it the way the neighbouring cluster calls in `main.go` already do?

```go
serverTLS, err := tls.BuildServerTLSOptions(ctx, config, scheme, setupLog)
if err != nil {
	setupLog.Error(err, "failed to build server TLS options")
	os.Exit(1)
}
```

Failing startup turns a silent permanent degradation into a loud, self-correcting one.

It would also unblock testing. The function builds its own client from `*rest.Config` + `*runtime.Scheme`, so there is no seam for `fake.NewClientBuilder()`. Accepting a `client.Client` instead would make the whole body testable with the patterns already in `certificates_test.go`.

pkg/deploy/tls/server_tls.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		log.Error(err, "failed to fetch TLS profile, using Go defaults")
+		return ServerTLS{}
+	}
+
+	adherence, err := tlspkg.FetchAPIServerTLSAdherencePolicy(ctx, cl)
+	if err != nil {
+		log.Error(err, "failed to fetch TLS adherence policy, using Go defaults")
+		return ServerTLS{}
+	}
```
tolusha: `FetchAPIServerTLSProfile` and `FetchAPIServerTLSAdherencePolicy` each issue their own `Get` against `APIServer/cluster`. If the object is edited between them - plausible during a compliance rollout, which flips `tlsAdherence` and `tlsSecurityProfile` together - we record a profile from revision N and a policy from revision N+1.

Two consequences:

1. `shouldHonorClusterTLSProfile` is evaluated against the new adherence while `NewTLSConfigFromProfile` is applied to the old profile, so the webhook can start with a stale cipher set while logging compliance.
2. The watcher baseline becomes a blend of two revisions, so its first reconcile detects a "change" that never happened and restarts the operator - during exactly the maintenance window an admin is watching.

We could fetch once and derive both, which is what the upstream `Reconcile` already does. `GetTLSProfileSpec` and `APIServerName` are both exported, and `configv1` and `client` are already imported here.

```suggestion
	apiServer := &configv1.APIServer{}
	if err := cl.Get(ctx, client.ObjectKey{Name: tlspkg.APIServerName}, apiServer); err != nil {
		log.Error(err, "failed to read APIServer/cluster, using Go defaults")
		return ServerTLS{}
	}

	profile, err := tlspkg.GetTLSProfileSpec(apiServer.Spec.TLSSecurityProfile)
	if err != nil {
		log.Error(err, "failed to resolve TLS profile, using Go defaults")
		return ServerTLS{}
	}

	adherence := apiServer.Spec.TLSAdherence
```

pkg/deploy/tls/server_tls.go
```diff
@@ -0,0 +1,125 @@
...
+	if shouldHonorClusterTLSProfile(adherence) {
+		tlsConfigFn, unsupported := tlspkg.NewTLSConfigFromProfile(profile)
+		if len(unsupported) > 0 {
+			log.Info("TLS profile contains ciphers unsupported by Go", "unsupported", unsupported)
+		}
+
+		serverTLS.TLSOpts = []func(*cryptotls.Config){tlsConfigFn}
+
+		log.Info(
+			"Applying cluster TLS profile to metrics and webhook servers",
```
tolusha: `Metrics.TLSOpts` is inert because `SecureServing` is false - see the reply on @rohanKanojia's thread at `cmd/main.go:255`. So the profile reaches the webhook server only.

This log line is the operator's single observable claim of TLS compliance. An SRE grepping logs to confirm `StrictAllComponents` adherence gets a false positive from it. Worth fixing regardless of how the metrics question is resolved.

```suggestion
			"Applying cluster TLS profile to the webhook server",
```

pkg/deploy/tls/server_tls.go
```diff
@@ -0,0 +1,125 @@
...
+		InitialTLSProfileSpec:     serverTLS.InitialTLSProfileSpec,
+		InitialTLSAdherencePolicy: serverTLS.InitialTLSAdherencePolicy,
+		OnProfileChange: func(_ context.Context, old, new configv1.TLSProfileSpec) {
+			log.Info("TLS security profile changed, restarting operator")
+			onCancel()
+		},
+		OnAdherencePolicyChange: func(_ context.Context, old, new configv1.TLSAdherencePolicy) {
+			log.Info("TLS adherence policy changed, restarting operator")
+			onCancel()
+		},
```
tolusha: Registration is gated on `profileFetched`, not on whether we are actually honoring the profile. On the default configuration of every existing OpenShift cluster (`tlsAdherence` unset), editing `APIServer.spec.tlsSecurityProfile` restarts the operator for a profile it is not using.

That matters more than the wasted cycles because of the blast radius: the operator runs at `replicas: 1` (`config/manager/manager.yaml:25`) and both webhooks are `failurePolicy: Fail`, so every restart is a window where all `CheCluster` creates and updates are rejected by the API server.

The `old`/`new` parameters the callback already receives make the gate cheap. Reading the startup adherence is correct here - `OnAdherencePolicyChange` handles the transition *into* strict separately, so a cluster moving to `StrictAllComponents` still restarts and picks the profile up on the way back.

This also drops the `new` parameter name, which shadows the Go builtin and is flagged by `revive`/`predeclared`.

```suggestion
		OnProfileChange: func(_ context.Context, _, newSpec configv1.TLSProfileSpec) {
			if !shouldHonorClusterTLSProfile(serverTLS.InitialTLSAdherencePolicy) {
				log.V(1).Info("Cluster TLS profile changed but adherence policy is not strict, not restarting")
				return
			}
			log.Info("TLS security profile changed, restarting operator", "minTLSVersion", newSpec.MinTLSVersion)
			onCancel()
		},
		OnAdherencePolicyChange: func(_ context.Context, _, newPolicy configv1.TLSAdherencePolicy) {
			log.Info("TLS adherence policy changed, restarting operator", "adherencePolicy", newPolicy)
			onCancel()
		},
```

pkg/deploy/tls/server_tls.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return watcher.SetupWithManager(mgr)
+}
+
+// shouldHonorClusterTLSProfile returns true when tlsAdherence requires strict adherence.
+// Unknown values return true for forward compatibility.
+func shouldHonorClusterTLSProfile(adherence configv1.TLSAdherencePolicy) bool {
```
tolusha: `libgocrypto.ShouldHonorClusterTLSProfile` is identical to this - same switch, same cases, same forward-compatible `default: return true`. It is already in the tree as of this PR, at `vendor/github.com/openshift/library-go/pkg/crypto/tls_adherence.go:16`.

The upstream API type asks implementors to use it rather than reimplement:

```go
// vendor/github.com/openshift/api/config/v1/types_apiserver.go:273-275
// TLSAdherencePolicy defines which components adhere to the TLS security profile.
// Implementors should use the ShouldHonorClusterTLSProfile helper function from library-go
// rather than checking these values directly.
```

Dropping the local copy keeps us aligned if OpenShift ever changes the semantics of a policy value. It needs a `go mod tidy` to promote `library-go` from `// indirect` to a direct require, and the call sites become:

```go
import libgocrypto "github.com/openshift/library-go/pkg/crypto"

if libgocrypto.ShouldHonorClusterTLSProfile(adherence) {
```

pkg/deploy/tls/server_tls.go
```diff
@@ -0,0 +1,125 @@
...
+	switch adherence {
+	case configv1.TLSAdherencePolicyNoOpinion, configv1.TLSAdherencePolicyLegacyAdheringComponentsOnly:
+		return false
+	default:
+		return true
+	}
+}
```
tolusha: No `server_tls_test.go` in the diff. The package already has the harness - `init_test.go` calls `infrastructure.InitializeForTesting(infrastructure.OpenShiftV4)`, which is exactly the global that gates line 42, and `certificates_test.go` demonstrates the fake-client patterns. So the first test is close to free, modulo the `client.Client` injection from the comment on line 41.

The cases worth pinning, roughly in order of value:

| Function | Case | What it protects |
|---|---|---|
| `shouldHonorClusterTLSProfile` | `""`, `LegacyAdheringComponentsOnly` | The default on every existing cluster - must be `false` |
| `shouldHonorClusterTLSProfile` | `StrictAllComponents`, unknown value | The feature's purpose, and the deliberate fail-closed default |
| `BuildServerTLSOptions` | `APIServer` absent / unreadable | Pins whatever we decide on line 41 |
| `BuildServerTLSOptions` | strict + `Modern` profile | Asserts `CipherSuites` is **not** set - `NewTLSConfigFromProfile` skips it when `minVersion == VersionTLS13`, a correct but non-obvious branch a library bump could silently change |
| `BuildServerTLSOptions` | strict + `Old` profile | Asserts `MinVersion == VersionTLS10`, making the downgrade explicit |

One note on assertion shape - `assert.NotNil(t, serverTLS.TLSOpts)` would not be worth writing. The meaningful check applies the returned function and inspects the result:

```go
cfg := &cryptotls.Config{}
for _, opt := range got.TLSOpts {
	opt(cfg)
}
assert.Equal(t, uint16(cryptotls.VersionTLS12), cfg.MinVersion)
```

The manual OCP verification and the published gist are genuinely good diligence - this is just the part that regresses silently.

cmd/main.go
```diff
@@ -221,6 +223,8 @@ func main() {
 
 	config := ctrl.GetConfigOrDie()
 
+	serverTLS := tls.BuildServerTLSOptions(context.Background(), config, scheme, setupLog)
```
tolusha: `context.Background()` has no deadline, and `ctrl.GetConfigOrDie()` does not set `rest.Config.Timeout` - controller-runtime's `config.GetConfig` only adjusts `QPS`/`Burst`. Two sequential `Get` calls run here, before `ctrl.NewManager`, so `/healthz:6789` is not listening yet.

From the probe config (`config/manager/manager.yaml:118-126`: delay 15s, period 10s, `failureThreshold: 10`), a hung API server costs roughly 115 seconds before the kubelet kills the container, then repeats into `CrashLoopBackOff`. It self-heals, but slowly and with a misleading symptom - connection refused on healthz rather than anything mentioning TLS.

```suggestion
	tlsCtx, cancelTLSFetch := context.WithTimeout(context.Background(), 30*time.Second)
	serverTLS := tls.BuildServerTLSOptions(tlsCtx, config, scheme, setupLog)
	cancelTLSFetch()
```

cmd/main.go
```diff
@@ -331,7 +349,7 @@ func main() {
 
 	// Start the Cmd
 	setupLog.Info("starting manager")
-	if err := mgr.Start(sigHandler); err != nil {
+	if err := mgr.Start(ctx); err != nil {
```
tolusha, line 355: Two things worth confirming about the restart path that `onCancel` triggers.

**Exit code.** Cancelling `ctx` makes `mgr.Start` return `nil`, `main` returns, and the process exits 0 - indistinguishable from a normal SIGTERM shutdown in `kubectl describe pod` (`Reason: Completed`), in exit-code metrics, and in restart dashboards. The only evidence is a log line in the previous container's output. An explicit non-zero exit with a "restarting to reload TLS configuration" message would make the restart attributable.

**Grace period.** More importantly, cancelling the derived context bypasses the `terminationPeriod` sleep in `signal.SetupSignalHandler`, which exists specifically for finalizers:

```go
// pkg/common/signal/signal_handler.go:45-51
// We need provide more time for operator controllers go routing to complete finalizers logic.
// Otherwise resource won't be clean up gracefully
// and Che custom resource will stay with non empty "finalizers" field.
time.Sleep(time.Duration(terminationPeriod) * time.Second)
```

If a `CheCluster` deletion is mid-finalizer when a TLS profile change lands, that is the exact failure mode this sleep was added to prevent. Does it make sense to route the TLS restart through the same grace path?

On the positive side, `LeaderElectionReleaseOnCancel: true` means the lease is released immediately rather than leaving the next leader to wait out the 40s `leaseDuration` - that keeps the restart window genuinely short.

cmd/main.go
tolusha: Confirming this from the vendored copy in-tree. `SecureServing` is unset repo-wide (`grep -rn SecureServing --include=*.go . | grep -v vendor/` returns nothing), and the metrics server returns before `TLSOpts` is ever read:

```go
// vendor/sigs.k8s.io/controller-runtime/pkg/metrics/server/server.go:273-282
func (s *defaultServer) createListener(ctx context.Context, log logr.Logger) (net.Listener, error) {
	if !s.options.SecureServing {
		return s.options.ListenConfig.Listen(ctx, "tcp", s.options.BindAddress)   // <- returns here
	}
	cfg := &tls.Config{NextProtos: []string{"h2"}}
	for _, op := range s.options.TLSOpts {                                        // <- never reached
		op(cfg)
	}
```

So `Metrics.TLSOpts` is currently dead - metrics stay cleartext on `:60000` (`config/manager/manager.yaml:49`), and on a `StrictAllComponents` cluster that endpoint has no TLS to constrain.

Deferring the work seems reasonable, but two things would be good to settle before merge:

1. The PR description says this "Configures operator webhook **and metric server** to apply the cluster TLS profile." Could we amend it so the delivered scope is accurate?
2. Setting an option that has no effect will read as working coverage to the next person. Either drop `Metrics.TLSOpts` until `SecureServing` lands, or add a `TODO` saying it is inert - and either way the log line at `pkg/deploy/tls/server_tls.go:79` should stop claiming metrics coverage.

For whenever the follow-up happens: enabling `SecureServing` also needs serving certs, a Service, and reconciling `config/prometheus/monitor.yaml`, which already declares `scheme: https, port: https` against a port the Deployment does not define and is not wired into any kustomization.

pkg/deploy/tls/server_tls.go
tolusha: @dkwon17 your reading of `cipherCodes` is right - it filters unsupported ciphers rather than erroring, so continuing is fine for a partially-supported profile.

The case that is not covered is when *every* cipher gets filtered out. `cipherCodes` then returns a nil slice, `NewTLSConfigFromProfile` assigns `tlsConf.CipherSuites = nil`, and Go silently falls back to its **default** cipher list - which may be broader than the profile the cluster demanded. So the one scenario where we most need to know something went wrong currently produces a single `Info` line and a config that quietly does not match policy.

Rather than returning early on any `unsupported`, would it work to keep going but escalate when nothing survives?

```go
tlsConfigFn, unsupported := tlspkg.NewTLSConfigFromProfile(profile)
if len(unsupported) > 0 {
	log.Info("TLS profile contains ciphers unsupported by Go", "unsupported", unsupported)
}
if len(profile.Ciphers) > 0 && len(unsupported) == len(profile.Ciphers) {
	log.Error(nil, "no ciphers from the cluster TLS profile are supported by Go; server will use Go defaults, which may not satisfy tlsAdherence",
		"profileCiphers", profile.Ciphers)
}
```

That keeps the permissive behaviour you described for the common case and only raises the level when the result is genuinely non-compliant. It is only reachable via a `Custom` profile, so it is an edge case - but it is the edge case where silence is most expensive.

pkg/deploy/tls/server_tls.go
dkwon17: Implemented in a0b86cf3c56f7ca866dfc770564bd4a6d42666df

pkg/deploy/tls/server_tls.go
dkwon17: Test suite added in a0b86cf3c56f7ca866dfc770564bd4a6d42666df

cmd/main.go
dkwon17: I opted to send the SIGTERM signal so that the termination logic in the signal handler can be run (f23ba723b3c445414a4e14639e28baabe92b15ae)

dkwon17:
> cmd/main.go:73 - the tls alias is redundant; the package is already named tls and crypto/tls is not imported in main.go.

> pkg/deploy/tls/server_tls.go:81 - logging the full cipher list at Info on every startup is verbose; V(1) might suit it better. MinTLSVersion at Info is useful.

> RegisterSecurityProfileWatcher returns nil when it skips registration, so the caller cannot distinguish "watcher installed" from "watcher skipped because the fetch failed". One log.Info on that branch would close the gap.

All three have been fixed in a5888cb48f5b4a138a712740bfb2dc0d844f05e1

dkwon17:
Closed in favour of https://github.com/eclipse-che/che-operator/pull/2203

PR #2190, opened 2026-09-03
fix: resolve oauth-proxy image from cluster release payload for multi-arch support 

### LINKED ISSUE #23895: Che deployment blocked on OpenShift for IBM Z(s390x) — quay.io/openshift/origin-oauth-proxy has no s390x manifest
### Describe the bug

When attempting to **Deploy Eclipse Che** on an **OpenShift cluster running on s390x** worker nodes, the deployment fails before Che becomes operational.

The `che-operator` pod itself starts successfully on the s390x node, and most component images (`che-dashboard`, `che-server`, `configbump`, `kube-rbac-proxy`) correctly carry s390x manifests and pull without issue. The deployment breaks at the `che-gateway` pod, which runs four sidecar containers. Three start successfully, the fourth — the OAuth proxy authentication sidecar — enters `ImagePullBackOff` because its image has no s390x manifest.

The image in question is hardcoded in `eclipse-che/che-operator` at [config/manager/manager.yaml](https://github.com/eclipse-che/che-operator/blob/cd0093d37fc488198efcf5a4a8658e7f96ee5387/config/manager/manager.yaml#L77):
```
 - name: RELATED_IMAGE_gateway_authentication_sidecar
   value: quay.io/openshift/origin-oauth-proxy:4.22
```

The [quay.io/openshift/origin-oauth-proxy](https://quay.io/repository/openshift/origin-oauth-proxy?tab=tags) community images are built for `linux/amd64` only. We verified the published tags and they don't carry an s390x manifest. As a result, Che cannot complete deployment on s390x OpenShift regardless of which version of the operator image is used, because this dependency is missing and making Che undeployable on s390x OpenShift.

### Che version

next (development version)

### Steps to reproduce

  1. Provision an OpenShift cluster running on s390x worker nodes
  2. Install DevWorkspace Operator
  3. Apply deploy/deployment/openshift/combined.yaml --server-side
  4. Create a CheCluster CR
  5. Observe che-gateway pod — oauth-proxy container enters ImagePullBackOff

### Expected behavior

All four gateway containers start successfully and the Che deployment reaches a healthy state on s390x.

### Runtime

OpenShift

### Screenshots

Image

Image

### Installation method

other (please specify in additional context)

### Environment

other (please specify in additional context)

### Eclipse Che Logs

```shell
oc describe pod -n eclipse-che che-gateway-cd77699ff-x9w76 | grep -A3 "Failed\|ErrImage\|image:"
  Warning  Failed          2m29s (x3 over 3m9s)   kubelet            Failed to pull image "quay.io/openshift/origin-oauth-proxy:4.22": unable to pull image or OCI artifact: pull image err: choosing an image from manifest list docker://quay.io/openshift/origin-oauth-proxy:4.22: no image found in manifest list for architecture "s390x", variant "", OS "linux"; artifact err: image reference: choose manifest instance: no image found in manifest list for architecture "s390x", variant "", OS "linux"
  Warning  Failed          2m29s (x3 over 3m9s)   kubelet            Error: ErrImagePull
  Normal   BackOff         110s (x7 over 3m4s)    kubelet            Back-off pulling image "quay.io/openshift/origin-oauth-proxy:4.22"
  Warning  Failed          110s (x7 over 3m4s)    kubelet            Error: ImagePullBackOff
```

### Additional context

**Installation method detail:**
Direct manifest apply via `oc apply --server-side -f deploy/deployment/openshift/combined.yaml` from [eclipse-che/che-operator](https://github.com/eclipse-che/che-operator/tree/main)

**Environment detail:**
OpenShift cluster running on IBM Z (s390x) hardware. Worker node architecture confirmed as `s390x` via `oc get node`.

**Suggested fix — long-term:**
  
Rather than hardcoding a community image reference for the OpenShift oauth-proxy sidecar, the operator could resolve this image dynamically from the OpenShift cluster's own release payload during CheCluster reconciliation: `oc adm release info --image-for=oauth-proxy`

Every OpenShift cluster carries an architecture-native oauth-proxy image in its release payload. Using this reference ensures the correct image is used on any architecture the cluster runs on — s390x, amd64, arm64, ppc64le — without maintaining per-architecture image references or depending on any external community image. This approach is consistent with how Red Hat's downstream Dev Spaces product handles this dependency.

**Why this fix:**
The [quay.io/openshift/origin-oauth-proxy](https://quay.io/repository/openshift/origin-oauth-proxy?tab=tags) images are built by OpenShift's community CI pipeline for `linux/amd64` only. This source repo ([openshift/oauth-proxy](https://github.com/openshift/oauth-proxy)) is maintained primarily through automated bot PRs with minimal human review activity is what we saw and is a forked repository too, so we are not sure if it is a practical path to add s390x build support for [openshift/oauth-proxy](https://github.com/openshift/oauth-proxy). The fix therefore needs to live in [eclipse-che/che-operator](https://github.com/eclipse-che/che-operator) as suggested above. 

Please guide us on how to proceed.

swapnilsingh-ibm:
### What does this PR do?
Fixes `che-gateway` deployment on OpenShift on IBM Z (s390x) where the oauth-proxy sidecar enters `ImagePullBackOff` because `quay.io/openshift/origin-oauth-proxy:4.22`(hardcoded as `RELATED_IMAGE_gateway_authentication_sidecar`) carries only a `linux/amd64` manifest, making Che undeployable on s390x OpenShift.

The operator now resolves the oauth-proxy image dynamically at CheCluster reconciliation time from the cluster's own `openshift/oauth-proxy` ImageStream (namespace `openshift`). This ImageStream is managed by the Cluster Version Operator and always carries an architecture-native, digest-pinned image from the cluster's release payload — accessible from any pod via the node-level global pull secret without additional configuration.

Falls back to `RELATED_IMAGE_gateway_authentication_sidecar` when the ImageStream is unavailable, preserving existing behaviour for non-OCP environments. No RBAC changes required — the operator's ClusterRole already has `image.openshift.io/imagestreams: get, list`. Also extends `build/scripts/olm/release-catalog.sh` to produce `linux/s390x` and `linux/ppc64le` manifests for the OLM catalog image.

_See attached screenshots:_

1. `che-gateway` reaching `4/4 Running` on s390x

2. All pods healthy in `eclipse-che` namespace

3. oauth-proxy container image resolved from cluster release payload

4. Operator log confirming dynamic image resolution

5. All gateway unit tests passing

Fixes [https://github.com/eclipse-che/che/issues/23895](https://github.com/eclipse-che/che/issues/23895)

### How to test this PR?

Deploy the operator on an OpenShift cluster running on a non-amd64 architecture (s390x, ppc64le, or arm64):
 
#### OpenShift
```bash

OPERATOR_IMAGE=<...>
sed 's|quay.io/eclipse/che-operator:next|'${OPERATOR_IMAGE}'|g' \
  deploy/deployment/openshift/combined.yaml | oc apply --server-side -f -

oc wait checluster eclipse-che -n eclipse-che \
  --for=jsonpath='.status.chePhase'=Active --timeout=120s
```

Verify:

- Must show quay.io/openshift-release-dev/... not quay.io/openshift/origin-oauth-proxy:4.22
```bash
oc get deployment che-gateway -n eclipse-che \
  -o jsonpath='{.spec.template.spec.containers[?(@.name=="oauth-proxy")].image}'
```

- Must show 4/4 Running
```bash
oc get pods -n eclipse-che -l app=che-gateway
```

#### Common Test Scenarios
- [x] Deploy Eclipse Che — verified on OpenShift 4.22 on IBM Z (s390x)
- [ ] Stop a workspace

### PR Checklist

### Reviewers


tolusha:
@swapnilsingh-ibm 
Thank you for contribution. I will take a look

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -175,6 +179,52 @@ func oauthScopeConfig(instance *chev2.CheCluster) string {
 	return ""
 }
 
+// resolveOpenShiftOAuthProxyImage returns the oauth-proxy image from the cluster's own release
+// payload via the openshift/oauth-proxy ImageStream, which is maintained by the Cluster Version
+// Operator and always carries an architecture-native, digest-pinned reference. Returns an empty
+// string when the ImageStream is unavailable so the caller can fall back to the operator default.
+func resolveOpenShiftOAuthProxyImage(ctx *chetypes.DeployContext) string {
```
tolusha, line 186: What about resolving image ref from the internal openshift registry, for instance:
`image-registry.openshift-image-registry.svc:5000/openshift/oauth-proxy@<...>`

Otherwise it won't work in disconnected environment.

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -184,7 +234,12 @@ func getOauthProxyContainerSpec(ctx *chetypes.DeployContext) corev1.Container {
 	var image, probePath string
 	var args = []string{"--config=/etc/oauth-proxy/oauth-proxy.cfg"}
 	if infrastructure.IsOpenShiftOAuthEnabled() {
-		image = defaults.GetGatewayOpenShiftAuthenticationSidecarImage(ctx.CheCluster)
+		// Prefer the architecture-native image from the cluster's release payload;
+		// fall back to RELATED_IMAGE_gateway_authentication_sidecar when unavailable.
+		image = resolveOpenShiftOAuthProxyImage(ctx)
```
tolusha, line 239: Let's guard this with:
`if defaults.GetCheFlavor() == "che"  { ... }`

 since downstream already has different images.

pkg/deploy/gateway/oauth_proxy.go
swapnilsingh-ibm: Fixed. The function now reads `status.dockerImageRepository` and constructs the reference as `{internalRegistry}@{digest}`, which uses the cluster's internal registry -- accessible in all OCP environments including disconnected clusters.

pkg/deploy/gateway/oauth_proxy.go
swapnilsingh-ibm: Fixed. Added if `defaults.GetCheFlavor() == "che"` guard around the dynamic resolution so downstream builds are not affected.

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -175,6 +179,60 @@ func oauthScopeConfig(instance *chev2.CheCluster) string {
...
+		if internalRegistry != "" && digest != "" {
+			ref := internalRegistry + "@" + digest
+			logrus.Infof("Resolved oauth-proxy image from cluster release payload: %s", ref)
+			return ref
+		}
+		ref, _, _ := unstructured.NestedString(firstItem, "dockerImageReference")
+		if ref != "" {
+			logrus.Infof("Resolved oauth-proxy image from cluster release payload: %s", ref)
+			return ref
+		}
```
tolusha: Let's not resolve public image since it can break disconnected env.

swapnilsingh-ibm: Ok, Fixed.

tolusha:

Task completed.

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -175,6 +179,56 @@ func oauthScopeConfig(instance *chev2.CheCluster) string {
...
+		Version: "v1",
+		Kind:    "ImageStream",
+	})
+
+	if err := ctx.ClusterAPI.NonCachingClient.Get(
+		context.TODO(),
+		types.NamespacedName{Name: "oauth-proxy", Namespace: "openshift"},
+		imageStream,
+	); err != nil {
+		logrus.Warnf("Failed to resolve oauth-proxy image from cluster release payload, using default: %v", err)
```
tolusha, line 197: **Transient read failure silently downgrades the running gateway.**

Any error from this `Get` returns `""`, so the caller falls back to `GetGatewayOpenShiftAuthenticationSidecarImage`, which is the amd64-only default on s390x. `DefaultDeploymentDiffOpts` (`pkg/deploy/deployment.go:43`) does not ignore `Container.Image`, so one throttled or timed-out API call rewrites the gateway Deployment and rolls the sidecar into `ImagePullBackOff`. The next successful reconcile flips it back, which can produce a rollout loop on exactly the platform this PR is fixing.

Consider distinguishing "not found" from transient errors and only falling back on the former. `NonCachingClientWrapper.GetIgnoreNotFound` (`pkg/common/k8s-client/k8s_client.go:111`) already makes that distinction. On a transient error we could return an error and let the reconcile be requeued, so the previously resolved image stays in place.

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -175,6 +179,56 @@ func oauthScopeConfig(instance *chev2.CheCluster) string {
 	return ""
 }
 
+// resolveOpenShiftOAuthProxyImage resolves the oauth-proxy image from the cluster's own
+// openshift/oauth-proxy ImageStream. Returns empty string when unavailable.
+func resolveOpenShiftOAuthProxyImage(ctx *chetypes.DeployContext) string {
```
tolusha, line 184: Consider returning `(string, error)` instead of a sentinel empty string. Three distinct outcomes collapse into `""` today: ImageStream absent, API error, and ImageStream present but unusable. The caller cannot tell them apart, which is what makes the transient-failure problem flagged on line 197 possible. Returning `(string, error)` would let the call site fall back on "absent" and requeue on "transient".

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -175,6 +179,56 @@ func oauthScopeConfig(instance *chev2.CheCluster) string {
...
+			continue
+		}
+		firstItem, ok := items[0].(map[string]interface{})
+		if !ok {
+			continue
+		}
+		// Use internal registry reference — works in both connected and disconnected environments.
+		// Do not fall back to dockerImageReference (public image) as it breaks disconnected clusters.
+		digest, _, _ := unstructured.NestedString(firstItem, "image")
+		if internalRegistry != "" && digest != "" {
```
tolusha, line 221: `internalRegistry` does not change inside the loop, so when `status.dockerImageRepository` is empty we still walk every tag and item, then log "ImageStream found but contains no image reference" at line 228 - which points at the tags rather than the actual cause.

Would it make sense to check it up front instead?

```suggestion
		if digest != "" {
```

with an early return added just after the `internalRegistry` lookup:

```go
	if internalRegistry == "" {
		logrus.Warn("openshift/oauth-proxy ImageStream has no status.dockerImageRepository, using default")
		return ""
	}
```

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -175,6 +179,56 @@ func oauthScopeConfig(instance *chev2.CheCluster) string {
...
+	); err != nil {
+		logrus.Warnf("Failed to resolve oauth-proxy image from cluster release payload, using default: %v", err)
+		return ""
+	}
+
+	// Use the internal registry path — accessible in all OCP clusters including disconnected ones.
+	internalRegistry, _, _ := unstructured.NestedString(imageStream.Object, "status", "dockerImageRepository")
+
+	tags, _, _ := unstructured.NestedSlice(imageStream.Object, "status", "tags")
+	for _, tag := range tags {
```
tolusha, line 205: Tag selection is non-deterministic: the loop takes the first tag that happens to carry items, with no preference for a particular one. `openshift/oauth-proxy` normally carries a single tag today, but `status.tags` ordering is not part of the API contract. If a cluster ever exposes more than one tag, the operator could pick a different image across upgrades with no visible reason.

Consider selecting an explicit tag name, or documenting why "first tag wins" is acceptable here.

tolusha (commented):
## Automated review - resolving oauth-proxy from the cluster ImageStream

Thanks for iterating on this. The three earlier threads (internal registry reference, flavor guard, no public-image fallback) all look properly addressed in `edaf643e` and `ce604e6d`, and the approach is a genuine improvement over shipping an amd64-only sidecar reference on s390x. RBAC is already in place: `config/rbac/cluster_role.yaml:165` grants `get, list` on `image.openshift.io/imagestreams`, so the "no RBAC changes required" note in the description checks out.

Overall: **request changes**, on one correctness issue and one behavioural regression. Everything else is a suggestion.

**Blocking**

1. A transient ImageStream read failure downgrades the running gateway to the amd64 default and can flap the Deployment (inline on `oauth_proxy.go:197`).
2. `spec.containerRegistry` overrides are silently bypassed on the dynamic path (inline on `oauth_proxy.go:243`).

**General points that did not fit on a line**

- **No signal when resolution falls back.** When the ImageStream is missing or unusable, the operator logs a warning and quietly uses the default. On s390x that means an `ImagePullBackOff` whose root cause is only visible in operator logs. A `CheCluster` status condition or a Kubernetes Event would make the degraded state visible to `kubectl describe`.
- **Disconnected installs need a documentation update.** Pullthrough of `openshift/oauth-proxy` requires the image to be mirrored. Since the resolved digest is not in the CSV `relatedImages` and never passes through `addDigests.sh`, standard `oc adm catalog mirror` output will not include it. The disconnected-install docs should call this out explicitly.
- **The CSV no longer describes what runs.** `RELATED_IMAGE_gateway_authentication_sidecar` stays in the CSV while the pod may run a different image. Anything auditing deployed images against `relatedImages` will report a mismatch.
- **Rollback is asymmetric.** Downgrading the operator after this change reintroduces the `ImagePullBackOff` on s390x. Worth a release note.
- **`opm:latest` is unpinned** in `olm-catalog/index.Dockerfile:14`. With two more architectures in the catalog build, an upstream `opm` change now breaks four platforms at once rather than two. Pinning it would make the multi-arch build reproducible. Also expect catalog build time to roughly double, since s390x and ppc64le run under QEMU emulation.
- **Existing arch-specific mechanism.** `GetArchitectureDependentEnvName` (`pkg/common/utils/env.go:87`) already supports `RELATED_IMAGE_gateway_authentication_sidecar_s390x`. That is not an argument against this PR (it needs a per-arch image built and maintained), but mentioning in the description why it was not sufficient would help reviewers understand the tradeoff.
- **Revisiting `dockerImageReference`.** Following up on https://github.com/eclipse-che/che-operator/pull/2190#discussion_r4024840481 - dropping it entirely is right for disconnected clusters, but it also removes the only option for a connected cluster that has `ImageRegistry` disabled. A conditional second fallback (use it only when the internal registry is unavailable *and* the cluster is not disconnected) could cover that case, if it is one you care about.

**Verification note:** no Go toolchain is available in this environment, so `go build` and `go test ./pkg/deploy/gateway/...` were not run. All findings come from source reading against `main` and the OpenShift ImageStream / capability documentation.

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -175,6 +179,56 @@ func oauthScopeConfig(instance *chev2.CheCluster) string {
...
+		tagMap, ok := tag.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		items, _, _ := unstructured.NestedSlice(tagMap, "items")
+		if len(items) == 0 {
+			continue
+		}
+		firstItem, ok := items[0].(map[string]interface{})
```
tolusha, line 214: `items[0]` assumes the newest image is first. That ordering is a convention rather than a contract, and an item without an `image` field makes the whole tag unusable even when a later item has one. Iterating over `items` rather than indexing `[0]` is a small change that removes both assumptions.

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -175,6 +179,56 @@ func oauthScopeConfig(instance *chev2.CheCluster) string {
...
+		if !ok {
+			continue
+		}
+		// Use internal registry reference — works in both connected and disconnected environments.
+		// Do not fall back to dockerImageReference (public image) as it breaks disconnected clusters.
+		digest, _, _ := unstructured.NestedString(firstItem, "image")
+		if internalRegistry != "" && digest != "" {
+			ref := internalRegistry + "@" + digest
+			logrus.Infof("Resolved oauth-proxy image from cluster release payload: %s", ref)
```
tolusha, line 223: This logs on every gateway reconcile, not only when the resolved image changes. Consider `Debugf`, or logging only when the value differs from the last resolved reference.

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -175,6 +179,56 @@ func oauthScopeConfig(instance *chev2.CheCluster) string {
...
+// openshift/oauth-proxy ImageStream. Returns empty string when unavailable.
+func resolveOpenShiftOAuthProxyImage(ctx *chetypes.DeployContext) string {
+	imageStream := &unstructured.Unstructured{}
+	imageStream.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "image.openshift.io",
+		Version: "v1",
+		Kind:    "ImageStream",
+	})
+
+	if err := ctx.ClusterAPI.NonCachingClient.Get(
```
tolusha, line 192: `NonCachingClient` issues a live API call on every reconcile for a value that changes at most once per cluster upgrade, and it is the same traffic that makes the transient-failure problem above reachable. Caching the resolved reference for the lifetime of the operator process would address both.

Worth noting `chetypes.ClusterAPI` already carries the hint `We should NonCachingClientWrapper instead wherever it possible` (`pkg/common/chetypes/types.go:36`).

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -175,6 +179,56 @@ func oauthScopeConfig(instance *chev2.CheCluster) string {
...
+		if len(items) == 0 {
+			continue
+		}
+		firstItem, ok := items[0].(map[string]interface{})
+		if !ok {
+			continue
+		}
+		// Use internal registry reference — works in both connected and disconnected environments.
+		// Do not fall back to dockerImageReference (public image) as it breaks disconnected clusters.
+		digest, _, _ := unstructured.NestedString(firstItem, "image")
```
tolusha, line 220: `status.tags[].items[].image` is the Image object name. In practice that is a `sha256:` digest, but the code concatenates it with `@` without checking, so a non-digest value would produce a malformed reference like `registry/ns/repo@v4.4` that only fails at pull time inside the pod, where the cause is hard to trace.

```suggestion
		imageName, _, _ := unstructured.NestedString(firstItem, "image")
		if internalRegistry != "" && strings.HasPrefix(imageName, "sha256:") {
			ref := internalRegistry + "@" + imageName
```

`strings` is already imported in this file.

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -175,6 +179,56 @@ func oauthScopeConfig(instance *chev2.CheCluster) string {
 	return ""
 }
 
+// resolveOpenShiftOAuthProxyImage resolves the oauth-proxy image from the cluster's own
```
tolusha, line 182: The function name, this doc comment, and the messages at lines 197, 223 and 228 all describe the "cluster release payload", but the returned value is an internal registry pullthrough reference. After the change made in https://github.com/eclipse-che/che-operator/pull/2190#discussion_r3968291776 that distinction matters, since the payload reference is deliberately not used. Aligning the wording (for example "resolved from the openshift/oauth-proxy ImageStream via the internal registry") would keep the next reader from reintroducing the `dockerImageReference` fallback.

Minor: the "works in both connected and disconnected environments" claim also appears at lines 201, 218, 219 and again in the test doc comments. One statement of it is enough.

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -175,6 +179,56 @@ func oauthScopeConfig(instance *chev2.CheCluster) string {
...
+	if err := ctx.ClusterAPI.NonCachingClient.Get(
+		context.TODO(),
+		types.NamespacedName{Name: "oauth-proxy", Namespace: "openshift"},
+		imageStream,
+	); err != nil {
+		logrus.Warnf("Failed to resolve oauth-proxy image from cluster release payload, using default: %v", err)
+		return ""
+	}
+
+	// Use the internal registry path — accessible in all OCP clusters including disconnected ones.
```
tolusha, line 201: Internal registry pullthrough depends on the `ImageRegistry` cluster capability. `image-registry.openshift-image-registry.svc:5000` only resolves when that capability is enabled, and the default `system:image-puller` bindings that let a namespace pull from `openshift/` are created by the openshift-controller-manager only under the same capability. On a cluster installed with `baselineCapabilitySet: None` the ImageStream still exists (it is CVO-managed) while the pull fails with an authorization error, and by then the operator has already committed the reference.

Was this combination tested on a capability-trimmed cluster? If it cannot be guaranteed, it would be worth documenting as a prerequisite.

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -184,7 +238,13 @@ func getOauthProxyContainerSpec(ctx *chetypes.DeployContext) corev1.Container {
 	var image, probePath string
 	var args = []string{"--config=/etc/oauth-proxy/oauth-proxy.cfg"}
 	if infrastructure.IsOpenShiftOAuthEnabled() {
-		image = defaults.GetGatewayOpenShiftAuthenticationSidecarImage(ctx.CheCluster)
+		// Guarded to upstream Che only; downstream manages its own oauth-proxy image.
+		if defaults.GetCheFlavor() == "che" {
+			image = resolveOpenShiftOAuthProxyImage(ctx)
```
tolusha, line 243: **`spec.containerRegistry` overrides no longer apply to this container.**

The default path runs through `defaults.PatchDefaultImageName` (`pkg/common/operator-defaults/defaults.go:320`), which applies `spec.containerRegistry.hostname` and `.organization`. The dynamic path bypasses it entirely. The API doc for `CheClusterContainerRegistry` (`api/v2/checluster_types.go:401`) says it overrides the registry in **all** default container images and is "particularly useful for installing Che in a restricted environment", so an administrator who has set it will silently not get it honoured for the oauth-proxy sidecar.

This is the finding I would most like your view on. Options: skip dynamic resolution when `spec.containerRegistry` is set, apply the override to the resolved reference, or document the exception. The per-container escape hatch via `spec.networking.auth.gateway.deployment.containers[]` exists, but it is not discoverable from the `containerRegistry` docs.

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -184,7 +238,13 @@ func getOauthProxyContainerSpec(ctx *chetypes.DeployContext) corev1.Container {
 	var image, probePath string
 	var args = []string{"--config=/etc/oauth-proxy/oauth-proxy.cfg"}
 	if infrastructure.IsOpenShiftOAuthEnabled() {
-		image = defaults.GetGatewayOpenShiftAuthenticationSidecarImage(ctx.CheCluster)
+		// Guarded to upstream Che only; downstream manages its own oauth-proxy image.
+		if defaults.GetCheFlavor() == "che" {
```
tolusha, line 242: `constants.CheFlavor` and `CheCluster.IsCheFlavor()` (`api/v2/checluster_types.go:1315`) already exist for this check.

```suggestion
		if ctx.CheCluster.IsCheFlavor() {
```

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -175,6 +179,56 @@ func oauthScopeConfig(instance *chev2.CheCluster) string {
 	return ""
 }
 
+// resolveOpenShiftOAuthProxyImage resolves the oauth-proxy image from the cluster's own
+// openshift/oauth-proxy ImageStream. Returns empty string when unavailable.
+func resolveOpenShiftOAuthProxyImage(ctx *chetypes.DeployContext) string {
+	imageStream := &unstructured.Unstructured{}
+	imageStream.SetGroupVersionKind(schema.GroupVersionKind{
```
tolusha, line 186: Consider extracting `"oauth-proxy"`, `"openshift"` and the GVK into constants. `kube_rbac_proxy.go` faces the same s390x problem and will likely want the same lookup, so a small shared helper taking a name and namespace would keep the second caller from copying this block.

pkg/deploy/gateway/oauth_proxy_test.go
```diff
@@ -122,3 +124,81 @@ func TestAccessTokenDefinedForKubernetesOauthProxyConfig(t *testing.T) {
...
+
+// TestResolveOpenShiftOAuthProxyImage_ImageStreamAbsent verifies that an empty string
+// is returned when the ImageStream is not present so the caller falls back to the default.
+func TestResolveOpenShiftOAuthProxyImage_ImageStreamAbsent(t *testing.T) {
+	infrastructure.InitializeForTesting(infrastructure.OpenShiftV4)
+
+	ctx := test.NewCtxBuilder().Build()
+
+	resolved := resolveOpenShiftOAuthProxyImage(ctx)
+	assert.Equal(t, "", resolved)
```
tolusha, line 203: The three new tests all exercise `resolveOpenShiftOAuthProxyImage` directly. The behaviour that actually ships is in `getOauthProxyContainerSpec`: the flavor guard and the fallback to the default image when resolution returns `""`. Neither is covered, so a regression that drops the `GetCheFlavor` check or inverts the fallback would still pass.

Two table-driven cases over `getOauthProxyContainerSpec` (flavor `che` with a populated ImageStream, and a non-`che` flavor) would close this.

pkg/deploy/gateway/oauth_proxy_test.go
```diff
@@ -122,3 +124,81 @@ func TestAccessTokenDefinedForKubernetesOauthProxyConfig(t *testing.T) {
...
+					"dockerImageReference": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:503de130e594b7864ab9b63b910d313b4e17cdd09ddd59323729fa221ba1b39c",
+				},
+			},
+		},
+	}, "status", "tags")
+
+	ctx := test.NewCtxBuilder().WithObjects(imageStream).Build()
+
+	resolved := resolveOpenShiftOAuthProxyImage(ctx)
+	assert.Equal(t, "", resolved)
```
tolusha, line 192: This test cannot fail for the reason it is named for. Asserting `""` is also what an entirely empty fixture returns, so it passes whether or not the `dockerImageReference` fallback was removed. The `SetNestedSlice` error is discarded too, so a malformed fixture is indistinguishable from a correct one.

```suggestion
	resolved := resolveOpenShiftOAuthProxyImage(ctx)
	assert.Empty(t, resolved)
	assert.NotContains(t, resolved, "quay.io")
```

pkg/deploy/gateway/oauth_proxy_test.go
```diff
@@ -122,3 +124,81 @@ func TestAccessTokenDefinedForKubernetesOauthProxyConfig(t *testing.T) {
 	assert.Contains(t, config, "pass_access_token = true")
 	assert.NotContains(t, config, "pass_authorization_header = true")
 }
+
+// TestResolveOpenShiftOAuthProxyImage_InternalRegistry verifies that the internal registry
+// reference is preferred — this works in both connected and disconnected OCP environments.
+func TestResolveOpenShiftOAuthProxyImage_InternalRegistry(t *testing.T) {
+	infrastructure.InitializeForTesting(infrastructure.OpenShiftV4)
```
tolusha, line 131: `resolveOpenShiftOAuthProxyImage` never consults the infrastructure type, and `init_test.go:23` already initialises the package to `OpenShiftV4`. These three calls (lines 131, 168, 198) are no-ops for the function under test, but they do flip package-global state that the four pre-existing tests in this file explicitly set to `Kubernetes`, creating an ordering dependency for no benefit.

pkg/deploy/gateway/oauth_proxy_test.go
```diff
@@ -122,3 +124,81 @@ func TestAccessTokenDefinedForKubernetesOauthProxyConfig(t *testing.T) {
...
+			"items": []interface{}{
+				map[string]interface{}{
+					"image":                digest,
+					"dockerImageReference": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@" + digest,
+				},
+			},
+		},
+	}, "status", "tags")
+
+	ctx := test.NewCtxBuilder().WithObjects(imageStream).Build()
```
tolusha, line 158: Worth adding while the fixtures are fresh: empty `status.tags`, a first tag with no items followed by a valid one, a non-map entry in `tags`, and an `image` value that is not a `sha256:` digest.

One fidelity note, not something to fix in this PR: the test scheme does not register `image.openshift.io/v1` (`pkg/common/test/test-client/test_client.go:85`). The fake client keys off the object's own GVK with no RESTMapper, so these tests would still pass if the GVK in the production code were wrong. The GVK is effectively untested.

build/scripts/olm/release-catalog.sh
```diff
@@ -83,7 +83,7 @@ build () {
           CHANNEL="${CHANNEL}" \
           BUNDLE_IMG="${BUNDLE_IMAGE}" \
           IMAGE_TOOL="${IMAGE_TOOL}" \
-          ARCHS="linux/arm64,linux/amd64"
+          ARCHS="linux/arm64,linux/amd64,linux/s390x,linux/ppc64le"
```
tolusha, line 86: Both updated branches are the `MULTI_ARCH == "true"` ones, which `release-next-catalog-and-operator-image.yaml` uses for `next`. The stable path in `build/scripts/release/make-release.sh` pins `BUILDX_PLATFORMS="linux/amd64,linux/arm64"` at line 31 and calls `release-catalog.sh` without `--multi-arch` at line 215, so tagged releases will still ship a two-arch bundle and catalog.

Is the stable path intentionally out of scope here? If so a follow-up issue would help, since the s390x fix only reaches users through a release.

tolusha (approved):

PR #2198, opened 2026-09-16
feat: Add periodic auto-update for internal OpenVSX registry extensions

svor:
### What does this PR do?
Adds a CronJob-based mechanism to periodically check for and update extensions in the internal OpenVSX registry. When enabled via CheCluster CR, a CronJob runs on a configurable schedule (default: weekly) to:

- Compare installed extension versions against the upstream open-vsx.org registry
- Download and publish newer stable (non-pre-release) versions
- Optionally filter by VS Code engine compatibility
- Support excluding specific extensions from auto-update

New CRD fields under `spec.components.openVSXRegistry.extensionAutoUpdate`:
- `enable` (bool) — enables/disables the CronJob
- `schedule` (string) — cron expression (default: `"0 0 * * 0"` = weekly Sunday midnight)
- `vsCodeEngineVersion` (string) — optional engine version filter
- `excludeExtensions` ([]string) — extension IDs to skip (e.g., `"redhat/java"`)

Also fixes:
- **Forwarded headers for cached API URLs**: Passes `X-Forwarded-Host` and `X-Forwarded-Proto` headers to internal OpenVSX API calls from both publish and update Jobs, preventing the server from caching internal service URLs in API responses.
- **ActiveDeadlineSeconds**: Increased from 5m to 30m for publish and update Jobs to support large extension lists.
- **RBAC**: Added `cronjobs` to the `batch` API group in all ClusterRole manifests.
- **Controller watch**: Added `.Owns(&batchv1.CronJob{})` so the controller reconciles on CronJob changes.

Screenshot 2026-09-16 at 14 38 42

Screenshot 2026-09-16 at 14 37 39

Screenshot 2026-09-16 at 14 44 31

https://redhat.atlassian.net/browse/CRW-12351

### How to test this PR?

1. Deploy the operator:
 
#### OpenShift
```bash

OPERATOR_IMAGE=<...>
```
#### on Minikube

```bash
```

2. Enable the internal OpenVSX registry and add extensions to the `openvsx-server-extensions` ConfigMap:
```
oc patch checluster eclipse-che -n eclipse-che --type merge -p '{"spec":{"components":{"openVSXRegistry":{"enable":true}}}}'
```
List of extensions:
```
https://open-vsx.org/api/redhat/vscode-xml/0.29.0/file/redhat.vscode-xml-0.29.0.vsix
https://open-vsx.org/api/redhat/vscode-yaml/1.18.0/file/redhat.vscode-yaml-1.18.0.vsix
https://open-vsx.org/api/vscjava/vscode-java-debug/0.58.2/file/vscjava.vscode-java-debug-0.58.2.vsix
https://open-vsx.org/api/redhat/java/1.43.1/file/redhat.java-1.43.1.vsix
https://open-vsx.org/api/vscode/html-language-features/1.95.3/file/vscode.html-language-features-1.95.3.vsix
https://open-vsx.org/api/redhat/vscode-microprofile/0.14.0/file/redhat.vscode-microprofile-0.14.0.vsix
https://open-vsx.org/api/redhat/vscode-quarkus/1.21.0/file/redhat.vscode-quarkus-1.21.0.vsix
```

3. Wait for extensions to be published by the initial Job, then enable auto-update:
```
oc patch checluster eclipse-che -n eclipse-che --type merge -p '{"spec":{"components":{"openVSXRegistry":{"extensionAutoUpdate":{"enable":true,"schedule":"*/5 * * * *"}}}}}'
```

4. Verify the CronJob is created:
```
oc get cronjob -n eclipse-che
```

5. Wait for the CronJob to trigger (should be in 5 minutes):

6. Check the Job logs for update activity

7. Test disabling auto-update — CronJob should be deleted:
```
oc patch checluster eclipse-che -n eclipse-che --type merge -p '{"spec":{"components":{"openVSXRegistry":{"extensionAutoUpdate":{"enable":false}}}}}'
```
```
oc get cronjob -n eclipse-che  # should show no CronJob
```

8. Test with optional fields (engine version filter, exclude list):
```
oc patch checluster eclipse-che -n eclipse-che --type merge -p '{"spec":{"components":{"openVSXRegistry":{"extensionAutoUpdate":{"enable":true,"vsCodeEngineVersion":"1.92.0","excludeExtensions":["redhat/java"]}}}}}'
```

#### Common Test Scenarios
- [ ] Deploy Eclipse Che
- [ ] Stop a workspace

### PR Checklist

### Reviewers

## Summary by CodeRabbit

* **New Features**
  * Added optional automatic extension updates from the upstream OpenVSX registry.
  * Configure update schedules, excluded extensions using namespace/name IDs, and an optional VS Code engine version.
  * Updates are disabled by default and run weekly when enabled.
* **Bug Fixes**
  * Extension publishing jobs now include resource limits and support longer-running operations.
  * Automatic update jobs are removed when the feature or OpenVSX registry is disabled.
* **Platform Support**
  * Added Kubernetes support for managing scheduled extension update jobs.


tolusha:

Task completed.

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
```diff
@@ -0,0 +1,188 @@
...
+									ImagePullPolicy: corev1.PullPolicy(imagePullPolicy),
+									VolumeMounts: []corev1.VolumeMount{
+										{
+											Name:      "extensions",
+											MountPath: "/home/openvsx/extensions",
+											ReadOnly:  true,
+										},
+									},
+									Env:     env,
+									Command: []string{"/home/openvsx/update-extensions.sh", "/home/openvsx/extensions/extensions.list"},
```
tolusha, line 144: **The pinned OpenVSX image does not contain `update-extensions.sh`.** This is the one blocking item.

`config/manager/manager.yaml:85` pins `RELATED_IMAGE_openvsx: quay.io/che-incubator/che-openvsx:v1.1.1`, and `PatchDefaultImageName` only rewrites host and organization - the tag is never substituted, so `v1.1.1` is what runs.

- `v1.1.1` was pushed to quay on 2026-08-10.
- `che/update-extensions.sh` was added in `che-incubator/che-openvsx` commit `c4f13e7f` on 2026-09-16 13:09, shipping in `v1.2.0` / `latest` (pushed 13:14).

So as merged, every scheduled run execs a path that does not exist in the container: the pod fails at startup, `backoffLimit: 3` burns three retries, and this repeats every schedule forever with nothing surfaced on `CheCluster.Status`.

The same commit added forwarded-header support to `publish-extensions.sh`, so the new `OVSX_FORWARDED_*` vars on the publish Job are equally inert in `v1.1.1`.

Would it be possible to bump `RELATED_IMAGE_openvsx` to `v1.2.0` in this PR, or land the image bump first and note the ordering here?

Everything else in the contract checks out against the image source - script path, the `<extensions-file>` argument, `extensions.list`, and all six env var names including the comma-separated `namespace/name` format for `EXCLUDE_EXTENSIONS`.

api/v2/checluster_webhook.go
```diff
@@ -295,6 +295,17 @@ func (r *CheClusterValidator) validateOpenVSXRegistry(checluster *CheCluster) er
 		}
 	}
 
+	if checluster.Spec.Components.OpenVSXRegistry.ExtensionAutoUpdate != nil {
+		autoUpdate := checluster.Spec.Components.OpenVSXRegistry.ExtensionAutoUpdate
+
+		if autoUpdate.Schedule != nil && *autoUpdate.Schedule != "" {
+			fields := strings.Fields(*autoUpdate.Schedule)
+			if len(fields) != 5 {
```
tolusha: **An invalid 5-field cron expression stalls the entire reconcile chain.**

Counting fields accepts `"0 0 * * 8"`, `"0 25 * * *"`, and `"abc def ghi jkl mno"`. The API server then rejects the CronJob (`batch/v1` validates with `robfig/cron.ParseStandard`), `syncExtensionUpdateCronJob` returns an error, `Reconcile` returns `done=false`, and `ReconcilerManager.ReconcileAll` aborts the whole chain at the first `done=false`.

`OpenVSXServerReconciler` is registered before `editorsdefinitions`, `dashboard`, `gateway`, `CheServerReconciler` and `imagepuller`, so a single typo in `schedule` permanently blocks Dashboard, Gateway and Che Server reconciliation.

The same check also rejects schedules Kubernetes accepts - `@weekly`, `@daily`, `@hourly`, `@every 1h`. `@weekly` is exactly equivalent to this feature's own default and is the most natural thing a user would type.

Two options, either works:

- Validate with `robfig/cron/v3` (the parser Kubernetes itself uses). It is not currently in `go.sum` even indirectly, so it would be a new direct dependency.
- Keep the check simple but stop failing the chain on a spec error: log it, surface it on `CheCluster.Status`, and return `done=true` so a bad schedule degrades one feature instead of the whole install.

pkg/common/diffs/diffs.go
```diff
@@ -89,6 +90,10 @@ var ServiceMonitor = cmp.Options{
 	cmpopts.IgnoreFields(monitoringv1.ServiceMonitor{}, "TypeMeta", "ObjectMeta"),
 }
 
+var CronJob = cmp.Options{
```
tolusha, line 94: **`diffs.CronJob` is missing the defaulting exclusions.**

`CronJobSpec` embeds a full `PodSpec` and a `ConfigMapVolumeSource`, so the API server defaults a number of fields the blueprint does not set. Compare `DefaultDeploymentDiffOpts` (`pkg/deploy/deployment.go:43-54`), which has to ignore `DNSPolicy`, `SchedulerName`, `DeprecatedServiceAccount`, `TerminationMessagePath`, `TerminationMessagePolicy` and `ConfigMapVolumeSource.DefaultMode` for exactly this reason.

Two effects:

- Every reconcile produces a non-empty diff, so `doSync` takes the update branch: a full `Difference:` dump plus a misleading `"Object updated"` log line and a no-op `Update`, forever. No hot loop - the API server skips the identical write and does not bump `resourceVersion` - but it is permanent log noise and wasted round-trips.
- `DefaultDeploymentDiffOpts` also carries a `resource.Quantity` comparer. Without it, if a namespace `LimitRange` injects default container resources, `cmp.Diff` hits `resource.Quantity`'s unexported fields and panics.

The fake test client does no API-server defaulting, so no test in the current style can catch either. Could we model this on `DefaultDeploymentDiffOpts`?

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
```diff
@@ -0,0 +1,188 @@
...
+								},
+							},
+							RestartPolicy:                 corev1.RestartPolicyNever,
+							TerminationGracePeriodSeconds: ptr.To(int64(30)),
+						},
+					},
+					Parallelism:             ptr.To(int32(1)),
+					BackoffLimit:            ptr.To(int32(3)),
+					Completions:             ptr.To(int32(1)),
+					TTLSecondsAfterFinished: ptr.To(int32(300)),
```
tolusha: **`ttlSecondsAfterFinished: 300` silently defeats `failedJobsHistoryLimit: 3`** (line 116).

The TTL-after-finished controller deletes any finished Job - Complete or Failed - 300 seconds after it finishes, along with its pods and their logs. So the history limit can never retain anything.

For the one-shot publish Job this is harmless: it runs at install time with an administrator watching. For an unattended weekly CronJob the profile is different - the job fails at 00:00 Sunday and by 00:05 there is no Job, no pod and no log left to look at.

Dropping `TTLSecondsAfterFinished` from the CronJob's JobTemplate and letting the history limits do their job would fix it, or raising it past one schedule interval.

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
```diff
@@ -0,0 +1,188 @@
...
+		schedule = *autoUpdate.Schedule
+	}
+
+	image := defaults.GetOpenVSXImage(ctx.CheCluster)
+	imagePullPolicy := utils.GetPullPolicyFromDockerImage(image)
+
+	labels := deploy.GetLabels(constants.OpenVSXServerExtensionUpdateCronJobName)
+	credentialsSecret := openvsx.GetCredentialsSecretName(ctx)
+
+	env := []corev1.EnvVar{
```
tolusha, line 76: **No proxy, trusted CA, or upstream registry override.**

This is the only Che workload whose core function is egress to a third-party internet endpoint, and the only one that gets no proxy configuration. The pattern is well established elsewhere - `deployment_dashboard.go:61-81`, `gateway/oauth_proxy.go:222-230` and `server/server_configmap.go:118-119` all inject `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` from `ctx.Proxy`, and `tls/certificates.go:146-149` mounts `ctx.Proxy.TrustedCAMapName`. On a cluster behind an authenticated or TLS-intercepting proxy every run is a silent no-op. The script supports `NODE_EXTRA_CA_CERTS` for exactly this.

Related: the script also supports `UPSTREAM_REGISTRY_URL` (default `https://open-vsx.org`), which is not exposed. Disconnected and mirrored installs are the main reason to run an internal registry at all, and today they cannot use this feature - every run fails trying to reach open-vsx.org. Exposing `upstreamRegistryUrl` is a small change and it also gives security-conscious operators a vetted-mirror option.

`MAX_VERSIONS_TO_CHECK` is a third supported knob that is not exposed, though that one matters less.

tolusha (commented):
## Review: periodic auto-update for internal OpenVSX registry extensions

Nice feature, and the wiring is more complete than most - owner reference, `.Owns(&batchv1.CronJob{})`, scheme registration, RBAC across all six manifest copies, and cleanup on **both** the `extensionAutoUpdate.enable=false` and `openVSXRegistry.enable=false` paths, each with its own test. That last part is the bit most feature PRs forget.

I verified the env/path/argument contract against the `che-incubator/che-openvsx` image source rather than inferring it from names, and the contract itself is correct: script path, `<extensions-file>` argument, the `extensions.list` ConfigMap key, and all six env var names including the comma-separated `namespace/name` format for `EXCLUDE_EXTENSIONS`. The `vsCodeEngineVersion`-omitted semantics documented on the CRD field match what `check_engine_compatibility` actually does, which is a good sign.

**One blocking item**, left inline on the `Command` line: the operator pins `che-openvsx:v1.1.1` (pushed 2026-08-10) but `update-extensions.sh` only exists in `v1.2.0` / `latest` (pushed 2026-09-16, from commit `c4f13e7f`). As merged the CronJob execs a path that is not in the container, and the new `OVSX_FORWARDED_*` vars on the publish Job are inert for the same reason. A `RELATED_IMAGE_openvsx` bump would resolve both.

Two more I would look at before merge, also inline: an invalid-but-5-field cron passes the webhook and wedges the whole reconcile chain, and `diffs.CronJob` is missing the API-server defaulting exclusions that `DefaultDeploymentDiffOpts` carries.

### A few things that did not fit on a line

**Rollback leaves an orphaned CronJob.** The owner reference points at the CheCluster, not the operator Deployment. If someone enables auto-update on 7.123.x and then rolls the operator back to 7.122.x - a routine response to an unrelated regression - the CR survives, the CronJob survives, and the older operator has no code that knows it exists. It is not deleted, not reconciled, not visible to any controller, and it keeps firing on schedule. Its `deleteResources` cannot clean it up either, since that code is not in 7.122.x. Might be worth a label plus an orphan sweep, or a documented rollback step. Also worth confirming this is unrelated to the red `v19-upgrade-stable-to-next` before writing that off as flake.

**Nothing reports whether the feature is working.** This is the operator's first unattended recurring worker, and there is no status field, condition, event or metric for it. Combined with the script exiting 0 unless *every* extension fails, `successfulJobsHistoryLimit: 1`, and the TTL issue above, the only way an admin learns that weekly updates have never once succeeded is to run `kubectl get jobs` and notice failures nobody told them to look for. The CronJob is already watched, so `status.lastSuccessfulTime` arrives for free - surfacing it on `CheCluster.Status` would go a long way.

**No jitter on a globally synchronized default.** `"0 0 * * 0"` means every Che install with this enabled hits open-vsx.org, a free Eclipse Foundation service, at exactly 00:00 UTC Sunday. Deriving a stable per-cluster offset from the CheCluster UID would be a good neighbour.

**Duplication with `openvsx_server_extensions.go`.** `openvsx_server_extension_update.go:57-145` is a near-verbatim fork of `openvsx_server_extensions.go:80-155` - roughly 60 lines differing only in `Command` and three env vars. Not hypothetical: this PR had to make the same three edits (forwarded host, forwarded proto, 300 to 1800) in both copies. A shared pod-spec constructor would make the next such change a one-liner, and there will be a next one.

**Test gaps beyond the inline note.** The update path is never exercised - every test either creates or deletes, so nothing changes `schedule` on an existing CronJob and re-reconciles. Given `diffs.CronJob` is new and is the most fragile piece here, that is the path that most needed a test. Also missing: an idempotency check, an owner-reference assertion, a PSS assertion, the empty-string-schedule branch, and anything covering either publish-Job side fix. Worth knowing that the fake client does no API-server defaulting, so the `diffs.CronJob` problem is structurally uncatchable in the current test style regardless of how many tests are added.

**Hardcoded operational knobs.** `activeDeadlineSeconds: 1800`, `backoffLimit`, the TTL, history limits and the absent resource requests are all Go literals. "Our customer has 200 extensions, just bump the timeout" is currently a code change plus a release. Other Che components already expose `deployment.containers[].resources` through the CRD, so there is precedent if you want it.

### Also

- Needs a rebase (`mergeable_state: dirty`).
- Two required Prow jobs are red on `c5594ac1`: `v19-upgrade-stable-to-next` and `v19-devworkspace-happy-path`. All GitHub Actions checks are green.

### Verified and deliberately not raised

`OVSX_FORWARDED_PROTO: "https"` hardcoded (matches repo convention, Che requires TLS); `ctx.CheHost` being empty (`CheHostReconciler` gates the chain earlier); the RBAC verb set (every verb is exercised, and `cluster_role.yaml` is hand-maintained since there are no `+kubebuilder:rbac` markers for `batch`); NetworkPolicy coverage (`app.kubernetes.io/part-of: che.eclipse.org` is set, so both existing policies select the pod); the CSV version bump (normal regeneration); and the CronJob update path (`isRecreate` covers only Service/Ingress/Route/Job, so CronJobs are correctly updated in place).

---

*Automated review. 15 inline comments above.*

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
```diff
@@ -0,0 +1,188 @@
...
+	deploy.EnsurePodSecurityStandards(
+		&cronJob.Spec.JobTemplate.Spec.Template.Spec,
+		constants.DefaultSecurityContextRunAsUser,
+		constants.DefaultSecurityContextFsGroup,
+	)
+
+	return cronJob, nil
+}
+
+func (r *OpenVSXServerReconciler) deleteExtensionUpdateCronJob(ctx *chetypes.DeployContext) error {
```
tolusha: **Disabling deletes the CronJob rather than suspending it.**

`DeletePropagationBackground` garbage-collects the owned Jobs too. If a user sets `enable: false` while an update run is in flight, that Job is killed mid-publish, which could leave the internal registry with a partially applied batch, and all job history goes with it.

`batchv1.CronJobSpec.Suspend` covers this case and is cheaper to reconcile. Could you confirm the update script is safe to interrupt?

pkg/deploy/openvsx/openvsx-server/openvsx_server.go
```diff
@@ -122,6 +122,11 @@ func (r *OpenVSXServerReconciler) Reconcile(ctx *chetypes.DeployContext) (reconc
 		r.extensionsVersion = extensionsVersion
 	}
 
+	err = r.syncExtensionUpdateCronJob(ctx)
```
tolusha: **The CronJob cannot be removed while the server Deployment is unhealthy.**

`syncExtensionUpdateCronJob` runs after the `isServerReady` gate at lines 105-107. If the OpenVSX server Deployment has zero available replicas, reconcile returns early, so setting `extensionAutoUpdate.enable=false` has no effect - the CronJob stays and keeps firing update Jobs against a server that is down.

Deletion is a cleanup action and does not depend on server readiness. Does it make sense to hoist the sync above the readiness gate, or at least handle the disable branch there?

pkg/deploy/openvsx/openvsx-server/openvsx_server.go
```diff
@@ -188,6 +193,19 @@ func deleteResources(ctx *chetypes.DeployContext) {
 		logger.Error(err, "Failed to delete ConfigMap", "Name", constants.OpenVSXServerExtensionsConfigMapName)
 	}
 
+	err = cw.DeleteByKeyIgnoreNotFound(
```
tolusha: This is the same call with the same options as `deleteExtensionUpdateCronJob`. We could just call the helper here.

api/v2/checluster_types.go
```diff
@@ -534,6 +537,31 @@ type OpenVSXDatabase struct {
 	Storage *PVC `json:"pvc,omitempty"`
 }
 
+// Configuration for periodic auto-update of extensions from the upstream open-vsx.org registry.
```
tolusha, line 555: **Worth documenting the trust boundary here.**

Once enabled, this resolves "latest" from open-vsx.org and publishes it into the registry every workspace in the cluster installs extensions from, with no signature verification, no digest pinning, no allowlist (the exclude list is a denylist), no semver constraint and no approval step. A compromised upstream publisher account propagates to every workspace within one schedule interval.

Defaulting `enable` to `false` is the right call and limits the blast radius. The doc comment currently reads as a pure convenience feature though, and this text lands in the CRD, the CSV and the published API reference - it is the highest-leverage place to say what enabling it means.

Worth noting `che-incubator/che-openvsx` has an `ft/open-vsx-sign` branch, so signing looks planned. Until it lands, a sentence here would help.

Separately, worth filing upstream: `check_engine_compatibility` in `update-extensions.sh` interpolates `$engine_constraint` - raw JSON from the upstream registry - unescaped into a `node -e` program. A crafted `engines.vscode` value would execute arbitrary code in a pod that holds `OVSX_PAT`. Passing the values through `process.env` instead of string interpolation would fix it.

api/v2/checluster_types.go
```diff
@@ -534,6 +537,31 @@ type OpenVSXDatabase struct {
...
+	// +optional
+	// +kubebuilder:default:=false
+	Enable bool `json:"enable,omitempty"`
+	// Cron schedule for the auto-update job.
+	// +optional
+	// +kubebuilder:default:="0 0 * * 0"
+	Schedule *string `json:"schedule,omitempty"`
+	// VS Code engine version used to filter compatible extensions.
+	// Only extensions compatible with this engine version will be updated.
+	// When omitted, the latest non-pre-release version of each extension is used regardless of engine compatibility.
```
tolusha, line 570: **Omitting `vsCodeEngineVersion` is arguably not a safe default.**

The doc is accurate about the behaviour, which is what makes the default stand out: out of the box, extensions are published regardless of engine compatibility. Nothing rejects them at publish time, so the failure surfaces much later in a user's workspace as an extension that silently will not activate - about as far from the cause as a failure can get.

The operator already knows the bundled editor version through the editors-definitions reconciler. Would defaulting this from the deployed editor make sense, so the safe behaviour is the default and the permissive one is the opt-in?

api/v2/checluster_types.go
```diff
@@ -534,6 +537,31 @@ type OpenVSXDatabase struct {
...
+	// +optional
+	VSCodeEngineVersion *string `json:"vsCodeEngineVersion,omitempty"`
+	// List of extension IDs to exclude from auto-update.
+	// Each entry should be in the format "namespace/name" (e.g., "redhat/java", "redhat/vscode-xml").
+	// Extensions in this list will be skipped during the update check.
+	// +optional
+	ExcludeExtensions []string `json:"excludeExtensions,omitempty"`
```
tolusha, line 577: **`excludeExtensions` validation and shape.**

Validation: nothing enforces the documented `namespace/name` shape, and an entry containing a comma silently splits into two bogus IDs after `strings.Join(..., ",")`. A typo fails open - the extension gets updated anyway, with no warning. A pattern on the items would catch both at admission:

```suggestion
	// List of extension IDs to exclude from auto-update.
	// Each entry should be in the format "namespace/name" (e.g., "redhat/java", "redhat/vscode-xml").
	// Extensions in this list will be skipped during the update check.
	// +optional
	// +kubebuilder:validation:items:Pattern=`^[a-zA-Z0-9][a-zA-Z0-9-_.]*\/[a-zA-Z0-9][a-zA-Z0-9-_.]*$`
	ExcludeExtensions []string `json:"excludeExtensions,omitempty"`
```

Shape, worth considering while the API is still unreleased: `[]string` can express exactly one policy - "never update this" - and has nowhere to hang attributes. An allowlist, a semver pin ("patch releases only"), or a per-extension engine override would each need a new parallel field rather than an extension of this one. Something like `[]ExtensionPolicy{Id, Mode, VersionRange}` absorbs all of them with today's behaviour as the degenerate case. Changing it now costs nothing; after release it costs a deprecation cycle.

Per-extension policy seems like a plausible near-term ask - it is the natural follow-up the first time an auto-updated extension breaks someone's workspaces.

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	return ctx.ClusterAPI.ClientWrapper.Sync(
+		context.TODO(),
+		cronJob,
+		&k8sclient.SyncOptions{DiffOpts: diffs.CronJob},
+	)
+}
+
+func (r *OpenVSXServerReconciler) getExtensionUpdateCronJobSpec(ctx *chetypes.DeployContext) (*batchv1.CronJob, error) {
```
tolusha, line 62: **Dead error return, and an unguarded dereference.**

`getExtensionUpdateCronJobSpec` has no path that returns a non-nil error, so the `error` in the signature, the `if err != nil` at the call site and the `failed to get extension update CronJob spec: %w` wrapper are all unreachable.

Separately, line 59 dereferences `ExtensionAutoUpdate` unguarded. That is safe from `syncExtensionUpdateCronJob` because of the `IsExtensionAutoUpdateEnabled()` check, but the tests already call this function directly without it. Taking `*ExtensionAutoUpdate` as a parameter would put the precondition in the type.

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update_test.go
```diff
@@ -0,0 +1,329 @@
...
+	cronJob, err := reconciler.getExtensionUpdateCronJobSpec(ctx)
+	if !assert.NoError(t, err) {
+		return
+	}
+
+	assert.Equal(t, customSchedule, cronJob.Spec.Schedule)
+	assert.Equal(t, batchv1.ForbidConcurrent, cronJob.Spec.ConcurrencyPolicy)
+	assert.Equal(t, defaults.GetOpenVSXImage(ctx.CheCluster), cronJob.Spec.JobTemplate.Spec.Template.Spec.Containers[0].Image)
+
+	envMap := make(map[string]string)
```
tolusha, line 197: **These env assertions skip `OVSX_PAT`.**

The `if e.Value != ""` filter drops `OVSX_PAT` entirely, since its value lives in `ValueFrom` - so the secret wiring, the security-relevant part, is never asserted. It also means an assertion like `assert.Equal(t, "https", envMap["OVSX_FORWARDED_PROTO"])` cannot distinguish "correct" from "absent" in the general case.

Asserting over the full `Env` slice and checking `OVSX_PAT.ValueFrom.SecretKeyRef.Name` / `.Key` explicitly would cover it.

Same idea for the absence checks in `...NoExcludeExtensions` and `...NoEngineVersion`: looping `assert.NotEqual` over `container.Env` passes vacuously if `Env` is ever empty. `assert.NotContains` over a collected name set says the same thing and cannot pass vacuously.

controllers/che/checluster_controller.go
```diff
@@ -179,6 +180,7 @@ func (r *CheClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&rbacv1.RoleBinding{}).
 		Owns(&corev1.ServiceAccount{}).
 		Owns(&appsv1.Deployment{}).
+		Owns(&batchv1.CronJob{}).
```
tolusha, line 186: **Worth a watch predicate here.**

kube-controller-manager writes CronJob `.status` on every schedule tick (`lastScheduleTime`), on each Job start (`.status.active`) and on each completion - roughly three status writes per run, each enqueuing a full CheCluster reconcile that walks the entire chain.

At the weekly default that is negligible. On the `*/5` schedule the testing instructions suggest it is around 36 extra full-chain reconciles an hour, each also issuing the no-op `Update` from the `diffs.CronJob` comment.

Every other `Owns()` here is unfiltered too, so this follows convention - but CronJob is the first owned type whose status is mutated on a timer by an external controller rather than only in response to operator action. `builder.WithPredicates(predicate.GenerationChangedPredicate{})` on this one entry would filter the status churn.

helmcharts/stable/templates/che-operator.ClusterRole.yaml
```diff
@@ -24,6 +24,7 @@ rules:
   - batch
   resources:
   - jobs
+  - cronjobs
```
tolusha: **This file looks hand-edited rather than generated.**

`make update-dev-resources` only runs `update-helmcharts CHANNEL=next` (`Makefile:140`), and `check-resources.sh` validates only that - which is why `resources-validation` is green despite this file being outside its scope. Every prior change to it came from release automation (`ci: Copy 7.121.0 csv to main`, etc.).

The effect is that the stable chart grants `batch/cronjobs` to a released operator (7.122.x) that has no CronJob code, and the edit will be clobbered at the next release copy. Unless it is needed for something specific like getting `v19-upgrade-stable-to-next` to pass, this hunk could come out of the PR.

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
svor: the  image is going to be updated by https://github.com/eclipse-che/che-operator/pull/2200

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
svor: yes, the update script is safe to interrupt

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
svor: For clusters behind a corporate proxy, the cluster admin should be able to configure the cluster-wide proxy (e.g., the Proxy CR on OpenShift) to allow outbound HTTPS to open-vsx.org. In case we have additional request to improve it, we can address it in another PR

api/v2/checluster_types.go
svor: The engine_constraint value comes from the same upstream registry that provides the extension binaries. If the upstream metadata is compromised, the extension code itself is the bigger concern. This is out of scope for this PR in any case

api/v2/checluster_types.go
svor: The editor definition is a devfile that specifies the container image - it doesn't expose the VS Code engine version. That version is embedded in the editor binary inside the image and isn't available to the operator without running the container or parsing image layers.

Also, different workspaces can use different editor versions, so there's no single "correct" engine version to default to. It's the admin's choice based on their deployment - which is why we expose it as an optional field.

api/v2/checluster_types.go
svor: The doc comment already describes the expected format. Will consider adding validation pattern as a follow-up. The []ExtensionPolicy shape is premature - prefer to evolve based on real feedback after the feature ships.

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
svor: The dereference is safe because the only caller (syncExtensionUpdateCronJob) guards with IsExtensionAutoUpdateEnabled(), and all tests set ExtensionAutoUpdate explicitly. Will add a nil guard if the function gets more callers.

controllers/che/checluster_controller.go
svor: Valid observation. As noted, all other Owns() calls in the controller are unfiltered - adding a predicate only to CronJob would break that consistency. The extra reconciles are no-ops and inexpensive. Happy to add the predicate if we see it become a problem in practice, but for now prefer to keep it consistent with the existing pattern.

helmcharts/stable/templates/che-operator.ClusterRole.yaml
svor: removed

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update_test.go
svor: fixed

pkg/deploy/openvsx/openvsx-server/openvsx_server.go
svor: Changed deleteExtensionUpdateCronJob from a method to a package-level function and replaced the duplicated block in deleteResources with a call to it.

pkg/deploy/openvsx/openvsx-server/openvsx_server.go
svor: Moved syncExtensionUpdateCronJob above the isServerReady gate so CronJob creation/deletion works regardless of server health.

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
svor: Removed TTLSecondsAfterFinished from the CronJob spec — the CronJob's history limits (successfulJobsHistoryLimit: 1, failedJobsHistoryLimit: 3) now handle cleanup, keeping failed Job logs available for diagnosis.

pkg/common/diffs/diffs.go
svor: pkg/common/diffs/diffs.go — diffs.CronJob now ignores API-server-defaulted PodSpec fields (DNSPolicy, SchedulerName, DeprecatedServiceAccount, TerminationMessagePath, TerminationMessagePolicy, ConfigMapVolumeSource.DefaultMode, SecretVolumeSource.DefaultMode) and includes a resource.Quantity comparer to prevent panics when a LimitRange injects default container resources.

api/v2/checluster_webhook.go
svor: 1. Removed cron schedule validation from the webhook (checluster_webhook.go) — Kubernetes itself validates the cron expression when the CronJob is created, so the webhook check was both too weak (accepted invalid values) and too strict (rejected @weekly, @daily, etc.).
2. Made CronJob sync non-blocking in the reconciler (openvsx_server.go) — if syncExtensionUpdateCronJob fails (e.g., bad schedule rejected by the API server), it logs the error and continues with done=true instead of blocking the entire reconcile chain. A bad schedule degrades only the auto-update feature, not Dashboard, Gateway, or CheServer.

tolusha:
/che-ai-assistant claude Check if changes meets the claude rules of this repository, if not create a PR

Task completed.

tolusha:
## Claude rules compliance review

Reviewed this PR against the repository's agent rules (`CLAUDE.md` → `AGENTS.md` → `.claude/rules/*.md`) at commit `b5127157b`.

**Result:** 5 deviations found. Fixes are proposed in **#2202**, which targets this PR's branch (`sv-update-extensions`) so it can be merged in before this PR lands.

### Findings

| # | Rule | Severity | Issue |
|---|------|----------|-------|
| 1 | `kubebuilder-defaults.md` | **functional** | Defaults declared but never applied |
| 2 | `code-style.md` | medium | Errors returned without `fmt.Errorf` context |
| 3 | `k8s-client.md` | medium | Deprecated `deploy.Get*` and raw `Client.Get` in tests |
| 4 | `code-style.md` | low | Six added lines exceed 120 characters |
| 5 | `redhat-compliance.md` | low | Copyright year not bumped on an edited file |

---

**1. `kubebuilder-defaults.md` — defaults are never applied** (`api/v2/checluster_types.go`)

> "When adding a `+kubebuilder:default` annotation to a field, propagate the default up to every parent struct's `+kubebuilder:default` annotation as well."

`ExtensionAutoUpdate.Enable` (`false`) and `ExtensionAutoUpdate.Schedule` (`"0 0 * * 0"`) both carry defaults, but the parent pointer field has none:

```go
// Configuration for periodic auto-update of extensions from the upstream open-vsx.org registry.
// +optional
ExtensionAutoUpdate *ExtensionAutoUpdate `json:"extensionAutoUpdate,omitempty"`
```

and the grandparent `openVSXRegistry` default is still `{enable: false}`. This is the exact case the rule exists for — the generated CRD confirms the nested defaults sit under an object with no `default:` of its own, so the API server never instantiates `extensionAutoUpdate` and neither default ever materializes. In practice this means `spec.components.openVSXRegistry.extensionAutoUpdate.schedule` is empty on every CheCluster unless the user types it.

The Go code masks this at runtime (`schedule := constants.DefaultExtensionAutoUpdateSchedule` falls back when `Schedule` is nil), which is why tests pass — but the CRD-level default is still declared and dead, and the default schedule is invisible in the CR.

**2. `code-style.md` — errors returned without context** (`pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go`)

> "Wrap errors with context using `fmt.Errorf` at every call site instead of passing them through directly … Error messages must start with `failed to` (lowercase). When the object has a namespace, include it as `namespace/name`."

Three call sites return bare:

```go
if err := controllerutil.SetControllerReference(ctx.CheCluster, cronJob, ctx.ClusterAPI.Scheme); err != nil {
    return err                                                    // L46
}

return ctx.ClusterAPI.ClientWrapper.Sync(...)                     // L50 — unwrapped

return ctx.ClusterAPI.ClientWrapper.DeleteByKeyIgnoreNotFound(...) // L173 — unwrapped
```

This matters here because `syncExtensionUpdateCronJob` is deliberately non-blocking (`openvsx_server.go:105` only logs the error), so an unwrapped error surfaces in the operator log with no indication of which object failed.

**3. `k8s-client.md` — deprecated and disallowed client calls** (`openvsx_server_extension_update_test.go`)

> "**Do NOT use** the legacy functions from `pkg/deploy/sync.go` (`deploy.Sync`, `deploy.Get*`, …) — that file is deprecated."
> "When reading objects, use `ClientWrapper.GetIgnoreNotFound` — do **not** use `Client.Get`."

Two violations:
- `deploy.GetNamespacedObject(ctx, constants.OpenVSXServerComponentName, deployment)` in `reconcileWithReadyDeployment` — `GetNamespacedObject` lives in the deprecated `pkg/deploy/sync.go:73`.
- `ctx.ClusterAPI.Client.Get(context.TODO(), cronJobKey, cronJob)` in `TestExtensionAutoUpdateCronJobCleanedUpWhenRegistryDisabled`.

**4. `code-style.md` — lines over 120 characters**

Six added lines in `openvsx_server_extension_update_test.go` run 123–174 characters (the `test.IsObjectExists(...)` assertions at 159/173 chars, plus three `assert` calls). The rule asks for one argument per line when a call exceeds the limit.

(`openvsx_server_test.go:88` is also 174 chars, but it matches the surrounding assertions in that file, so #2202 leaves it alone.)

**5. `redhat-compliance.md` — copyright year**

> "When editing an existing file, update the copyright year range end to the current year."

`pkg/common/diffs/diffs.go` was modified but still reads `Copyright (c) 2019-2025`. Every other touched Go file is already at `2019-2026`.

---

### Rules that pass ✅

- **`api-types.md`** — v2 only, `zz_generated.deepcopy.go` is properly generated (not hand-edited), CRDs regenerated, no v1 changes needed.
- **`reconcilers.md`** — `OpenVSXServerReconciler` keeps its `Reconcile`/`Finalize` contract; the CronJob sync is placed before the server-readiness gate so disabling takes effect even when the server is unhealthy.
- **`resource-naming.md`** — `openvsx-server-extension-update` contains no `che` / `eclipse-che` / `devspaces` hardcoding; the flavor-dependent bits come from `defaults.GetOpenVSXImage()` and `deploy.GetLabels()`.
- **`k8s-client.md`** (production code) — uses `ClientWrapper.Sync` with `&k8sclient.SyncOptions{DiffOpts: diffs.CronJob}`, adds the missing `diffs.CronJob` option, sets the owner reference for the same-namespace object, and uses `DeleteByKeyIgnoreNotFound` for the disable path. Textbook.
- **`testing.md`** — `test.NewCtxBuilder()`, `test.EnsureReconcile(...)`, co-located `_test.go`, good coverage of enable/disable/cleanup/defaults/optional-env paths.
- **`redhat-compliance.md`** (rest) — EPL-2.0 headers on both new files, no secrets (the PAT is a `SecretKeyRef`), and every commit carries the `Assisted-by: Claude Opus 4.6` trailer.
- **`build-and-test.md`** — verified locally: `go build -mod=vendor ./...` OK, `go vet` clean, `gofmt -l` clean, `go test ./pkg/deploy/openvsx/...` passes. A full `go test ./...` shows 2 failing packages (`controllers/che`, `pkg/deploy/gateway`), but they fail identically on the base branch — environment-related (`NO_PROXY` picks up the sandbox cluster IP), not caused by this PR.

### One rule I can't fix from a follow-up PR

**`pull-requests.md`** — "Check all applicable items in the *Common Test Scenarios* checklist" and "Check all applicable items in the *PR Checklist*". Both checklists in this PR's description are entirely unchecked. The screenshots show the feature was clearly exercised on a cluster, so at least *Deploy Eclipse Che*, *Code produced is complete*, *Code builds without errors*, and *Tests are covering the bugfix* look applicable. Worth ticking the boxes that apply — this is the only rule deviation left for you.

### Also worth a look (not a rule violation)

`bundle/next/.../che-operator.clusterserviceversion.yaml` picked up an incidental version bump `7.122.0-1062.next` → `7.123.0-1066.next` plus a `createdAt` change, side effects of running `make update-dev-resources` locally. Intentional or not, it's unrelated to the feature.

---

api/v2/checluster_types.go
```diff
@@ -534,6 +538,31 @@ type OpenVSXDatabase struct {
...
 
+// Configuration for periodic auto-update of extensions from the upstream open-vsx.org registry.
+// When enabled, a CronJob periodically checks for newer versions of published extensions
+// and re-publishes them to the internal registry.
+// +k8s:openapi-gen=true
+type ExtensionAutoUpdate struct {
+	// Enables the extension auto-update CronJob.
+	// +optional
+	// +kubebuilder:default:=false
+	Enable bool `json:"enable,omitempty"`
```
tolusha: Let's rename it to `Enabled`. There is a mess with enable/enabled here.

pkg/common/constants/constants.go
```diff
@@ -50,8 +50,10 @@ const (
 	OpenVSXServerServicePort   = int32(8080)
 
 	// OpenVSXRegistry Server extensions
-	OpenVSXServerExtensionPublishJobName = "openvsx-server-publisher"
-	OpenVSXServerExtensionsConfigMapName = "openvsx-server-extensions"
+	OpenVSXServerExtensionPublishJobName    = "openvsx-server-publisher"
+	OpenVSXServerExtensionsConfigMapName    = "openvsx-server-extensions"
+	OpenVSXServerExtensionUpdateCronJobName = "openvsx-server-extension-update"
```
tolusha:
```suggestion
	OpenVSXServerExtensionUpdateCronJobName = "openvsx-server-extensions-update"
```

api/v2/checluster_types.go
svor: renamed

pkg/common/constants/constants.go
svor: renamed

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func deleteExtensionUpdateCronJob(ctx *chetypes.DeployContext) error {
+	cronJobKey := types.NamespacedName{
+		Name:      constants.OpenVSXServerExtensionUpdateCronJobName,
+		Namespace: ctx.CheCluster.Namespace,
+	}
+
+	if err := ctx.ClusterAPI.ClientWrapper.DeleteByKeyIgnoreNotFound(
+		context.TODO(),
```
tolusha: if you have `chetypes.DeployContext`  pls use ctx.Context for kubernetes operations instead of context.TODO() all over the changes.

pkg/common/diffs/diffs.go
```diff
@@ -89,6 +91,17 @@ var ServiceMonitor = cmp.Options{
...
+	cmpopts.IgnoreFields(batchv1.CronJob{}, "TypeMeta", "ObjectMeta", "Status"),
+	cmpopts.IgnoreFields(corev1.Container{}, "TerminationMessagePath", "TerminationMessagePolicy"),
+	cmpopts.IgnoreFields(corev1.PodSpec{}, "DNSPolicy", "SchedulerName", "DeprecatedServiceAccount"),
+	cmpopts.IgnoreFields(corev1.ConfigMapVolumeSource{}, "DefaultMode"),
+	cmpopts.IgnoreFields(corev1.SecretVolumeSource{}, "DefaultMode"),
+	cmp.Comparer(func(x, y resource.Quantity) bool {
+		return x.Cmp(y) == 0
+	}),
```
tolusha: pls remove comparer for resource.Quantity

pkg/common/diffs/diffs.go
```diff
@@ -89,6 +91,17 @@ var ServiceMonitor = cmp.Options{
 	cmpopts.IgnoreFields(monitoringv1.ServiceMonitor{}, "TypeMeta", "ObjectMeta"),
 }
 
+var CronJob = cmp.Options{
```
tolusha, line 93: I can see that the CronJob is being constantly reconciled.

```
Difference:
  &v1.CronJob{
  	... // 2 ignored fields
  	Spec: v1.CronJobSpec{
  		... // 2 identical fields
  		StartingDeadlineSeconds:    nil,
  		ConcurrencyPolicy:          "Forbid",
- 		Suspend:                    &false,
+ 		Suspend:                    nil,
  		JobTemplate:                {ObjectMeta: {Labels: {"app.kubernetes.io/component": "openvsx-server-extensions-update", "app.kubernetes.io/instance": "che", "app.kubernetes.io/managed-by": "che-operator", "app.kubernetes.io/name": "che", ...}}, Spec: {Parallelism: &1, Completions: &1, ActiveDeadlineSeconds: &1800, BackoffLimit: &3, ...}},
  		SuccessfulJobsHistoryLimit: &1,
  		FailedJobsHistoryLimit:     &3,
  	},
  	... // 1 ignored field
  }
```

api/v2/checluster_types.go
```diff
@@ -548,6 +552,31 @@ type OpenVSXDatabase struct {
 	Storage *PVC `json:"pvc,omitempty"`
 }
 
+// Configuration for periodic auto-update of extensions from the upstream open-vsx.org registry.
+// When enabled, a CronJob periodically checks for newer versions of published extensions
+// and re-publishes them to the internal registry.
+// +k8s:openapi-gen=true
+type ExtensionAutoUpdate struct {
```
tolusha, line 559:
```go
// Configuration for periodic auto-update of extensions from the upstream open-vsx.org registry.
// When enabled, a CronJob periodically checks for newer versions of published extensions
// and re-publishes them to the internal registry.
// +k8s:openapi-gen=true
type ExtensionAutoUpdate struct {
      // Enables periodic automatic updates of extensions published to the internal OpenVSX registry.
      // When set to `false`, the auto-update CronJob is deleted, including its job history.
      // +optional
      // +kubebuilder:default:=false
      Enable *bool `json:"enable,omitempty"`
      // Cron schedule, in Unix cron format, for the auto-update job.
      // The schedule is interpreted in the kube-controller-manager time zone, which is UTC on most clusters.
      // +optional
      // +kubebuilder:default:="0 0 * * 0"
      Schedule *string `json:"schedule,omitempty"`
      // Version of the VS Code engine used to select which extension versions to publish.
      // For each extension, the newest version whose `engines.vscode` constraint is satisfied
      // by this value is published, for example `1.99.0`.
      // When omitted, the latest non-pre-release version of each extension is used
      // regardless of engine compatibility.
      // +optional
      VSCodeEngineVersion *string `json:"vsCodeEngineVersion,omitempty"`
      // List of extensions excluded from auto-update.
      // Each entry is an extension ID in `<namespace>/<name>` format, also known as `publisher/name`,
      // without a version, for example `redhat/java`.
      // Excluded extensions keep their currently published version.
      // +optional
      ExcludedExtensions []string `json:"excludedExtensions,omitempty"`
}
```

api/v2/checluster_types.go
```diff
@@ -1329,6 +1358,12 @@ func (c *CheCluster) IsInternalOpenVSXRegistryEnabled() bool {
 	return c.Spec.Components.OpenVSXRegistry.Enable
 }
 
+func (c *CheCluster) IsExtensionAutoUpdateEnabled() bool {
```
tolusha, line 1377: Should be enabled by default if openvsx registry is enabled?

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
```diff
@@ -0,0 +1,197 @@
...
+									ImagePullPolicy: corev1.PullPolicy(imagePullPolicy),
+									VolumeMounts: []corev1.VolumeMount{
+										{
+											Name:      "extensions",
+											MountPath: "/home/openvsx/extensions",
+											ReadOnly:  true,
+										},
+									},
+									Env:     env,
+									Command: []string{"/home/openvsx/update-extensions.sh", "/home/openvsx/extensions/extensions.list"},
```
tolusha, line 154: Sorry, I can't find how `/home/openvsx/extensions/extensions.list` is mounted

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
```diff
@@ -0,0 +1,197 @@
...
+		{
+			Name:  "OVSX_REGISTRY_URL",
+			Value: openvsx.GetOpenVSXServerServiceURL(ctx),
+		},
+		{
+			Name:  "OVSX_FORWARDED_HOST",
+			Value: ctx.CheHost,
+		},
+		{
+			Name:  "OVSX_FORWARDED_PROTO",
```
tolusha: How it will be used?

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
svor: I believe I fixed it in all places

pkg/common/diffs/diffs.go
svor: removed

pkg/common/diffs/diffs.go
svor: Added Suspend: ptr.To(false) to match the Kubernetes server-side default, which stops the constant reconciliation diff.

api/v2/checluster_types.go
svor: applied

api/v2/checluster_types.go
svor: let's have it disabled by default

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
svor: The ConfigMap `openvsx-server-extensions` (which has a key `extensions.list`) is mounted as a volume at `/home/openvsx/extensions`. The ConfigMap key `extensions.list` becomes the file `/home/openvsx/extensions/extensions.list` inside the container. This is the same pattern used by the publish Job in `openvsx_server_extensions.go`

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
svor: it is used in the `publish-extensions.sh` and `update-extensions.sh` scripts to set the X-Forwarded-Proto header on requests to the OpenVSX server, but it's part of openvsx image and should not be a part of che-operator - I removed this ENV. 

This is needed because the OpenVSX server caches extension download URLs based on the request origin - without the forwarded header, it would cache internal http:// URLs, which are not reachable from `https://cheHost/openvsx` that access the registry through the external HTTPS gateway.

pkg/deploy/openvsx/openvsx-server/openvsx_server.go
```diff
@@ -102,6 +100,10 @@ func (r *OpenVSXServerReconciler) Reconcile(ctx *chetypes.DeployContext) (reconc
 		return reconcile.Result{}, false, fmt.Errorf("failed to sync Extensions Config: %w", err)
 	}
 
+	if err = r.syncExtensionUpdateCronJob(ctx); err != nil {
+		logger.Error(err, "Failed to sync extension update CronJob, continuing reconciliation")
```
tolusha: Pls, return error

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
```diff
@@ -0,0 +1,193 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func (r *OpenVSXServerReconciler) syncExtensionUpdateCronJob(ctx *chetypes.DeployContext) error {
+	if !ctx.CheCluster.IsExtensionAutoUpdateEnabled() {
+		return deleteExtensionUpdateCronJob(ctx)
```
tolusha, line 39: I think this never be called.

pkg/deploy/openvsx/openvsx-server/openvsx_server.go
```diff
@@ -188,6 +190,11 @@ func deleteResources(ctx *chetypes.DeployContext) {
 		logger.Error(err, "Failed to delete ConfigMap", "Name", constants.OpenVSXServerExtensionsConfigMapName)
 	}
 
+	err = deleteExtensionUpdateCronJob(ctx)
```
tolusha: Let's move deletion deployments,jobs at the beginning and then everything else

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
```diff
@@ -0,0 +1,193 @@
...
+		&k8sclient.SyncOptions{DiffOpts: diffs.CronJob},
+	); err != nil {
+		return fmt.Errorf("failed to sync CronJob %s/%s: %w", cronJob.Namespace, cronJob.Name, err)
+	}
+
+	return nil
+}
+
+func (r *OpenVSXServerReconciler) getExtensionUpdateCronJobSpec(ctx *chetypes.DeployContext) (*batchv1.CronJob, error) {
+	autoUpdate := ctx.CheCluster.Spec.Components.OpenVSXRegistry.ExtensionAutoUpdate
```
tolusha, line 63: Add nil check pls

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
svor: The `!IsExtensionAutoUpdateEnabled()` check here covers the scenario where an admin disables auto-update (enable: false) while keeping the registry running - the CronJob needs to be cleaned up.

pkg/deploy/openvsx/openvsx-server/openvsx_server_extension_update.go
svor: done

pkg/deploy/openvsx/openvsx-server/openvsx_server.go
svor: done

pkg/deploy/openvsx/openvsx-server/openvsx_server.go
svor: done

tolusha:
Pls add requests/limits resources for jobs.

tolusha (approved):

bundle/next/eclipse-che/manifests/org.eclipse.che_checlusters.yaml
```diff
@@ -14763,6 +14766,43 @@ spec:
...
+                            enable:
+                              default: false
+                              description: |-
+                                Enables periodic automatic updates of extensions published to the internal OpenVSX registry.
+                                When set to `false`, the auto-update CronJob is deleted, including its job history.
+                              type: boolean
+                            excludedExtensions:
+                              description: |-
+                                List of extensions excluded from auto-update.
+                                Each entry is an extension ID in `<namespace>/<name>` format, also known as `publisher/name`,
```
ibuziuk: Would be great to exclude by ID e.g. `redhat.java`

https://github.com/che-incubator/quarkus-api-example/blob/main/.vscode/extensions.json

ibuziuk (approved):

bundle/next/eclipse-che/manifests/org.eclipse.che_checlusters.yaml
svor: fixed and updated the description

PR #2201, opened 2026-09-18
feat: add CheCluster field to toggle oauth-proxy image resolution from ImageStream

tolusha:
### What does this PR do?

Adds a new CheCluster field `spec.networking.auth.gateway.oAuthProxy.resolveImageFromImageStream` that controls whether the `oauth-proxy` image is resolved from the cluster's own `openshift/oauth-proxy` ImageStream instead of using the image shipped with the Operator.

- The image is referenced via the internal registry, so it works in both connected and disconnected environments.
- The field is specific to OpenShift with built-in OAuth.
- Enabled by default. When the image cannot be resolved, the Operator falls back to the default one.

This makes the behaviour introduced in #2190 configurable, so users can opt out and pin the Operator-shipped image.

Also replaces the `defaults.GetCheFlavor() == "che"` string comparison in `getOauthProxyContainerSpec` with the existing `CheCluster.IsCheFlavor()` helper.

N/A — API/behaviour change with no UI impact.

Follow-up to #2190.

### How to test this PR?

1. Deploy the operator:

#### OpenShift
```bash

OPERATOR_IMAGE=<...>
oc wait --for=condition=Ready pod -l app.kubernetes.io/component=che-operator -n eclipse-che --timeout=60s
```

2. Verify the default behaviour — the gateway `oauth-proxy` container uses the image from the cluster ImageStream:

```bash
oc get deployment che-gateway -n eclipse-che -o jsonpath='{.spec.template.spec.containers[?(@.name=="oauth-proxy")].image}'
# expected: image-registry.openshift-image-registry.svc:5000/openshift/oauth-proxy@sha256:...
```

3. Disable the resolution and verify the Operator-shipped image is used instead:

```bash
oc patch checluster eclipse-che -n eclipse-che --type=merge \
  -p '{"spec":{"networking":{"auth":{"gateway":{"oAuthProxy":{"resolveImageFromImageStream":false}}}}}}'

oc rollout status deployment/che-gateway -n eclipse-che
oc get deployment che-gateway -n eclipse-che -o jsonpath='{.spec.template.spec.containers[?(@.name=="oauth-proxy")].image}'
# expected: the oauth-proxy image shipped with the Operator
```

4. Re-enable it (or remove the field) and verify the image goes back to the ImageStream-resolved one.

#### on Minikube

```bash
```

#### Common Test Scenarios
- [x] Deploy Eclipse Che
- [x] Start an empty workspace
- [x] Open terminal and build/run an image
- [x] Stop a workspace

### PR Checklist

### Reviewers


tolusha:

**Available commands**:

tolusha:


Task completed.

api/v2/checluster_types.go
```diff
@@ -854,6 +854,14 @@ type OAuthProxy struct {
 	// +kubebuilder:default:=86400
 	// +kubebuilder:validation:Minimum:=0
 	CookieExpireSeconds *int32 `json:"cookieExpireSeconds,omitempty"`
+	// Allows resolving the oauth-proxy image from the cluster's own `openshift/oauth-proxy` ImageStream
+	// instead of using the image shipped with the Operator. The image is referenced by the internal
+	// registry, so it works in both connected and disconnected environments.
+	// This field is specific to OpenShift with built-in OAuth.
+	// When the image cannot be resolved, the Operator falls back to the default one.
```
tolusha: Two things this description could add, since this text becomes the CRD description users read via `oc explain`.

**Flavor restriction.** The gate at `pkg/deploy/gateway/oauth_proxy.go:242` is `IsCheFlavor() && IsOAuthProxyImageResolvedFromImageStream()`. Downstream ships the same generated CRD, so the field is settable there but does nothing and warns about nothing.

**Pinning.** The fallback is not equivalent to the resolved image:

| Value | Image | Pinning |
|---|---|---|
| `true` | `...svc:5000/openshift/oauth-proxy@sha256:...` | by digest |
| `false` | `quay.io/openshift/origin-oauth-proxy:4.22` (`RELATED_IMAGE_gateway_authentication_sidecar`) | mutable tag |

Opting out swaps a digest-pinned payload image for a floating tag, on the container that terminates authentication. Worth making explicit so admins choose it knowingly.

```suggestion
	// This field is specific to OpenShift with built-in OAuth and upstream Che;
	// on other flavors the setting is ignored.
	// When the image cannot be resolved, the Operator falls back to the image shipped
	// with the Operator, which is referenced by a mutable tag rather than by digest.
```

api/v2/checluster_types.go
```diff
@@ -854,6 +854,14 @@ type OAuthProxy struct {
...
+	// Enabled by default.
+	// +optional
```
tolusha, line 865: The comment says "Enabled by default", but the generated schema has no `default:`. So `oc explain` shows no default and the value is never persisted onto the CR - an admin has no way to see the effective setting.

`CookieExpireSeconds` two lines above uses `+kubebuilder:default:=86400`, and several other booleans in this file do the same.

This is behaviour-preserving: the schema default and `constants.DefaultOAuthProxyResolveImageFromImageStream` agree, and the `ptr.Deref` guard stays useful for CRs created before the schema change.

```suggestion
	// +optional
	// +kubebuilder:default:=true
```

api/v2/checluster_types.go
```diff
@@ -1356,6 +1364,17 @@ func (c *CheCluster) IsNetworkPoliciesEnabled() bool {
 		ptr.Deref(c.Spec.Networking.NetworkPolicy.Enabled, constants.NetworkPolicyEnabled)
 }
 
+func (c *CheCluster) IsOAuthProxyImageResolvedFromImageStream() bool {
```
tolusha: `IsOAuthProxyImageResolvedFromImageStream()` reads as an assertion about what happened, but it reports configuration. The image may still come from the Operator-shipped default when the lookup fails or the flavor is not `che` - in which case the name says the opposite of the truth.

`IsOAuthProxyImageResolutionFromImageStreamEnabled()` would match `IsNetworkPoliciesEnabled()` and `IsAgentSandboxEnabled()` directly above and below it. Would need the call site at `pkg/deploy/gateway/oauth_proxy.go:242` updated too.

Separately - the 10-line early-return form here is correct and worth keeping. The `X != nil && ptr.Deref(...)` one-liner used by both neighbours short-circuits to `false` on nil, but this field must default to `true`. A brief comment would stop a future reader "simplifying" it into a behaviour change.

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -239,7 +239,7 @@ func getOauthProxyContainerSpec(ctx *chetypes.DeployContext) corev1.Container {
 	var args = []string{"--config=/etc/oauth-proxy/oauth-proxy.cfg"}
 	if infrastructure.IsOpenShiftOAuthEnabled() {
 		// Guarded to upstream Che only; downstream manages its own oauth-proxy image.
```
tolusha, line 240: The comment was written for the single-clause version. It still describes the flavor operand only, so a reader skimming it will not learn that a user-facing toggle also gates this line.

```suggestion
		// Guarded to upstream Che only (downstream manages its own oauth-proxy image)
		// and to clusters that have not opted out via
		// spec.networking.auth.gateway.oAuthProxy.resolveImageFromImageStream.
```

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -239,7 +239,7 @@ func getOauthProxyContainerSpec(ctx *chetypes.DeployContext) corev1.Container {
...
-		if defaults.GetCheFlavor() == "che" {
+		if ctx.CheCluster.IsCheFlavor() && ctx.CheCluster.IsOAuthProxyImageResolvedFromImageStream() {
```
tolusha: The fallback on line 246 goes through `defaults.GetGatewayOpenShiftAuthenticationSidecarImage`, which calls `PatchDefaultImageName` (`pkg/common/operator-defaults/defaults.go:320-339`) and rewrites the host and organization from `spec.containerRegistry`. The ImageStream path bypasses that entirely.

So a user who configured `spec.containerRegistry` to point at a private mirror gets that setting silently ignored for the oauth-proxy container, because resolution is on by default and wins. Their explicit registry preference loses to an implicit default, and nothing logs it.

This PR is the cure - setting `resolveImageFromImageStream: false` restores `containerRegistry` behaviour - but nothing tells the user that. Two options:

- Document the interaction in the field's doc comment, so it reaches the CRD description; or
- Treat a configured `spec.containerRegistry` as an implicit `false`, which is almost certainly the intent.

Does the second make sense here? It removes a footgun that only shows up in mirrored and air-gapped installs - the environments #2190 was written to serve.

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -239,7 +239,7 @@ func getOauthProxyContainerSpec(ctx *chetypes.DeployContext) corev1.Container {
...
 			image = resolveOpenShiftOAuthProxyImage(ctx)
 		}
 		if image == "" {
```
tolusha, line 244: `resolveOpenShiftOAuthProxyImage` logs on success (line 223), on lookup failure (197) and on an empty ImageStream (228). When the field is `false` the function is never entered, so nothing is logged and this fallback is silent.

From the Operator log, "an admin turned it off" and "the cluster read failed" are indistinguishable, and both end with the gateway running a different image. Since `deploy.DefaultDeploymentDiffOpts` does not ignore `Container.Image`, both also trigger a gateway rollout - so this is the log line someone will want when an unexplained restart shows up.

Something like this at the top of the `IsOpenShiftOAuthEnabled()` branch would cover it:

```go
if !ctx.CheCluster.IsOAuthProxyImageResolvedFromImageStream() {
    logrus.Info("oauth-proxy image resolution from ImageStream is disabled, using the default image")
}
```

tolusha (commented):
## ok-pr-review

Ran `summary` -> `review` -> `deep-review` -> `impact`.

**Verdict: Comment.** The change is small, correct and backward compatible. Six inline comments plus the notes below - nothing that blocks a running system.

### What works well

- **The default preserves #2190 behaviour exactly.** Choosing `true` rather than `false` means no cluster changes what it runs by upgrading. For a toggle retrofitted onto shipped behaviour that is the right call, and it is also the better-pinned image.
- **The `IsCheFlavor()` swap is exactly equivalent** - `IsCheFlavor()` is `defaults.GetCheFlavor() == constants.CheFlavor` and `constants.CheFlavor = "che"` - so the readability win carries no behavioural risk.
- **Nil-guard is correct.** `Networking`, `Auth` and `Gateway` are value types; `OAuthProxy` is the only pointer and the only thing checked. Matches the existing guard in `cookieExpireAsString`.
- **Generated manifests are complete and consistent.** All seven `next`-track CRD copies carry the identical field, and `bundle/stable` / `helmcharts/stable` are correctly untouched. No CI job verifies this, so it matters that it was done properly.
- **Single decision point.** One field, one helper, one call site, no parallel structures to keep in sync.
- **The `v1 "k8s.io/api/core/v1"` import rewrite in `zz_generated.deepcopy.go` is expected, not drift** - `pr-check.yml` runs `make fmt`, which runs `goimports` over every non-vendor Go file and fails on any diff.

### Compatibility

Clean. The CRD change is purely additive, no version bump or conversion webhook is needed, the Go API change is additive, and downgrade degrades to the pre-#2190 behaviour. No concerns here.

### General findings

**1. Test coverage.** The checklist leaves "Tests are covering the bugfix" unchecked, and the deeper gap is that `getOauthProxyContainerSpec` has never been tested at all - `oauth_proxy_test.go` covers the resolver in isolation, but the function that actually decides (flavor gate, toggle, fallback assignment) has no test before or after this PR. Nothing verifies the fallback the field documentation promises. Four cases worth pinning:

- `resolveImageFromImageStream: false` with a healthy ImageStream present -> Operator-shipped image wins. This is the entire point of the PR.
- `OAuthProxy` nil -> resolution enabled. Guards the default-`true` contract, which is one character away from flipping.
- `ResolveImageFromImageStream` nil with a non-nil `OAuthProxy` (a CR that only sets `cookieExpireSeconds`) -> resolution enabled. The realistic upgrade path, and a different branch from the nil-struct case.
- `IsCheFlavor() == false` with the toggle `true` -> Operator image. Pins the precedence between the operands.

The first three need no fixture; the fourth reuses `test.NewCtxBuilder().WithObjects(imageStream)` from `oauth_proxy_test.go:158` almost verbatim. Roughly 40 lines for the set.

Worth saying that the existing resolver tests are rigorous - `TestResolveOpenShiftOAuthProxyImage_NoInternalRegistry` deliberately plants a `dockerImageReference` and asserts `""` is still returned, pinning the disconnected-safety invariant from #2190. That is good work, which is why the gap next to it stands out.

**2. `context.TODO()` at `oauth_proxy.go:193`.** `DeployContext` already carries a real context at `pkg/common/chetypes/types.go:33`, and newer code in this repo uses it (`agent_sandbox.go:98`, `networkpolicies.go:352`). As written, reconcile cancellation does not propagate into what is an uncached GET issued on every gateway reconcile. Fair to note `context.TODO()` appears 118 times across `pkg/`, so it is the legacy majority pattern - but the right context is already on the struct being passed in. Inherited from #2190, two-word fix.

**3. Transient failure causes two gateway rollouts.** `SyncDeploymentSpecToCluster` compares with `deploy.DefaultDeploymentDiffOpts`, which ignores probes and replicas but not `Container.Image`. A transient failure of the ImageStream GET returns `""`, the fallback substitutes the Operator image, the diff sees a change and the gateway rolls out; the next successful reconcile rolls it back. Two restarts of the auth gateway from one API blip. Inherited from #2190 - flagged because inline comment 6 is what makes it diagnosable.

**4. No tag selection in the resolver loop** (`oauth_proxy.go:205-226`). The loop takes the first tag that has any items, regardless of which tag it is. The payload stream carries one tag today (`v4.4`, per the test fixture), so this works - but if a payload ever ships two, the chosen image depends on slice ordering rather than intent, and the failure mode is a working-but-wrong image rather than an error. Naming the expected tag would make the intent legible. Inherited, not introduced here.

**5. Unused `list` verb** in the imagestreams grant (`config/rbac/cluster_role.yaml:164-171`). The code only ever performs a single `Get` on `openshift/oauth-proxy`. Pre-existing and out of scope, but this PR is the natural place to notice it.

**6. User documentation.** "Relevant user documentation updated" is unchecked for a new user-facing CRD field - presumably a che-docs follow-up.

### Two smaller notes

- The per-reconcile uncached GET means the default-on path costs one API round-trip per gateway reconcile, forever. This PR is the first mechanism to switch that off, which is a real operational benefit worth a line in the PR description. A follow-up could memoise the digest, since it only changes on cluster upgrade.
- The seven-line realignment in `constants.go` is unavoidable gofmt alignment, not a hidden change. Worth a note in the PR body so reviewers do not go looking.

<sub>Generated by ok-pr-review (summary, review, deep-review, impact).</sub>

tolusha:

Task completed.

api/v2/checluster_types.go
```diff
@@ -854,6 +854,16 @@ type OAuthProxy struct {
...
 	// +kubebuilder:validation:Minimum:=0
 	CookieExpireSeconds *int32 `json:"cookieExpireSeconds,omitempty"`
+	// Allows resolving the oauth-proxy image from the cluster's own `openshift/oauth-proxy` ImageStream
+	// instead of using the image shipped with the Operator. The image is referenced by the internal
+	// registry, so it works in both connected and disconnected environments.
+	// This field is specific to OpenShift with built-in OAuth and upstream Che,
+	// on other flavors the setting is ignored.
+	// When the image cannot be resolved, the Operator falls back to the image shipped
+	// with the Operator, which is referenced by a mutable tag rather than by digest.
+	// Enabled by default.
```
tolusha, line 865: There are three sources for the oauth-proxy image, and the highest-precedence one is not mentioned here.

`spec.networking.auth.gateway.deployment.containers[name=oauth-proxy].image` overrides both branches of this field. `getContainersSpec` builds the container at `pkg/deploy/gateway/gateway.go:457`, then `OverrideDeployment` runs at `gateway.go:470` and assigns `container.Image` at `pkg/deploy/deployment.go:245`.

Since this text becomes the CRD description, one sentence here saves an admin who sets both from concluding the toggle is broken.

```suggestion
	// An explicit image in `gateway.deployment.containers` takes precedence over this setting.
	// Enabled by default.
```

api/v2/checluster_types.go
```diff
@@ -854,6 +854,16 @@ type OAuthProxy struct {
...
+	// +optional
+	ResolveImageFromImageStream *bool `json:"resolveImageFromImageStream,omitempty"`
```
tolusha, line 867: CRD properties are effectively permanent once released, so the shape chosen here outlives the decision behind it.

The question this field answers is already three-valued in practice: the cluster payload, the Operator-shipped image, or an explicit image via `gateway.deployment`. A boolean covers the first two. The moment someone wants "resolve from a different ImageStream", "fail closed instead of falling back", or "prefer the payload only when it is newer", we cannot extend the boolean - we have to add a second field beside it and define how the two interact.

`imageSource: ImageStream | Operator` with `+kubebuilder:default:=ImageStream` carries identical semantics today and leaves room to add a value later instead of a field. Does it make sense to decide this before it ships?

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -239,7 +238,7 @@ func getOauthProxyContainerSpec(ctx *chetypes.DeployContext) corev1.Container {
 	var args = []string{"--config=/etc/oauth-proxy/oauth-proxy.cfg"}
 	if infrastructure.IsOpenShiftOAuthEnabled() {
 		// Guarded to upstream Che only; downstream manages its own oauth-proxy image.
-		if defaults.GetCheFlavor() == "che" {
+		if ctx.CheCluster.IsCheFlavor() && ctx.CheCluster.IsOAuthProxyImageResolutionFromImageStreamEnabled() {
```
tolusha, line 241: `getOauthProxyContainerSpec` is called only from `gateway.go:601`, and nothing in the package exercises it - so this line, which is the whole user-visible behaviour of the PR, is not covered by CI. The three tests from #2190 call `resolveOpenShiftOAuthProxyImage` directly and bypass the gate.

Two cases would close it, using the fixture already in `oauth_proxy_test.go`:

- toggle on + seeded ImageStream -> container image is the digest reference
- toggle off + the same ImageStream -> container image is `RELATED_IMAGE_gateway_authentication_sidecar`

Plus a 4-case table test on `IsOAuthProxyImageResolutionFromImageStreamEnabled` (nil `OAuthProxy`, nil pointer, `true`, `false`) - the first two branches are the ones a future "simplify this helper" refactor would silently break.

One caveat for whoever writes them: `infrastructure.InitializeForTesting` mutates process-global state and is already called eight times in that file on top of `init_test.go:23`, so please set the mode explicitly rather than inheriting it from the preceding test.

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -190,7 +189,7 @@ func resolveOpenShiftOAuthProxyImage(ctx *chetypes.DeployContext) string {
 	})
 
 	if err := ctx.ClusterAPI.NonCachingClient.Get(
-		context.TODO(),
+		ctx.Context,
```
tolusha, line 192: Good change, and it is safe: all three non-test `chetypes.DeployContext` constructors populate `Context` (`controllers/che/checluster_controller.go:245`, `controllers/usernamespace/usernamespace_controller.go:256`, `pkg/common/test/deploy_context.go:68`), so there is no nil-context path here.

Worth knowing what it means downstream: a cancelled reconcile context now makes this `Get` fail, which falls through to the Operator-shipped image. `Container.Image` is not in `DefaultDeploymentDiffOpts` (`pkg/deploy/deployment.go:43-54`), so that swap rolls the gateway, and the next successful reconcile rolls it back. Rare, but it is one more path into the flap.

pkg/common/constants/constants.go
```diff
@@ -20,13 +20,14 @@ const (
...
-	DefaultKubeRbacProxyLogLevel         = int32(0)
-	DefaultOAuthProxyCookieExpireSeconds = int32(86400)
+	DefaultGatewayMemoryLimit                    = "256Mi"
+	DefaultGatewayMemoryRequest                  = "64Mi"
+	DefaultGatewayCpuLimit                       = "500m"
+	DefaultGatewayCpuRequest                     = "50m"
+	DefaultTraefikLogLevel                       = "INFO"
+	DefaultKubeRbacProxyLogLevel                 = int32(0)
+	DefaultOAuthProxyCookieExpireSeconds         = int32(86400)
+	DefaultOAuthProxyResolveImageFromImageStream = true
```
tolusha, line 30: This constant is the entire behavioural payload of the PR. The reasoning is in the PR description, which will be harder to find in a year than the line itself.

```suggestion
	// Resolve from the cluster release payload by default: that image is digest-pinned
	// and matches the cluster's own architecture.
	DefaultOAuthProxyResolveImageFromImageStream = true
```

tolusha (commented):
## Automated review - summary, code review, deep review, impact

**Verdict: comment.** No blocking defects. The gate, the helper, and the `IsCheFlavor()` refactor are correct (`constants.CheFlavor` is `"che"`, so the swap is behaviour-identical), the CRD change is cleanly additive, and RBAC is unchanged and already minimal - `get`/`list` on `image.openshift.io/imagestreams` at `config/rbac/cluster_role.yaml:164-170` is exactly what the code exercises. The `list` verb narrowed in `473f0975` and restored in `eec2f534` nets to zero against `main`.

Five inline comments below, plus replies on the three threads from the 2026-09-18 pass that are still open. The naming fix and the doc-comment rewrite from that pass both landed cleanly.

### CI - all three failures are unrelated infra

- `helm-on-minikube`: `failed calling webhook "webhook.cert-manager.io": ... connection refused` while installing Dex.
- `ci/prow/v19-devworkspace-happy-path` and `ci/prow/v19-upgrade-stable-to-next`: both died in the shared `root-amd64` build-root step on a dead `vault.centos.org/centos/7` mirror, before any Che code ran. `/retest` should clear them.

Worth noting the consequence: neither OpenShift e2e job actually exercised this change, so OpenShift verification currently rests on the manual steps in the description.

`unit-tests`, `resources-validation`, `bundle-version-validation`, `source-code-validation` and CodeQL all pass, so the tree compiles and all seven `next`-channel CRD copies plus `zz_generated.deepcopy.go` are in sync, with `stable` correctly untouched.

### The pinning story is inverted

| Setting | Image | Pinning |
|---|---|---|
| `true` (default) | `...svc:5000/openshift/oauth-proxy@sha256:...` | digest |
| `false` | `quay.io/openshift/origin-oauth-proxy:4.22` (`config/manager/manager.yaml:76-77`) | mutable tag |

Opting out - the thing an admin does to take control - loosens pinning on the container that terminates authentication. With `ImagePullPolicy: IfNotPresent`, nodes that already cached `4.22` will not observe an upstream retag, so a fleet can run two builds under one tag. The doc comment is already honest about this, which is good; it is mainly a release-note point.

Related: with the default on, the trust anchor for the auth sidecar is cluster state rather than the Operator bundle - whoever can write `imagestreams` in the `openshift` namespace picks the digest. That is cluster-admin plus anyone with `edit` there, the same set that could already replace the gateway Deployment, so the marginal exposure is small. It arrived in #2190 without a switch though, and this PR is where it becomes a deliberate default. A second approver signing off on that explicitly seems worthwhile.

### Two things for release notes

- **A downgrade silently prunes the opt-out.** If an admin sets `resolveImageFromImageStream: false` and later rolls the Operator back to a bundle whose CRD lacks the property, structural-schema pruning drops it and the cluster reverts to ImageStream resolution - exactly what they opted out of.
- **The gateway tracks the payload eventually, not promptly.** No watch on ImageStreams and no `SyncPeriod`, so a republished `openshift/oauth-proxy` is picked up within the default 10h resync. Adding a watch would be worse - it would couple Che reconciliation to an object in a namespace it does not own - so documenting the window looks like the right call.

### One behaviour worth keeping in mind

Because `Container.Image` participates in the deployment diff, a failed ImageStream read is not inert: it swaps the image and rolls the gateway, and the next successful reconcile rolls it back. A cluster with an intermittently unavailable API path can flap the gateway between two images, each flap dropping in-flight authentication. The end state converges; the path there is not free. Distinguishing "read failed" from "no image in the stream" at `oauth_proxy.go:196` / `227`, and keeping the previously resolved image on a read error, would take the rollout out of the transient-failure path.

<sub>Generated by ok-pr-review (summary + review + deep-review + impact).</sub>

pkg/deploy/gateway/oauth_proxy.go
tolusha: Still open on `eec2f534`.

One addition since this comment: an admin hitting it today is not stuck. `spec.networking.auth.gateway.deployment.containers[name=oauth-proxy].image` overrides the resolved image, because `OverrideDeployment` (`pkg/deploy/gateway/gateway.go:470`) runs after `getContainersSpec` (`gateway.go:457`) and assigns `container.Image` at `pkg/deploy/deployment.go:245`. That is a workaround rather than a fix - it pins a literal reference that then stops tracking Operator upgrades - but it means the footgun already has an exit.

api/v2/checluster_types.go
tolusha: Still open on `eec2f534` - the generated schema on the current head still has no `default:` for this property.

Worth adding: the Go constant and the absent CRD default are not mechanically linked. They agree today, and if a future edit makes them disagree the schema simply stays silent and `constants.DefaultOAuthProxyResolveImageFromImageStream` wins. A wrong CRD default would be invisible rather than loud.

pkg/deploy/gateway/oauth_proxy.go
tolusha: Still open on `eec2f534`.

There is now a third cause landing in the same indistinguishable bucket. The controller does not watch ImageStreams (`controllers/che/checluster_controller.go:177-203`) and `cmd/main.go:247` sets no `SyncPeriod`, so after a cluster upgrade republishes `openshift/oauth-proxy` the gateway keeps the old digest until the default 10h informer resync. "Admin turned it off", "the read failed" and "we have not resynced yet" all look identical from the Operator log, and the first two also roll the gateway.

pkg/deploy/gateway/oauth_proxy.go
tolusha: Still open on `eec2f534` - line 240 still describes only the flavor operand.

ibuziuk (commented):
@tolusha am I correct that we are changing the default behaviour with the `resolveImageFromImageStream` e.g. now OAuth Proxy ImageStream will be used by default not the one that we bundle as part of the operator?

tolusha:
Correct, for upstream only.

PR #2204, opened 2026-09-22
chore: Use `k8s_client.go` instead of `sync.go`

tolusha:
### What does this PR do?

Completes the migration from the legacy `pkg/deploy/sync.go` helpers to the `K8sClient` wrapper (`pkg/common/k8s-client/`), and removes `pkg/deploy/sync.go` and `pkg/deploy/sync_test.go` entirely.

All call sites across reconcilers, controllers and tests now go through `DeployContext.ClusterAPI.ClientWrapper` / `NonCachingClientWrapper`:

| Removed from `sync.go` | Replacement on `K8sClientWrapper` |
| --- | --- |
| `Sync`, `SyncForClient` | `Sync(ctx, obj, &k8sclient.SyncOptions{DiffOpts: diffs.<Type>})` |
| `GetNamespacedObject`, `GetClusterObject` | `GetIgnoreNotFound(ctx, key, obj)` |
| `CreateIgnoreIfExists` | `CreateIfNotExists(ctx, obj)` |
| `Delete`, `DeleteNamespacedObject`, `DeleteByKeyWithClient` | `DeleteByKeyIgnoreNotFound(ctx, key, obj)` |
| `DeleteIgnoreIfNotFound` | `DeleteIgnoreNotFound(ctx, obj)` |
| `GetObjectType` | `k8sclient.GetObjectType` |

Notable consequences of the switch:

- Owner references are no longer set implicitly inside `Sync`. Call sites in the CheCluster namespace now set them explicitly via `controllerutil.SetControllerReference` before syncing, and cross-namespace objects deliberately get none.
- Diff options are passed explicitly through `SyncOptions.DiffOpts` using the shared definitions in `pkg/common/diffs/diffs.go`, instead of ad-hoc `cmp.Option` variadics.
- Errors are wrapped with context (`failed to ...` with `namespace/name`) at the call sites.
- Client operations use `ctx.Context` rather than `context.TODO()`.

No functional/behavioral change is intended — this is a pure refactoring.

N/A — internal refactoring with no user-visible change.

Follow-up of the ongoing `K8sClient` wrapper adoption; this is the final step that allows `pkg/deploy/sync.go` to be deleted.

### How to test this PR?

1. Deploy the operator:
 
#### OpenShift
```bash

OPERATOR_IMAGE=<...>
oc wait --for=condition=Ready pod -l app.kubernetes.io/component=che-operator -n eclipse-che --timeout=60s
```

#### on Minikube

```bash
```

2. Verify that the operator reconciles cleanly: all resources (Deployments, Services, Routes/Ingresses, ConfigMaps, Secrets, RBAC) are created, owner references are set on objects in the `eclipse-che` namespace, and there is no infinite reconciliation loop in the operator logs.

3. Unit tests: `make test`.

#### Common Test Scenarios
- [ ] Deploy Eclipse Che
- [ ] Stop a workspace

### PR Checklist

### Reviewers


tolusha:

**Available commands**:

pkg/deploy/deployment.go
```diff
@@ -66,28 +69,43 @@ func SyncDeploymentSpecToCluster(
...
-	if !exists || err != nil {
-		return false, err
+	exists, err := deployContext.ClusterAPI.ClientWrapper.GetIgnoreNotFound(deployContext.Context, key, actual)
+	if err != nil {
+		return false, fmt.Errorf("failed to get Deployment %s/%s: %w", key.Namespace, key.Name, err)
+	} else if !exists {
+		return false, nil
 	}
 
 	provisioned := actual.Status.UnavailableReplicas == 0
```
tolusha: Readiness is now reported on the same pass that creates or updates the Deployment.

On `main` this was unreachable after a write:

```go
done, err := Sync(deployContext, deploymentSpec, deploymentDiffOpts)
if err != nil || !done {
    ...
    return false, err   // created/updated -> not provisioned, requeue
}
```

Now `Sync` returns only `error`, so a successful create or update falls straight through to here. During a rolling update the `GetIgnoreNotFound` on line 104 goes through the cached client and can return the pre-update object, whose `UnavailableReplicas` is still 0. We then return `(true, nil)` mid-rollout. Six callers gate on this bool - `gateway.go:163`, `server_reconciler.go:156`, `dashboard.go:91`, `pluginregistry.go:178` and both openvsx deployments - so `chePhase` can reach `Active` while the old pods are still serving. On a fresh create the same expression is true because `.status` is empty.

The current tests cannot catch this: the fake client never populates deployment status, so old and new return identical values.

A generation check is correct regardless of cache freshness:

```suggestion
	provisioned := actual.Status.ObservedGeneration >= actual.Generation &&
		actual.Status.UnavailableReplicas == 0
```

Longer term, would it make sense for `Sync` to report whether it changed anything? Every reconciler here is built on the `(reconcile.Result, bool, error)` contract, and "did I have to act?" is something only the sync layer knows.

controllers/usernamespace/usernamespace_controller.go
```diff
@@ -387,8 +391,15 @@ func (r *CheUserNamespaceReconciler) reconcileSelfSignedCert(ctx context.Context
 		Immutable: cheCert.Immutable,
 	}
 
-	_, err := deploy.Sync(deployContext, targetCert, deploy.SecretDiffOpts)
-	return err
+	if err := r.nonCachedClientWrapper.Sync(
```
tolusha, line 395: `Secret` was dropped from the delete-and-recreate set, and this is the call site where it bites.

Deleted `pkg/deploy/sync.go`:
```go
func isUpdateUsingDeleteCreate(kind string) bool {
    return kind == "Service" || kind == "Ingress" || kind == "Route" || kind == "Job" || kind == "Secret"
}
```

`pkg/common/k8s-client/k8s_client.go:317`:
```go
func (k K8sClientWrapper) isRecreate(kind string) bool {
    return kind == "Service" || kind == "Ingress" || kind == "Route" || kind == "Job"
}
```

`targetCert` a few lines above copies `Immutable: cheCert.Immutable` from the source Secret. Per `k8s.io/api/core/v1`, immutable means "data stored in the Secret cannot be updated (only object metadata can be modified)". So if an admin marks the CA-cert Secret immutable, `Update` is rejected with `field is immutable` when `ca.crt` rotates - and unlike `SyncDeploymentSpecToCluster`, this path has no delete-and-retry. The reconciler then fails on every loop and the copy never converges. `Secret.Type` is immutable too.

Worth noting the recreate branch in `doSync` has no test coverage at all (`k8s_client_sync_test.go` has only `TestSync` and `TestSyncAndMergeLabelsAnnotations`), which is why nothing flagged the omission. A table-driven test over `isRecreate` would be ~15 lines and would close it permanently.

pkg/deploy/clusterrole.go
```diff
@@ -24,10 +27,19 @@ import (
 func SyncClusterRoleToCluster(
 	deployContext *chetypes.DeployContext,
 	name string,
-	policyRule []rbac.PolicyRule) (bool, error) {
+	policyRule []rbac.PolicyRule) error {
 
 	crSpec := getClusterRoleSpec(deployContext, name, policyRule)
-	return Sync(deployContext, crSpec, diffs.ClusterRole)
+
+	if err := deployContext.ClusterAPI.ClientWrapper.Sync(
```
tolusha, line 34: Cluster-scoped objects moved from the non-caching client to the cached one. Was that deliberate?

The deleted `getClientForObject` routed by namespace:
```go
// empty namespace (cluster scope object) or object in another namespace
if deployContext.CheCluster.Namespace == objectNamespace { return deployContext.ClusterAPI.Client }
return deployContext.ClusterAPI.NonCachingClient
```
A cluster-scoped object has `Namespace == ""`, so it always went to `NonCachingClient`.

Syncs are fine today because `getClusterRoleSpec` calls `GetLabels(...)`, which sets `part-of=che.eclipse.org`, and that matches the cache selector in `cmd/main.go:387-392`. So this is about consistency rather than a live bug here.

The reason to ask: the same PR takes the opposite position a few files over. `consolelink.go` uses `NonCachingClientWrapper` with the comment "ConsoleLink is a cluster scoped object, so it can't have an owner reference and must be synced with the non-caching client", and `identity_provider_reconciler.go` syncs the OAuthClient cached but deletes it non-cached. Three answers to one question in one PR.

Could we put the old rule back in one place - something like `deployContext.ClientFor(obj)` - so the choice is mechanical instead of a judgement call at 40 sites? And whichever way it goes, the `consolelink.go` comment should match.

pkg/deploy/server/rbac.go
```diff
@@ -109,14 +121,22 @@ func (s *CheServerReconciler) deletePermissions(ctx *chetypes.DeployContext) boo
...
+				&rbacv1.ClusterRoleBinding{},
+			); err != nil {
 				done = false
 				logrus.Errorf("Failed to delete ClusterRoleBinding '%s', cause: %v", name, err)
 			}
 
 			// Removes any legacy CRB https://github.com/eclipse/che/issues/19506
 			legacyName := ctx.CheCluster.Namespace + "-" + constants.DefaultCheServiceAccountName + "-" + name
-			if _, err := deploy.Delete(ctx, types.NamespacedName{Name: legacyName}, &rbacv1.ClusterRoleBinding{}); err != nil {
+			if err := ctx.ClusterAPI.ClientWrapper.DeleteByKeyIgnoreNotFound(
```
tolusha: This delete can silently no-op, which matters because it is a cleanup path.

`DeleteByKeyIgnoreNotFound` does a `Get` first and returns `nil` when the object is not found (`k8s_client.go:164-183`). With the cached client that `Get` only sees ClusterRoleBindings carrying `part-of=che.eclipse.org`. On `main` this went through `NonCachingClient` and hit the API server directly.

The legacy CRB is the awkward case: by definition it was created by an older operator version, so we cannot assume it carries the label. If it does not, the delete reports success, `done` stays true, the finalizer is removed, and a cluster-wide binding to a deleted ServiceAccount survives with no owner reference to garbage-collect it and nothing to retry. Same shape applies to `Finalize` in `dashboard.go:101,110` and `rbac/gateway_permissions.go:69,77`.

I could not confirm from a shallow clone whether older versions labelled these, so flagging rather than asserting. Using `NonCachingClientWrapper` for cluster-scoped deletes would remove the question entirely.

pkg/deploy/secret.go
```diff
@@ -31,15 +34,28 @@ var SecretDiffOpts = cmp.Options{
...
 	deployContext *chetypes.DeployContext,
 	name string,
 	namespace string,
-	data map[string][]byte) (bool, error) {
+	data map[string][]byte) error {
 
 	secretSpec := GetSecretSpec(name, namespace, data)
-	return Sync(deployContext, secretSpec, SecretDiffOpts)
+
+	if err := controllerutil.SetControllerReference(deployContext.CheCluster, secretSpec, deployContext.ClusterAPI.Scheme); err != nil {
```
tolusha, line 44: Two small things on this function.

The owner reference is set unconditionally even though the function takes an explicit `namespace`. The deleted `setOwnerReferenceIfNeeded` gated exactly this on `blueprint.GetNamespace() == deployContext.CheCluster.Namespace`. Every caller today passes `ctx.CheCluster.Namespace`, so nothing is broken - but `controllerutil.validateOwner` rejects a cross-namespace owner, and the signature invites someone to pass a user namespace. Dropping the parameter or restoring the check would make the constraint visible.

Separately, `SyncOptions` here leaves `SuppressDiff` at false, and `SecretDiffOpts` only ignores `TypeMeta` and `ObjectMeta`. So `k8s_client.go:268` prints the Secret's `Data` in plaintext to stdout whenever it changes. That is pre-existing behaviour - `sync.go:263-267` did the same - but this PR is touching the call site anyway:

```suggestion
		&k8sclient.SyncOptions{DiffOpts: SecretDiffOpts, SuppressDiff: true},
```

(The suggestion applies to line 53; same for `usernamespace_controller.go:397`.)

pkg/deploy/gateway/gateway.go
```diff
@@ -158,22 +165,45 @@ func syncAll(deployContext *chetypes.DeployContext) (bool, error) {
...
+			return false, err
 		}
 	}
 
 	return true, nil
 }
 
+// syncGatewayObject syncs the given object into the CheCluster namespace.
+func syncGatewayObject(deployContext *chetypes.DeployContext, obj client.Object, diffOpts ...cmp.Option) error {
+	kind := obj.GetObjectKind().GroupVersionKind().Kind
```
tolusha: `kind` is read before `Sync` populates the GVK, so it is empty in every error from this helper.

Specs built as Go literals without `TypeMeta` have no `Kind` until `K8sClientWrapper.ensureGVK` runs inside `Sync`. The messages come out as `failed to sync  eclipse-che/che-gateway`, with a blank where the useful token should be.

```suggestion
	kind := k8sclient.GetObjectType(obj)
```

Related: it would help to document on the `K8sClient` interface that `Sync` mutates its argument - it sets the GVK, overwrites `ResourceVersion`, and optionally merges labels and annotations. The current doc comment only says it "ensures that the object is up to date in the cluster".

pkg/deploy/route_test.go
```diff
@@ -171,14 +171,12 @@ func TestRouteSpec(t *testing.T) {
 func TestSyncRouteToCluster(t *testing.T) {
 	ctx := test.NewCtxBuilder().Build()
 
-	done, err := SyncRouteToCluster(ctx, "test", "", "service", 80, "test")
+	err := SyncRouteToCluster(ctx, "test", "", "service", 80, "test")
 	assert.Nil(t, err)
```
tolusha, line 173: This test is now weaker than before. It used to assert `done` four times; it now only checks that no error came back.

Same pattern in `rolebinding_test.go`, `secret_test.go`, `service_test.go`, `clusterrole_test.go`, `clusterrolebinding_test.go`, `ingress_test.go`, `job_test.go` and `editors_definitions_test.go` - each drops `assert.True(t, done)` with nothing in its place.

`service_account_test.go` in this same PR shows the move that keeps the coverage:

```go
err := SyncServiceAccountToCluster(ctx, "test")
assert.NoError(t, err)

exists, err := ctx.ClusterAPI.ClientWrapper.GetIgnoreNotFound(
    ctx.Context,
    types.NamespacedName{Name: "test", Namespace: ctx.CheCluster.Namespace},
    &corev1.ServiceAccount{},
)
assert.NoError(t, err)
assert.True(t, exists)
```

Applying that to the other eight would restore what was lost, and for a PR whose claim is behaviour preservation it is the coverage that matters most.

tolusha (commented):
Reviewed the full diff (57 files, 3,617 lines) against `main`, and verified locally on the PR head with Go 1.26.4:

- `go build ./...` - passes
- `go vet ./pkg/... ./controllers/...` - passes
- `go test ./pkg/... ./controllers/...` - only `pkg/deploy/gateway` fails, and the same tests fail identically on `main` in my sandbox (missing `/tmp/header-rewrite-traefik-plugin/headerRewrite.go` fixture, plus `TestReadProxyConfiguration` picking up the sandbox's own `172.30.0.1`). No PR-introduced test failures.

The migration itself is the right call, and splitting it into seven commits - one per removed helper - makes a change this size genuinely bisectable. The error wrapping is a real upgrade: `failed to sync Deployment eclipse-che/che-server: %w` beats the bare propagation it replaces, and `%w` keeps `apierrors` unwrapping intact. Moving from `context.TODO()` to `ctx.Context` means cancellation finally reaches the API calls. The before/after mapping table in the description is exactly what a reviewer needs.

My main concern is with the framing: "No functional/behavioral change is intended - this is a pure refactoring." I found three places where that does not hold, and they are all silent failures rather than loud ones. Details inline, summarised here:

1. **Deployment readiness** (`deployment.go:111`) - the `!done` short-circuit is gone, so we now compute `provisioned` from a cached read taken immediately after a write. Mid-rollout this reports ready. Six callers gate on that bool.
2. **`Secret` left the delete-and-recreate set** - `isUpdateUsingDeleteCreate` had it, `isRecreate` does not. Immutable Secrets on the user-namespace CA-cert path now fail permanently, with no recovery on that call site.
3. **Cluster-scoped deletes moved behind the label-filtered cache** - `DeleteByKeyIgnoreNotFound` returns `nil` on a cache miss, so a finalizer can report success while leaving a cluster-wide RBAC binding behind.

The thread running through all three is that `sync.go` was not only a pile of helpers. It was where two cross-cutting policies lived: which client for which object (`getClientForObject`), and when to attach an owner reference (`shouldSetOwnerReferenceForObject`). Deleting it without relocating those policies is what produced every finding above. The client rule is now answered three different ways inside this one PR, and the owner-reference rule is spread over 45 unguarded call sites.

The reason none of this fails a test is worth calling out on its own: `k8s_client_sync_test.go` has two tests, and grepping it for `Service`, `Ingress`, `Route`, `Job` or `isRecreate` returns nothing - **the delete-and-recreate branch has zero coverage**, which is exactly the branch whose membership changed. And the fake client never populates deployment status, so old and new `SyncDeploymentSpecToCluster` return identical values under test. Both regressions are invisible to the suite by construction.

A few smaller things I did not comment inline:

- The same anonymous-struct delete loop is byte-for-byte identical in `devfileregistry.go:48`, `pluginregistry.go:49` and `postgres.go:42`. Three copies feels like the point to extract a helper - `SyncTarget` in `k8s_client_types.go` already gives a declarative shape for this, and `metrics/metrics.go` uses it.
- Logging is now three systems in one package: `ctrl.Log` in `expose`/`consolelink`/`identity-provider`, `logrus` in `deployment.go:113`, `server/rbac.go:105`, `dashboard.go:107`, and raw `fmt.Printf` in `k8s_client.go:268`. Only the first honours `LOG_LEVEL`. Fine as an increment - maybe worth an issue so it converges?
- `SyncDeploymentSpecToCluster` issues three `Get` calls (lines 70, 104, plus one inside `Sync`). Same count as `main`, so no regression, but the first two could probably fold together.
- `tls/tls_utils.go:400,415,444,459` still uses `ctx.ClusterAPI.Client` with `context.TODO()`, in a file this PR already modifies. Might be worth finishing while it is open.
- One behaviour change that looks like an improvement but is not mentioned in the description: `identity_provider_reconciler.go` now calls `DeleteFinalizer(OAuthFinalizerName)` unconditionally, whereas `main` only reached it inside `if oauthClient != nil`. That closes a potential CheCluster-deletion hang. Good change - it just is not a no-op.

On compatibility: no CRD, CSV, webhook or `CheCluster` change, so nothing user-visible and the upgrade path is clean. Internally ~15 exported signatures change and 10 functions are deleted, which is contained for a main module but does break anyone vendoring `pkg/deploy`. One line in the description would save them the surprise.

Finally - since the PR is still a draft, openshift-ci skipped the run. A 57-file change to the operator's create/update/delete plumbing is precisely where the e2e signal is worth more than the unit tests, and all three issues above are invisible to the unit suite anyway. Converting to ready (or `/test all`) before merge would help a lot.

Nice work on the commit hygiene here - that part made the review much easier than it could have been.

pkg/deploy/identity-provider/identity_provider_reconciler.go
```diff
@@ -13,26 +13,33 @@
...
 const (
 	OAuthFinalizerName = "oauthclients.finalizers.che.eclipse.org"
 )
 
+var logger = ctrl.Log.WithName("identity-provider")
+
 var (
-	oAuthClientDiffOpts = cmpopts.IgnoreFields(oauth.OAuthClient{}, "TypeMeta", "ObjectMeta")
+	oAuthClientDiffOpts = cmp.Options{
+		cmpopts.IgnoreFields(oauth.OAuthClient{}, "TypeMeta", "ObjectMeta"),
```
tolusha, line 41: `oAuthClientDiffOpts` ignores only `TypeMeta` and `ObjectMeta`, so `OAuthClient.Secret` - a plain string - takes part in the comparison. When it changes, `doSync` runs `fmt.Printf("Difference:\n%s", diff)` (`k8s_client.go:268`) and both the old and the new client secret land on stdout, where the cluster log aggregator collects them. That puts the credential outside the encryption boundary it is supposed to live behind.

`SyncOptions.SuppressDiff` is exactly the fix. It is currently used in one place in the whole tree (`workspaces_config_controller.go:538`) and at none of the call sites this PR migrated - so passing `&k8sclient.SyncOptions{DiffOpts: oAuthClientDiffOpts, SuppressDiff: true}` at line 114 would close this.

Would it make sense to go further and suppress by kind inside `doSync` for `Secret` and `OAuthClient`? That way the next person adding a credential-bearing sync does not have to remember, and it would also close the plaintext half of the `secret.go:45` thread.

pkg/deploy/clusterrolebinding.go
```diff
@@ -25,10 +28,19 @@ func SyncClusterRoleBindingToCluster(
...
 	name string,
 	serviceAccountName string,
-	clusterRoleName string) (bool, error) {
+	clusterRoleName string) error {
 
 	crbSpec := getClusterRoleBindingSpec(deployContext, name, serviceAccountName, deployContext.CheCluster.Namespace, clusterRoleName)
-	return Sync(deployContext, crbSpec, diffs.ClusterRoleBinding)
+
+	if err := deployContext.ClusterAPI.ClientWrapper.Sync(
+		deployContext.Context,
```
tolusha, line 36: `diffs.ClusterRoleBinding` ignores only `TypeMeta`/`ObjectMeta`, so `RoleRef` is compared - but `isRecreate` covers only `Service`, `Ingress`, `Route` and `Job`. Any `roleRef` difference therefore resolves to `cli.Update`, which the API server rejects with `cannot change roleRef`, permanently, on every reconcile.

This is reachable from user input: `server/rbac.go:55-60` passes each entry of `spec.components.cheServer.clusterRoles` as both the CRB name and the role name, so a pre-existing CRB of that name pointing at a different role wedges the reconciler.

It predates this PR, but `isRecreate` is a function we are already editing, so this seems like the moment:

```go
func (k K8sClientWrapper) isRecreate(kind string) bool {
	return kind == "Service" || kind == "Ingress" || kind == "Route" || kind == "Job" ||
		kind == "RoleBinding" || kind == "ClusterRoleBinding"
}
```

Longer term, could we drop the kind list entirely and recreate on an `IsInvalid` response instead? That expresses the actual rule - recreate when the API server says a field is immutable - rather than a snapshot of which kinds have immutable fields today, and it would have caught the `Secret` case automatically.

One more thing about that function: it matches on the unqualified kind string, so a CRD named `Job` or `Service` in another API group would silently be delete-and-recreated.

pkg/deploy/gateway/oauth_proxy.go
```diff
@@ -232,7 +232,11 @@ func resolveOpenShiftOAuthProxyImage(ctx *chetypes.DeployContext) string {
 func getOauthProxyContainerSpec(ctx *chetypes.DeployContext) corev1.Container {
 	// append env var with ConfigMap revision to restore pod automatically when config has been changed
 	cm := &corev1.ConfigMap{}
-	exists, _ := deploy.GetNamespacedObject(ctx, "che-gateway-config-oauth-proxy", cm)
+	exists, _ := ctx.ClusterAPI.ClientWrapper.GetIgnoreNotFound(
+		ctx.Context,
```
tolusha, line 235: This read can return a stale resourceVersion now.

The old `deploy.Sync` returned `done=false` specifically after a successful in-place `Update` (`sync.go` `doUpdate` - creates returned `true`), so every `if done, err := deploy.Sync(...); !done { return done, err }` in `syncAll` was a "this object just changed, bail out and requeue" barrier, not just error handling. `syncGatewayObject` returns `nil` on a successful update, so we fall straight through.

Concretely: `syncAll` updates `che-gateway-config-oauth-proxy` around `gateway.go:119-124`, then calls `getGatewayDeploymentSpec` in the same pass, and this `GetIgnoreNotFound` hits the manager cache before the informer has caught up. `CM_REVISION` is built from the pre-update resourceVersion, so the gateway Deployment is not rolled - which is the whole point of that env var.

It self-heals on the reconcile triggered by the ConfigMap watch, so this is an extra round trip rather than a hang. But `Sync` writes the new resourceVersion back into the object it just updated, so we could thread `oauthProxyConfig.GetResourceVersion()` through instead of re-reading from the cache.

Worth checking the other members of `syncAll` for the same assumption - the barrier is gone for the ServiceAccount, Role, RoleBinding, oauth Secret and four ConfigMaps too.

pkg/deploy/devfileregistry/devfileregistry.go
```diff
@@ -43,10 +45,20 @@ func NewDevfileRegistryReconciler() *DevfileRegistryReconciler {
...
+		name string
+		obj  client.Object
+	}{
+		{constants.DevfileRegistryName, &corev1.Service{}},
+		{constants.DevfileRegistryName, &corev1.ConfigMap{}},
+		{gateway.GatewayConfigMapNamePrefix + constants.DevfileRegistryName, &corev1.ConfigMap{}},
+		{constants.DevfileRegistryName, &appsv1.Deployment{}},
+	}
+
+	for _, object := range objects {
```
tolusha, line 57: The same operation - delete an object that may not exist - is now handled three ways in this PR:

- here, plus `pluginregistry.go` and `postgres.go`: `_ =`, dropped and not logged
- `expose.go:65-67`: `logger.Error(...)`, continue
- `usernamespace_controller.go:429-436`: `fmt.Errorf(...)`, propagate

Each is defensible on its own. The `_ =` variant is the awkward one operationally: if the API server is degraded, these four deletes fail silently on every reconcile and nothing records it. Could we settle on the `expose.go` behaviour - log and continue - across all three files?

Separately, does the table plus loop earn its keep here? It is 16 lines where the four straight-line calls were 4, the anonymous struct cannot be named or reused, and the swallowed error came along unchanged.

pkg/deploy/tls/tls_utils.go
```diff
@@ -153,9 +153,13 @@ func GetTLSCrtChain(ctx *chetypes.DeployContext) ([]*x509.Certificate, error) {
 		route := &routev1.Route{}
 		for {
 			time.Sleep(time.Duration(1) * time.Second)
-			exists, err := deploy.GetNamespacedObject(ctx, routeSpec.Name, route)
+			exists, err := ctx.ClusterAPI.ClientWrapper.GetIgnoreNotFound(
+				ctx.Context,
```
tolusha, line 157: If the Route never appears this spins forever - no iteration cap, no deadline - and although `ctx.Context` is passed into the `Get`, the loop itself never checks it, so it keeps sleeping after the reconcile context is cancelled. It also holds the reconcile worker for the whole duration, so one stuck CheCluster stalls the operator.

Pre-existing, but since we are rewriting these lines anyway:

```go
		select {
		case <-ctx.Context.Done():
			return nil, ctx.Context.Err()
		case <-time.After(time.Second):
		}
```

Same applies to the Ingress loop around line 200.

pkg/deploy/networkpolicies/networkpolicies.go
```diff
@@ -22,7 +22,7 @@ import (
 	k8sclient "github.com/eclipse-che/che-operator/pkg/common/k8s-client"
 	defaults "github.com/eclipse-che/che-operator/pkg/common/operator-defaults"
 	"github.com/eclipse-che/che-operator/pkg/common/reconciler"
-	"github.com/eclipse-che/che-operator/pkg/deploy"
+	labels2 "github.com/eclipse-che/che-operator/pkg/deploy"
```
tolusha:
```suggestion
	"github.com/eclipse-che/che-operator/pkg/deploy"
```

`labels2` reads like an IDE artifact, and it drags 9 call-site renames plus a copyright bump into an otherwise no-op file. Reverting it would shrink the diff noticeably.

tolusha (commented):
## Automated review (additive to the 7 inline comments already on this PR)

The refactor itself is sound and worth landing - removing `getClientForObject` and
`shouldSetOwnerReferenceForObject` makes two previously-implicit decisions visible at each call
site, the error wrapping across ~35 files is a real operability win, and `syncGatewayObject` is a
good extraction. I read all 59 changed files.

Six new inline comments below. A few things that do not attach to a single line:

**The cache configuration was not revisited.** `cmd/main.go` has zero diff, but the set of kinds
read through the label-filtered cache expanded to include cluster-scoped RBAC objects. For a
pre-existing *unlabeled* object the cached `Get` misses, `doSync` takes the create path, and
`doCreate(..., false)` surfaces `AlreadyExists` - a hard failure on every reconcile, not a no-op.
The one-time migration mitigates this but only relabels objects carrying
`app.kubernetes.io/instance=<flavor>` or `app=<flavor>`. Could we either state in the description
that the migration is a hard prerequisite, or fall back to a non-cached `Get` before concluding an
object is absent?

**The `K8sClient` interface has zero users.** `k8s_client_types.go:24-48` defines a complete
7-method interface, but `grep -rn "k8sclient\.K8sClient\b"` returns nothing - `ClusterAPI` holds
the concrete `*K8sClientWrapper`. Typing those two fields as the interface costs nothing (the
methods are all on value receivers) and would unblock the testing gap below. A
`var _ K8sClient = (*K8sClientWrapper)(nil)` assertion would stop the interface drifting from the
implementation.

**Nothing can test the cached/non-cached split.** `pkg/common/test/deploy_context.go:75-76` wraps
the *same* fake client for both `ClientWrapper` and `NonCachingClientWrapper`. The riskiest
dimension of this PR - did each of ~35 migrated call sites pick the right client? - cannot be
caught by any unit test, and a future swap of one for the other stays green.

**`TestSync` does not verify `Sync`.** `k8s_client_sync_test.go:38-97` never reads the object back
from the fake client; it asserts `Kind`/`APIVersion` on the *blueprint*, which only exercises
`ensureGVK`. The second and third blocks are indistinguishable, so the test passes if `doSync` is
replaced with `return nil`. The `sync_test.go` being deleted here was stronger on this point - its
`TestUpdate` read back and asserted `actual.Labels["a"] == "b"`. (To be clear, the deletion itself
is fine: `TestCreateIfNotExists`, `TestDelete` and friends already cover the
already-exists/not-found cases, and those two *do* read back correctly.) Separately, `isRecreate`
has no test at all - no Service, Ingress, Route or Job is ever synced in that suite - which is
awkward given that a membership change to `isRecreate` is this PR's highest-risk change.

**Smaller things.** Copyright headers jumped to 2026 on files whose only change is an import alias.
`syncAll` and `SyncGatewayToCluster` still return `(bool, error)` but can now only produce
`(false, err)` or `(true, nil)`; collapsing them to `error` would match the rest of the PR.
`deleteByKeyIgnoreNotFound` does a `Get` before the `Delete` even though
`doDeleteIgnoreIfNotFound` already swallows `IsNotFound`. And
`chetypes/types.go:37-38` still says `// We should ClientWrapper instead wherever it possible`,
which now either describes finished work or names the remaining stragglers.

**Please consider expanding the description.** It says no functional or behavioural change is
intended, but several changes are behavioural - `Immutable` dropped, `Secret` removed from
`isRecreate`, owner references now unconditional at several sites, and the `done=false`-after-update
barrier removed. Right now reviewers have to reverse-engineer which deltas are deliberate.

**CI**: `helm-on-minikube` fails with
`failed calling webhook "webhook.cert-manager.io" ... dial tcp 10.101.37.78:443: connect: connection refused`
while applying the DevWorkspace Operator manifest - a cert-manager readiness flake, unrelated to
this PR. Everything else is green. The PR is still a draft, which is why CodeRabbit skipped it.

pkg/deploy/secret.go
tolusha: The first half of this is now stale - the latest commit removed the `namespace` parameter from `SyncSecretToCluster`, so it always targets the CheCluster namespace and the unconditional owner reference is safe there.

The plaintext-diff half still stands. `SyncOptions.SuppressDiff` already exists and would close it; I left a note on `identity_provider_reconciler.go` suggesting we suppress by kind inside `doSync` for `Secret` and `OAuthClient` so it is handled centrally rather than per call site.

pkg/deploy/clusterrole.go
tolusha, line 34: **Cluster-scoped Sync goes through the label-filtered cache**

The old `getClientForObject` sent every cluster-scoped object to `NonCachingClient`, because `Namespace == ""` never matches `CheCluster.Namespace`. `ClientWrapper` wraps the manager cache, and `cmd/main.go` restricts `ClusterRole` with `cache.ByObject{Label: partOfEclipseChe}`.

The selector matches the live object's labels, not the blueprint's. For an existing unlabeled ClusterRole: cached Get returns NotFound, `doSync(nil)` calls `doCreate(ignoreIfAlreadyExists=false)`, the API returns `AlreadyExists`, and the error propagates. Nothing on this path ever adds the label, so it fails on every reconcile from then on. The old code found the object and added the label as part of the update.

Would it make sense to switch to `NonCachingClientWrapper` here, matching what `consolelink.go` already does and documents?

pkg/deploy/clusterrolebinding.go
```diff
@@ -25,10 +28,19 @@ func SyncClusterRoleBindingToCluster(
 	deployContext *chetypes.DeployContext,
 	name string,
 	serviceAccountName string,
-	clusterRoleName string) (bool, error) {
+	clusterRoleName string) error {
 
 	crbSpec := getClusterRoleBindingSpec(deployContext, name, serviceAccountName, deployContext.CheCluster.Namespace, clusterRoleName)
-	return Sync(deployContext, crbSpec, diffs.ClusterRoleBinding)
+
+	if err := deployContext.ClusterAPI.ClientWrapper.Sync(
```
tolusha, line 35: **Same issue for ClusterRoleBinding, and this one is reachable from user config**

Same mechanism as the ClusterRole comment above.

This one is reachable without an upgrade. `pkg/deploy/server/rbac.go:58` syncs a ClusterRoleBinding named from `Spec.Components.CheServer.ClusterRoles`, which is admin-supplied and will not carry `app.kubernetes.io/part-of=che.eclipse.org`.

The one-time migrator does not cover it either - `addPartOfCheLabelForObjectsWithLabel` only relabels ClusterRole/ClusterRoleBinding that already have `instance=che` or `app=che`, and `addPartOfCheLabeltoUserDefinedObjects` handles only Secrets and ConfigMaps.

pkg/deploy/identity-provider/identity_provider_reconciler.go
```diff
@@ -93,15 +107,18 @@ func syncOAuthClient(ctx *chetypes.DeployContext) (bool, error) {
 		redirectURIs,
 		ctx.CheCluster.Spec.Networking.Auth.OAuthAccessTokenInactivityTimeoutSeconds,
 		ctx.CheCluster.Spec.Networking.Auth.OAuthAccessTokenMaxAgeSeconds)
-	done, err := deploy.Sync(ctx, oauthClientSpec, oAuthClientDiffOpts)
-	if !done {
-		return false, err
+
+	if err := ctx.ClusterAPI.ClientWrapper.Sync(
```
tolusha: **OAuthClient is read with one client and written with another**

`GetOAuthClient` reads through `NonCachingClientWrapper` (`identity_provider_util.go:55`), and this `Sync` writes through `ClientWrapper`. `cmd/main.go:414` puts `OAuthClient` behind the label selector when `IsOpenShiftOAuthEnabled()`.

For an existing OAuthClient without the `part-of` label the two clients disagree: the read finds it, so the reconciler correctly reuses its name and secret, then the cached Get misses and `Sync` fails with `AlreadyExists`. OpenShift authentication never converges.

Base behaviour was consistent here - `deploy.Sync` routed to `NonCachingClient` for cluster-scoped objects, matching the read side.

```suggestion
	if err := ctx.ClusterAPI.NonCachingClientWrapper.Sync(
```

controllers/usernamespace/usernamespace_controller.go
```diff
@@ -383,12 +387,18 @@ func (r *CheUserNamespaceReconciler) reconcileSelfSignedCert(ctx context.Context
 		Data: map[string][]byte{
 			"ca.crt": cheCert.Data["ca.crt"],
 		},
-		Type:      cheCert.Type,
-		Immutable: cheCert.Immutable,
+		Type: cheCert.Type,
```
tolusha, line 392: **Dropping Immutable cannot be applied to an existing immutable Secret**

`deploy.SecretDiffOpts` ignores only `TypeMeta` and `ObjectMeta`, so `Immutable` is compared. On clusters where the source CA secret was immutable, the existing target Secret has `Immutable: &true` while the new spec has `nil`. The diff is non-empty, `Secret` is not in `isRecreate`, so this becomes an `Update` that the API server rejects with `field is immutable` - permanently, in every affected user namespace.

Worth noting the old `isUpdateUsingDeleteCreate` did include `Secret`, so the previous code would have recreated its way out.

Could we restore the field?

```suggestion
		Immutable: cheCert.Immutable,
```

If dropping it is intentional, it needs a migration that deletes the existing immutable target Secrets, plus `Secret` back in `isRecreate`. Either way it is a behaviour change worth calling out in the description.

pkg/deploy/secret.go
```diff
@@ -31,15 +34,27 @@ var SecretDiffOpts = cmp.Options{
...
+	secretSpec := GetSecretSpec(name, deployContext.CheCluster.Namespace, data)
+
+	if err := controllerutil.SetControllerReference(deployContext.CheCluster, secretSpec, deployContext.ClusterAPI.Scheme); err != nil {
+		return fmt.Errorf("failed to set owner reference for Secret %s/%s: %w", secretSpec.Namespace, name, err)
+	}
+
+	if err := deployContext.ClusterAPI.ClientWrapper.Sync(
+		deployContext.Context,
+		secretSpec,
+		&k8sclient.SyncOptions{DiffOpts: SecretDiffOpts},
```
tolusha, line 51: **Secret contents can be printed to the log in plaintext**

`doSync` prints the diff whenever `len(DiffOpts) != 0 && !SuppressDiff`. `SecretDiffOpts` is non-empty, so any change to a Secret's `Data` prints the decoded before/after to stdout - and container logs usually reach a much wider audience than `get secret` RBAC.

Same fix that was applied to the OAuthClient sync in `e5788502`:

```suggestion
		&k8sclient.SyncOptions{DiffOpts: SecretDiffOpts, SuppressDiff: true},
```

pkg/deploy/server/rbac.go
```diff
@@ -95,12 +99,20 @@ func (s *CheServerReconciler) deletePermissions(ctx *chetypes.DeployContext) boo
 	done := true
 
 	for _, name := range names {
-		if _, err := deploy.Delete(ctx, types.NamespacedName{Name: name}, &rbacv1.ClusterRole{}); err != nil {
+		if err := ctx.ClusterAPI.ClientWrapper.DeleteByKeyIgnoreNotFound(
```
tolusha, line 102: **Cluster-scoped deletes through the cached client can leave RBAC behind**

`deleteByKeyIgnoreNotFound` treats a cache miss as "already gone" and returns nil. For an unlabeled ClusterRole or ClusterRoleBinding that means finalization reports success while the object stays bound to the service account after the CheCluster is deleted.

Line 135 already uses `NonCachingClientWrapper` for exactly this reason. Does it make sense to use it for lines 102, 111, and 124 as well?

pkg/deploy/expose/expose.go
```diff
@@ -73,20 +83,25 @@ func exposeWithGateway(deployContext *chetypes.DeployContext,
...
+
+	if err := deployContext.ClusterAPI.ClientWrapper.Sync(
+		deployContext.Context,
+		cfg,
+		&k8sclient.SyncOptions{DiffOpts: diffs.ConfigMapEnsureLabels},
+	); err != nil {
+		return "", false, fmt.Errorf("failed to sync ConfigMap %s/%s: %w", cfg.Namespace, cfg.Name, err)
 	}
 
 	cleanUpRouting()
```
tolusha, line 101: **cleanUpRouting now runs in the same pass as the gateway config write**

The old code returned early when `deploy.Sync` reported `done == false`, so the legacy Route or Ingress was deleted only on a later pass, after the gateway ConfigMap update had been observed. Now we always fall through to `cleanUpRouting()`.

On a route change this opens a window where the old Route is already gone and the gateway has not reloaded its config yet.

This is the main casualty of dropping the `(bool, error)` return - there is no longer any way for a caller to know whether a `Sync` mutated anything. Would it be worth having `Sync` return that, so the ordering barrier can be restored?

pkg/deploy/devfileregistry/devfileregistry.go
```diff
@@ -43,10 +45,20 @@ func NewDevfileRegistryReconciler() *DevfileRegistryReconciler {
...
+		key := types.NamespacedName{Name: object.name, Namespace: ctx.CheCluster.Namespace}
+		_ = ctx.ClusterAPI.ClientWrapper.DeleteByKeyIgnoreNotFound(ctx.Context, key, object.obj)
```
tolusha, line 59: **Delete error is discarded here but propagated elsewhere**

`DeleteByKeyIgnoreNotFound` already folds `NotFound` into nil, so a non-nil return here is a real API failure - an RBAC denial, a webhook rejection, an apiserver outage. Right now it is dropped with no log line.

Same pattern at `pluginregistry.go:61` and `postgres.go:55`, while structurally identical deletes elsewhere in the PR propagate. Could we either return the error or add a short comment explaining why it is safe to drop, the way `consolelink.go` does for its deliberate swallow?

pkg/deploy/secret.go
```diff
@@ -31,15 +34,27 @@ var SecretDiffOpts = cmp.Options{
 	cmpopts.IgnoreFields(corev1.Secret{}, "TypeMeta", "ObjectMeta"),
 }
 
-// SyncSecret applies secret into cluster or external namespace
+// SyncSecretToCluster applies secret into cluster or external namespace
```
tolusha, line 36: **Doc comment no longer matches the signature**

The `namespace` parameter is gone and the only caller (`pkg/deploy/tls/tls_utils.go:550`) is in-namespace.

```suggestion
// SyncSecretToCluster applies secret into the Che installation namespace
```

tolusha (commented):
## Automated review - eclipse-che/che-operator#2204

Reviewed all 58 changed files (+829/-851) against base `4aea588`. `go build ./...` and `go vet ./...` are clean, and the two failing tests (`pkg/deploy/gateway` missing the traefik plugin fixture, `TestReadProxyConfiguration` picking up a sandbox IP) reproduce identically on base, so they are not introduced here.

The refactoring is well structured. `sync.go` is fully retired, the commit split by helper made 58 files genuinely reviewable, error wrapping with `%w` is consistent, `ctx.Context` propagation replaces `context.TODO()` on production paths, and two real pre-existing bugs are fixed along the way: the old `DeleteByKeyWithClient` returned `(true, nil)` when the `Get` itself failed, and `identity_provider.Finalize` used to strand its finalizer when the OAuthClient was already gone. The `ObservedGeneration` fix in `6fc8507a` and the `SuppressDiff` fix in `e5788502` both landed well.

Four blocking items are inline: three cluster-scoped objects synced through the label-filtered cache, and the `Immutable` removal on the copied cert Secret. All four fail permanently rather than transiently, because the affected kinds sit outside `isRecreate` and nothing in the reconcile loop can heal them.

### One theme behind most of the findings

`consolelink.go` states the rule, in a comment added by this PR:

> `// ConsoleLink is a cluster scoped object, so it can't have an owner reference`
> `// and must be synced with the non-caching client`

That comment appears exactly once in the diff, and the rule is then not applied at three other cluster-scoped sites:

| Object | Cluster-scoped | Client |
|---|---|---|
| `ConsoleLink` | yes | `NonCachingClientWrapper` |
| `OAuthClient` (read, delete) | yes | `NonCachingClientWrapper` |
| `OAuthClient` (sync) | yes | `ClientWrapper` |
| `ClusterRole`, `ClusterRoleBinding` | yes | `ClientWrapper` |

The deleted `getClientForObject` encoded that rule in one place. Deleting it without relocating it turned one invariant into ~110 independent call-site decisions (94 `ClientWrapper`, 16 `NonCachingClientWrapper`) checked against a 20-kind selector list that lives in a different file. Same for `shouldSetOwnerReferenceForObject` - `SetControllerReference` now appears 14 times as identical boilerplate.

Both rules would fit naturally inside the wrapper: an internal `clientFor(obj)`, and an `Owner` field on `SyncOptions` handled in `doSync`. That is one change that removes the whole class.

### General findings - files not in this diff

**`pkg/common/test/deploy_context.go:75-76`** - `ClientWrapper` and `NonCachingClientWrapper` are backed by the *same* fake client, which applies no label filtering. The entire cached vs non-caching distinction is invisible to the unit suite by construction, so a green run carries no evidence about the bug class this refactoring is most likely to introduce. Wrapping the fake `ClientWrapper` in a label-selector filter mirroring `cmd/main.go:340-415` is the highest-leverage fix here - existing tests whose objects carry the `part-of` label keep passing, and the unlabeled-object behaviour becomes assertable.

**`isRecreate`** - `kind == "Service" || kind == "Ingress" || kind == "Route" || kind == "Job"`. The predicate it is really expressing is "does this kind have immutable fields that make in-place update fail", which is also true of `Secret` (`immutable`), `RoleBinding` and `ClusterRoleBinding` (`roleRef`), and `PersistentVolumeClaim`. `Secret` was in the old `isUpdateUsingDeleteCreate` and was dropped without a note. Also worth knowing: no test exercises the recreate branch at all - `k8s_client_sync_test.go` covers `ConfigMap` only, which is not in the list.

**Dead code from the signature change** - `ExposeWithHostPath` now returns `true` on every success path, so `if !done` in `pkg/deploy/dashboard/dashboard.go:60` and in the devfileregistry, pluginregistry, and openvsx reconcilers is unreachable.

**Minor** - the `logrus` to `ctrl.Log` move covers 2 files, leaving 14 on `logrus` and 23 on `ctrl.Log`; `K8sClient` has no `var _ K8sClient = (*K8sClientWrapper)(nil)` assertion and zero usages as a type, so it can drift silently; `SyncOption.ApplyToList` is named after controller-runtime's `ListOption` and only ever sets booleans to true, so options cannot be composed or overridden; copyright bumps to 2026 on mechanically-touched files add diff noise.

### Testing

`api/`, `config/crd/`, and `bundle/` are untouched, so there is no schema change, no conversion webhook, and no CR migration - the upgrade risk here is entirely behavioural. That matters for how this gets validated: **every issue above manifests only with pre-existing cluster state** (unlabeled cluster-scoped objects, already-immutable Secrets). A fresh install will pass cleanly and prove very little. The run worth doing is upgrade-from-previous-release on both Kubernetes and OpenShift.

One note on the description: "No functional/behavioral change is intended - this is a pure refactoring" is not quite accurate as it stands. The `Immutable` removal, `Secret` leaving the recreate set, and the cached vs non-caching switches are all behaviour changes, and flagging them explicitly would help reviewers more than discovering them.

---

pkg/deploy/clusterrole_test.go
```diff
@@ -49,14 +48,14 @@ func TestSyncClusterRole(t *testing.T) {
 	}
```
tolusha, line 47: **The idempotency assertion was dropped rather than replaced**

The comment above still says "sync twice to be sure update done correctly", but the check is now just `err != nil`, which passes whether the sync converged or updated on every call. The old `done == true` on an identical third sync was asserting convergence, because `doUpdate` returned false after a successful update.

`service_account_test.go` shows a good replacement pattern in this same PR - read the object back and assert its state. Same applies to the assertions removed from `route_test.go`, `job_test.go`, and `secret_test.go`.

pkg/deploy/identity-provider/identity_provider_reconciler.go
```diff
@@ -55,27 +61,35 @@ func (ip *IdentityProviderReconciler) Reconcile(ctx *chetypes.DeployContext) (re
...
-		logrus.Errorf("Error getting OAuthClients: %v", err)
+		logger.Error(err, "Failed to get OAuthClient")
 		return false
 	}
 
 	if oauthClient != nil {
-		if err := deploy.DeleteObjectWithFinalizer(ctx, types.NamespacedName{Name: oauthClient.Name}, &oauth.OAuthClient{}, OAuthFinalizerName); err != nil {
-			logrus.Errorf("Error deleting OAuthClient: %v", err)
-			return false
+		if err := ctx.ClusterAPI.NonCachingClientWrapper.DeleteByKeyIgnoreNotFound(
```
tolusha, line 69: **The OAuth finalizer is removed even when deleting the OAuthClient failed.**

A failed delete is logged and the finalizer is dropped anyway. Previously `DeleteObjectWithFinalizer` returned the error, `Finalize` returned `false`, and the delete was retried on the next pass.

Any transient failure now orphans the `OAuthClient` permanently. It is cluster-scoped, it has no owner reference, and the mechanism that would have retried the deletion is gone. The leftover is not inert - it is a live OAuth registration with the previous redirect URIs still valid, and the next Che install picks it up in `syncOAuthClient` via `GetOAuthClient`.

The comment says this is intentional, and the deadlock it avoids is real. Would it make sense to narrow it to the permanent cases rather than all of them?

```go
if err := ...DeleteByKeyIgnoreNotFound(...); err != nil {
    if !errors.IsForbidden(err) && !errors.IsNotFound(err) {
        logger.Error(err, "Failed to delete OAuthClient", "name", oauthClient.Name)
        return false
    }
    logger.Error(err, "Giving up on deleting OAuthClient, removing finalizer anyway", "name", oauthClient.Name)
}
```

If the unconditional drop stays, consider emitting a Kubernetes Event so an admin knows manual cleanup is needed. A log line from a terminating operator is easy to miss.

pkg/deploy/gateway/gateway.go
```diff
@@ -158,22 +165,45 @@ func syncAll(deployContext *chetypes.DeployContext) (bool, error) {
 	}
 
 	service := getGatewayServiceSpec(instance)
-	if done, err := deploy.Sync(deployContext, &service, deploy.ServiceDefaultDiffOpts); !done {
-		return done, err
+	if err := syncGatewayObject(deployContext, &service, deploy.ServiceDefaultDiffOpts); err != nil {
+		return false, err
 	}
 
 	if serverConfig, cfgErr := getGatewayServerConfigSpec(deployContext); cfgErr == nil {
```
tolusha, line 172: **`getGatewayServerConfigSpec` error is swallowed and the gateway reports success.**

There is no `else`. `cfgErr != nil` falls straight through to `return true, nil` - the gateway's own routing config is never applied, `SyncGatewayToCluster` reports done, and the reconciler advances.

This PR converted the two structurally identical blocks above it (`getGatewaySecretSpec` at :110, `getGatewayHeaderRewritePluginConfigSpec` at :135) into early returns. This one was left.

```suggestion
	serverConfig, err := getGatewayServerConfigSpec(deployContext)
	if err != nil {
		return false, fmt.Errorf("failed to get gateway server config: %w", err)
	}

	if err := syncGatewayObject(deployContext, serverConfig, configMapDiffOpts); err != nil {
		return false, err
	}
```

pkg/deploy/gateway/gateway.go
```diff
@@ -91,60 +95,63 @@ func SyncGatewayToCluster(deployContext *chetypes.DeployContext) (bool, error) {
...
-		oauthProxyConfig := getGatewayOauthProxyConfigSpec(deployContext, string(oauthSecret.Data["cookie_secret"]))
-		if done, err := deploy.Sync(deployContext, &oauthProxyConfig, configMapDiffOpts); !done {
-			return done, err
-		}
-	} else {
+	if err := syncGatewayObject(deployContext, oauthSecret, secretDiffOpts); err != nil {
+		return false, err
+	}
+
+	oauthProxyConfig := getGatewayOauthProxyConfigSpec(deployContext, string(oauthSecret.Data["cookie_secret"]))
```
tolusha, line 123: **The OAuth cookie secret reaches operator stdout, and one of the two paths is a ConfigMap.**

`secretDiffOpts` and `configMapDiffOpts` (lines 66-67) ignore only `TypeMeta` and `ObjectMeta`, so `Data` is compared. `syncGatewayObject` builds `&k8sclient.SyncOptions{DiffOpts: diffOpts}` with `SuppressDiff` at its zero value, so `k8s_client.go:268` prints the full diff. Two paths carry the same material:

1. `:119` syncs `oauthSecret` with `secretDiffOpts` - `Data["cookie_secret"]` is printed whenever the secret drifts or is regenerated.
2. This line reads `Data["cookie_secret"]` into `getGatewayOauthProxyConfigSpec` and the result is synced as a **ConfigMap** with `configMapDiffOpts` - the same value printed again.

The second one is worth emphasising. A fix scoped to "suppress diffs when syncing Secrets" - the obvious reading of the existing plaintext-diff thread on `secret.go` - would miss it entirely. The value is also at rest in a ConfigMap, readable by anyone with `get configmaps` in the namespace and outside whatever encryption-at-rest the cluster applies to Secrets.

The exposure predates this PR (`sync.go:doUpdate` had the identical `fmt.Printf`). What makes it a finding now is that this PR introduced `SuppressDiff` and applied it to the OAuthClient in response to review feedback, but not to the two sibling paths in the same PR carrying the same material.

Two fixes, and they compose:
- Short term: `SuppressDiff: true` on both, which needs `syncGatewayObject` to take `...k8sclient.SyncOption` (see the comment on line 182).
- Durable: route the diff in `k8s_client.go:268` through `logger.V(1).Info`, so it is suppressible by log level instead of by remembering a struct field at every call site.

pkg/deploy/gateway/gateway.go
```diff
@@ -158,22 +165,45 @@ func syncAll(deployContext *chetypes.DeployContext) (bool, error) {
...
+		if err := syncGatewayObject(deployContext, serverConfig, configMapDiffOpts); err != nil {
+			return false, err
 		}
 	}
 
 	return true, nil
 }
 
+// syncGatewayObject syncs the given object into the CheCluster namespace.
+func syncGatewayObject(deployContext *chetypes.DeployContext, obj client.Object, diffOpts ...cmp.Option) error {
```
tolusha, line 182: **Consider promoting this helper into `pkg/deploy`.**

This is the right shape, and it is the only place that has it. The same three-step body - `SetControllerReference`, `Sync`, wrap the error - is re-typed by hand in about a dozen other places: `service.go`, `secret.go`, `route.go`, `ingress.go`, `job.go`, `role.go`, `rolebinding.go`, `service_account.go`, `editors_definitions.go`, `pluginregistry.go`, `expose.go`, `dev_workspace_config.go`. The copies are near-identical but not identical - only this version includes `kind` in the message, which is why the empty-`kind` thread exists at all.

The old `deploy.Sync` was not merely plumbing. It encoded two decisions:

```go
cli := getClientForObject(blueprint.GetNamespace(), deployContext)  // cached vs non-caching
setOwnerReferenceIfNeeded(deployContext, blueprint)                 // owner ref only when same-namespace
```

Both are now the caller's responsibility, repeated fourteen times, with nothing enforcing them. The invariant "a namespaced Che-owned object gets a controller reference and goes through the cached client, a cluster-scoped one gets neither" used to live in one function. It now lives in a convention - and every cluster-scope finding on this PR (`clusterrole.go`, `clusterrolebinding.go`, `server/rbac.go`, plus `dashboard.go` and `gateway_permissions.go` below) is one copy diverging from it. Five so far.

We could restore the single decision point while keeping everything this PR improved - the explicit `error` return, the wrapped messages, the `SyncOptions` struct:

```go
func SyncOwned(ctx *chetypes.DeployContext, obj client.Object, opts ...k8sclient.SyncOption) error {
	if obj.GetNamespace() == ctx.CheCluster.Namespace {
		if err := controllerutil.SetControllerReference(ctx.CheCluster, obj, ctx.ClusterAPI.Scheme); err != nil {
			return fmt.Errorf("failed to set owner reference for %s %s/%s: %w",
				k8sclient.GetObjectType(obj), obj.GetNamespace(), obj.GetName(), err)
		}
		return ctx.ClusterAPI.ClientWrapper.Sync(ctx.Context, obj, opts...)
	}

	// cluster-scoped or foreign namespace: no owner reference, no cache
	return ctx.ClusterAPI.NonCachingClientWrapper.Sync(ctx.Context, obj, opts...)
}
```

Separately: taking `...cmp.Option` rather than `...k8sclient.SyncOption` means the helper structurally cannot express `SuppressDiff`, `DeleteOpts` or `MergeLabels`. That ceiling matters immediately - see the comments on line 123 and on `job.go`.

pkg/deploy/dashboard/dashboard.go
```diff
@@ -103,12 +98,20 @@ func (d *DashboardReconciler) Reconcile(ctx *chetypes.DeployContext) (reconcile.
 
 func (d *DashboardReconciler) Finalize(ctx *chetypes.DeployContext) bool {
 	done := true
-	if _, err := deploy.Delete(ctx, types.NamespacedName{Name: d.getClusterRoleName(ctx)}, &rbacv1.ClusterRole{}); err != nil {
+	if err := ctx.ClusterAPI.ClientWrapper.DeleteByKeyIgnoreNotFound(
```
tolusha, line 101: **Cluster-scoped delete through the cached client.**

Same class as the open threads on `clusterrole.go` / `clusterrolebinding.go` / `server/rbac.go`, but this site is not covered by any of them.

`deleteByKeyIgnoreNotFound` does a cached `Get` before deleting (`k8s_client.go:176`), and `cmd/main.go:387-392` registers `ClusterRole` / `ClusterRoleBinding` with `cache.ByObject{Label: partOfEclipseChe}`. A `ClusterRole` missing `app.kubernetes.io/part-of=che.eclipse.org` - created by an older operator version, or edited by an admin - reads as NotFound. The delete is silently skipped and `Finalize` reports success.

Cluster-scoped objects cannot have owner references, so nothing else reclaims them. The result is a live cluster-scoped RBAC object surviving CheCluster removal with no trace in the logs.

Consider `NonCachingClientWrapper` here, the way `metrics.go:266` already does in `deleteAbandonedResources`. Same for the `ClusterRoleBinding` at :110.

pkg/deploy/rbac/gateway_permissions.go
```diff
@@ -65,12 +66,20 @@ func (gp *GatewayPermissionsReconciler) Finalize(ctx *chetypes.DeployContext) bo
 
 func (gp *GatewayPermissionsReconciler) deleteGatewayPermissions(deployContext *chetypes.DeployContext) (bool, error) {
 	name := gp.gatewayPermissionsName(deployContext.CheCluster)
-	if done, err := deploy.Delete(deployContext, types.NamespacedName{Name: name}, &rbacv1.ClusterRoleBinding{}); !done {
-		return false, err
+	if err := deployContext.ClusterAPI.ClientWrapper.DeleteByKeyIgnoreNotFound(
```
tolusha, line 69: **Same cluster-scoped delete pattern here.**

Both deletes in `deleteGatewayPermissions` go through `ClientWrapper`, whose cached `Get` is label-filtered for `ClusterRole` and `ClusterRoleBinding`. An unlabeled object reads as NotFound and survives uninstall silently.

`server/rbac.go:111-130` shows the inconsistency most clearly - two adjacent `ClusterRoleBinding` deletes in the same loop, one on `ClientWrapper` and one on `NonCachingClientWrapper`. Whichever way this gets settled, it is worth settling the same way at all five sites.

pkg/deploy/server/chehost_reconciler.go
```diff
@@ -65,44 +67,50 @@ func (s *CheHostReconciler) syncCheService(ctx *chetypes.DeployContext) (bool, e
...
 			constants.DefaultServerPort,
-			getComponentName())
-		if !done {
+			getComponentName()); err != nil {
 			return "", false, err
 		}
 
 		ingress := &networking.Ingress{}
-		exists, err := deploy.GetNamespacedObject(ctx, getComponentName(), ingress)
+		exists, err := ctx.ClusterAPI.ClientWrapper.GetIgnoreNotFound(
```
tolusha, line 86: **Read-after-write against the informer cache in the same pass.**

`SyncIngressToCluster` above, then an immediate Get through `ClientWrapper`, which reads the manager cache. On the pass that creates the Ingress the informer has almost certainly not observed it yet, so `exists=false, err=nil` and `Reconcile` returns `done=false` with no error and no `RequeueAfter`. Progress then depends entirely on the Ingress watch firing.

The old shape avoided this structurally - `Sync` returned `done=false` right after the create, so the Get never ran in the same pass.

The watch should cover it, but it is worth making explicit rather than incidental. Either `NonCachingClientWrapper` for this read-back, or `reconcile.Result{RequeueAfter: time.Second}` when `!exists`. Same applies to the Route branch at :109.

pkg/deploy/job.go
```diff
@@ -51,10 +54,23 @@ func SyncJobToCluster(
...
-	return Sync(deployContext, jobSpec, JobDiffOpts)
+
+	if err := controllerutil.SetControllerReference(deployContext.CheCluster, jobSpec, deployContext.ClusterAPI.Scheme); err != nil {
+		return fmt.Errorf("failed to set owner reference for Job %s/%s: %w", jobSpec.Namespace, jobSpec.Name, err)
+	}
+
+	if err := deployContext.ClusterAPI.ClientWrapper.Sync(
+		deployContext.Context,
+		jobSpec,
+		&k8sclient.SyncOptions{DiffOpts: JobDiffOpts},
```
tolusha, line 68: **`Job` recreate has no delete propagation policy, so the old pods are orphaned.**

`Job` is in `isRecreate` (`k8s_client.go:318`), so any diff deletes and re-creates it. `doDeleteIgnoreIfNotFound` calls `k.cli.Delete(ctx, obj)` with no options, and the default propagation for Jobs is orphan - the previous run's pods keep running while the new Job starts its own.

Not a regression (`isUpdateUsingDeleteCreate` had the same gap), but `tls_utils.go` carries the workaround for exactly this symptom:

```go
// Delete pod (for some reasons pod isn't removed when job is removed)
```

`SyncOptions.DeleteOpts` exists for this and is currently unused everywhere. For the TLS secret job it matters concretely - two generations of pods racing to write the same Secret.

```suggestion
		&k8sclient.SyncOptions{
			DiffOpts:   JobDiffOpts,
			DeleteOpts: []client.DeleteOption{client.PropagationPolicy(metav1.DeletePropagationBackground)},
		},
```

tolusha (commented):
## Review summary

Thanks for taking this on - deleting a 293-line reflection-based sync layer for a net -91 lines is a real simplification, and `k8s_client.go` is the better abstraction. `ensureGVK` via `apiutil.GVKForObject` removes the empty-`Kind` footgun the old `SetGroupVersionKind` dance papered over.

The quiet win is `context.TODO()` -> `ctx.Context` across roughly 200 call sites. Reconciles now actually cancel on operator shutdown instead of running every remaining API call to completion against a client that is being torn down. That is the kind of change that only shows up in postmortems.

A few other things done well:
- Error wrapping is consistent and genuinely diagnostic - `failed to sync %s %s/%s: %w` with namespace and name at nearly every converted site.
- `consolelink.go` writes down the invariant: *"ConsoleLink is a cluster scoped object, so it can't have an owner reference and must be synced with the non-caching client."*
- `expose.go:60-75` fixes a latent bug - the old Ingress-branch closure assigned to the enclosing named return, so a cleanup failure could clobber the real result.
- `chehost_reconciler.go` only assigns `ctx.CheHost` on success now, instead of overwriting it with `""`.
- No RBAC grants are added or widened. The `PolicyRule` sets in `dashboard.go`, `gateway_permissions.go` and `tls_utils.go` are moved, not edited.

### The recurring theme

Most of the inline comments below are one thing repeated: the `(bool, error)` -> `error` collapse removed an implicit "stop this reconcile pass here" barrier, and a handful of call sites were relying on it. Worth one pass over every site where `!done` used to cause an early return, asking whether the next statement assumed the write had settled.

The second theme is that `deploy.Sync` encoded two decisions - cached vs non-caching client, and owner reference only when same-namespace - that are now the caller's responsibility at fourteen sites with nothing enforcing them. Every cluster-scope finding on this PR is one copy diverging. See the comment on `gateway.go:182`.

### Three things that do not fit on a line in this diff

**`isRecreate` could use a test pinning its contents.** `k8s_client.go` is not in this diff, so this cannot be an inline comment. The set is `{Service, Ingress, Route, Job}` and `Secret` is gone from what `isUpdateUsingDeleteCreate` covered - that is already under discussion. The structural point is that a private string switch with no test is how an entry gets dropped during a rewrite without anyone noticing. A table test asserting the membership would have caught it, and `roleRef` immutability is the next one waiting.

**The diff print bypasses the logger.** `fmt.Printf("Difference:\n%s", diff)` at `k8s_client.go:268` is now the single funnel for every mutating sync. No timestamp, no level, no `namespace`/`kind`/`name` keys, invisible to `--zap-log-level`, and unsuppressable in production without a new image. `k8s_client.go:34` already holds `logger = ctrl.Log.WithName("k8s")` - routing diffs through `logger.V(1).Info` would make the whole wrapper consistent and would fix the secret-exposure comment below by configuration rather than by remembering a struct field at every call site.

**Mixed logging is now load-bearing.** `expose.go` and `identity_provider_reconciler.go` introduce `ctrl.Log.WithName(...)` package loggers while `dashboard.go`, `server/rbac.go`, `tls_utils.go` and `gateway.go` stay on `logrus`. `k8s_client.go` itself uses `ctrl.Log`, so the target looks clear. Either convert the remaining `logrus` sites in the files this PR already touches, or defer all of them to a follow-up - the halfway state means log queries filtering on `logger=expose` silently miss half the output.

### One question

`pkg/deploy/sync.go` exported `Sync`, `GetNamespacedObject`, `GetClusterObject`, `CreateIgnoreIfExists`, `Delete`, `DeleteNamespacedObject`, `DeleteByKeyWithClient`, `DeleteIgnoreIfNotFound`, `DeleteObjectWithFinalizer` and `GetObjectType`, and about fourteen `Sync*ToCluster` signatures changed shape. All in-repo callers are updated. Does anything outside this repo import `pkg/deploy` as a library - the downstream Dev Spaces fork, or any tooling? There is no `internal/` boundary, so if the answer is yes this is a breaking change worth a release note.

### On coverage

Codecov puts the patch at 54.29% with 314 lines uncovered, `tls_utils.go` at 0%. For a refactor whose whole value proposition is "behaviour is unchanged," the test suite is the argument. The dropped `assert.True(t, done)` assertions narrow it further - there is a concrete replacement in the reply on `clusterrole_test.go`.

Nothing here is a reason not to land it. Draft status suggests another pass is already the plan.

pkg/deploy/tls/tls_utils.go
```diff
@@ -365,7 +370,7 @@ func K8sHandleCheTLSSecrets(ctx *chetypes.DeployContext) (reconcile.Result, erro
 			"LABELS":                         labels,
 		}
 
-		_, err = deploy.SyncJobToCluster(ctx, CheTLSJobName, CheTLSJobComponentName, cheTLSSecretsCreationJobImage, CheTLSJobServiceAccountName, jobEnvVars)
+		err = deploy.SyncJobToCluster(ctx, CheTLSJobName, CheTLSJobComponentName, cheTLSSecretsCreationJobImage, CheTLSJobServiceAccountName, jobEnvVars)
```
tolusha, line 373: **RBAC and the TLS job can now be applied in the same pass.**

On first install this is unchanged - the old `Sync` returned `done=true` on create. The difference is the upgrade path. When an operator upgrade widens `SyncTLSRoleToCluster`'s policy rules (:346), the old code returned on `done=false` and the Job was only created on the following pass, after the Role update had landed. Now the Job is created in the same pass and its pod may start before the updated Role is effective, failing with an RBAC denial.

Narrow window, and the job is retried, but the ordering was load-bearing. This is the clearest instance of the general theme: worth one pass over every site where `!done` used to cause an early return, asking whether the next statement assumed the write had settled.

pkg/deploy/tls/tls_utils.go
```diff
@@ -531,24 +540,22 @@ func GetAdditionalCACertsConfigMapVersion(ctx *chetypes.DeployContext) string {
 // Does nothing if secret with given name already exists.
 func CreateTLSSecret(ctx *chetypes.DeployContext, name string) (err error) {
```
tolusha, line 541: **`CreateTLSSecret` swallows non-NotFound errors.**

Three things in one function. All pre-existing, but the lines are touched here:

- The inner `if err := ...Get(...)` shadows the named return `err`, which is never assigned. The function always reaches `return nil`. `go vet` will not catch this.
- Any non-NotFound error falls straight through to `return nil`, so an API failure looks like success.
- The Get uses the cached client, whose Secret informer is label-filtered, so an existing but unlabeled secret reads as NotFound.

Codecov puts this file at 0% patch coverage with 43 lines missing, which makes it the riskiest gap in the PR - it holds TLS secret generation, the job recreate, and the wait loops. Worth tightening while it is open.

pkg/deploy/devworkspace/dev_workspace_config.go
```diff
@@ -74,10 +74,14 @@ func (d *DevWorkspaceConfigReconciler) Reconcile(ctx *chetypes.DeployContext) (r
 		return reconcile.Result{}, false, err
 	}
 
-	done, err := deploy.Sync(ctx, dwoc)
-	if !done || err != nil {
+	if err := controllerutil.SetControllerReference(ctx.CheCluster, dwoc, ctx.ClusterAPI.Scheme); err != nil {
 		return reconcile.Result{RequeueAfter: time.Second}, false, err
 	}
+
+	if err := ctx.ClusterAPI.ClientWrapper.Sync(ctx.Context, dwoc); err != nil {
```
tolusha, line 81: **This is the only `Sync` call site with no `DiffOpts`.**

Every sibling passes one. With an empty option set, `cmp.Diff(actual, obj)` compares the entire object graph including `ObjectMeta` - `resourceVersion`, `managedFields`, `creationTimestamp` - so the diff will rarely be empty and the operator likely issues an `Update` on the DWOC every reconcile. That is a permanent write loop against the API server for one object, plus the audit-log volume.

Worth confirming a second thing: `cmp.Diff` panics on unexported fields unless every type in the graph provides `Equal`, which is a live question for the `resource.Quantity` values DWOC carries in its storage-size fields. Pre-existing (`deploy.Sync(ctx, dwoc)` also passed no opts) and evidently not panicking in practice, so this is a question rather than a claim. But it is the one call site that got no diff options while every neighbour got one, and this PR is a good moment to find out why.

pkg/deploy/devfileregistry/devfileregistry.go
```diff
@@ -43,10 +44,20 @@ func NewDevfileRegistryReconciler() *DevfileRegistryReconciler {
...
```
tolusha, line 59: **The delete loop discards every error.**

Same as before the change (`_, _ =`), so no regression, but this is a natural place to at least log. When one of these deletes fails there is no log line, no event and no metric - the only evidence is the resource that is still there.

The same anonymous-struct table appears verbatim in `postgres.go:42-54` and `pluginregistry.go`. All three are "delete this list of obsolete component objects". A `deploy.DeleteAll(ctx, map[string]client.Object{...})` would replace all three and give the errors somewhere to go. As written, the refactor turned three one-liners into three eleven-line blocks that do the same thing and still discard everything. Related to the delete-error-handling thread already open on this PR.

pkg/deploy/service_test.go
```diff
@@ -24,19 +23,19 @@ import (
...
-	if !done || err != nil {
+	err := SyncServiceToCluster(ctx, "test", []string{"port"}, []int32{8080}, "test")
+	if err != nil {
 		t.Fatalf("Failed to sync service: %v", err)
 	}
 
 	// sync another service
-	done, err = SyncServiceToCluster(ctx, "test", []string{"port"}, []int32{9090}, "test")
-	if !done || err != nil {
+	err = SyncServiceToCluster(ctx, "test", []string{"port"}, []int32{9090}, "test")
```
tolusha, line 32: **The recreate path is the one behaviour this test walks past.**

This syncs port 8080, then 9090, then Gets the Service. `Service` is in `isRecreate`, so the second sync **deletes and recreates** it - and `isRecreate`'s membership is the subject of two open threads on this PR.

The test verifies neither that the recreate path was taken nor that `ClusterIP` survived it. Would it be possible to assert that here? It is the single behaviour most at risk in this refactor, and the test walks straight past it.

pkg/deploy/clusterrole_test.go
tolusha: Agreed, and there is a replacement that is strictly stronger than what was removed. Every one of these tests already fetches the object afterwards, so we can assert on the API instead of trusting a return value:

```go
actual := &rbacv1.ClusterRole{}
err = ctx.ClusterAPI.Client.Get(ctx.Context, types.NamespacedName{Name: "test"}, actual)
assert.NoError(t, err)
assert.Equal(t, "test-2", actual.Rules[0].APIGroups[0])

// idempotency: a second identical sync must not write
rv := actual.ResourceVersion
assert.NoError(t, SyncClusterRoleToCluster(ctx, "test", rules_v2))
assert.NoError(t, ctx.ClusterAPI.Client.Get(ctx.Context, types.NamespacedName{Name: "test"}, actual))
assert.Equal(t, rv, actual.ResourceVersion)
```

`ResourceVersion` stability verifies the property at the API rather than trusting the old `done` bool - a `Sync` that silently no-opped, or that deleted the object and failed to recreate it returning nil, would still pass the current version.

Worth applying across the nine files where the "sync twice to be sure update done correctly" comment is now stranded with no assertion behind it.

rohanKanojia (approved):

